Issue #16, opened 2020-03-21
Superuser creation fails because of null_value in user_type


yahya31415:
Stacktrace:

```
Traceback (most recent call last):
  File "/Users/yahya/Developer/care/venv/lib/python3.7/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
psycopg2.errors.NotNullViolation: null value in column "user_type" violates not-null constraint
DETAIL:  Failing row contains (2, argon2$argon2i$v=19$m=512,t=2,p=2$bzJ4OHp4Wk9uQUh3$yh93Cj2/p0dmT..., null, t, yahya31415, , , yahya31415@gmail.com, t, t, 2020-03-21 05:49:22.252838+00, null, null, , null, null, null, f).


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "manage.py", line 30, in <module>
    execute_from_command_line(sys.argv)
  File "/Users/yahya/Developer/care/venv/lib/python3.7/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/Users/yahya/Developer/care/venv/lib/python3.7/site-packages/django/core/management/__init__.py", line 375, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/Users/yahya/Developer/care/venv/lib/python3.7/site-packages/django/core/management/base.py", line 323, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/Users/yahya/Developer/care/venv/lib/python3.7/site-packages/django/contrib/auth/management/commands/createsuperuser.py", line 61, in execute
    return super().execute(*args, **options)
  File "/Users/yahya/Developer/care/venv/lib/python3.7/site-packages/django/core/management/base.py", line 364, in execute
    output = self.handle(*args, **options)
  File "/Users/yahya/Developer/care/venv/lib/python3.7/site-packages/django/contrib/auth/management/commands/createsuperuser.py", line 156, in handle
    self.UserModel._default_manager.db_manager(database).create_superuser(**user_data)
  File "/Users/yahya/Developer/care/venv/lib/python3.7/site-packages/django/contrib/auth/models.py", line 162, in create_superuser
    return self._create_user(username, email, password, **extra_fields)
  File "/Users/yahya/Developer/care/venv/lib/python3.7/site-packages/django/contrib/auth/models.py", line 145, in _create_user
    user.save(using=self._db)
[...]

PR #19, opened 2020-03-21
Fixed Super User Creation Issue

thetronjohnson:
Added required fields to users.models.py
#16


vigneshhari:
In the migrations i am seeing a user_type field defaulted to the integer value 3 which is not a valid value, kindly change that

thetronjohnson:
@vigneshhari Removed the defaulted value, is it okay now?

vigneshhari:
Merging Now

Issue #24, opened 2020-03-21
Add a contact number field for facility.


bodhish:
This is a requirment that came from `Ambulance Network`

abinmn:
Can I take up the issue?

bodhish:
@abinmn can you check with @vigneshhari

vigneshhari:
@abinmn Continue and add , make sure you add the migrations as well

abinmn:
Is this a mandatory field?

vigneshhari:
@bodhish

bodhish:
We are yet to go live. So it's safe to seed 1234567890

PR #30, opened 2020-03-21
Add contact number field for facility

abinmn:
#24


vigneshhari:
Set the phone number field's label as " Contact Number for Hospital "

PR #36, opened 2020-03-21
feat(requirements) - add django rest framework as dependency

tucosaurus:
Creating a separate PR for this since there might be other tasks/issues which are dependent on drf being installed. There is also code on master which uses this.
If I am missing something here and drf is already getting installed, please feel free to close this PR.
https://github.com/coronasafe/care/blob/master/care/users/api/views.py


vigneshhari:
Yeap, DRF will be needed once we start working for API's

PR #37, opened 2020-03-21
feat(*) - add authentication through jwt. fixes #17.

### LINKED ISSUE #17: need a current user endpoint.

tucosaurus:
Also solves #17


tucosaurus:
@vigneshhari Browsable Api works through [session authentication](https://www.django-rest-framework.org/api-guide/authentication/#sessionauthentication). In the event that we want to merge this in. Where should I add the api docs ?

vigneshhari:
Before we definitely switch to JWT's i wanted to have a talk with the front end folks to see what they are comfortable with so that there are no issues. That is causing a delay, for the time being, we can create a sample class for API's and let all our API Views Extend that, so if we wanted to add JWT's it would be really simple to do so. Until Then let's Continue with session-based auth.

vigneshhari:
If we work with DRF only then we can generate API docs with swagger, the front end folks would be happier too.

tomahawk-pilot:
@vigneshhari @tucosaurus we also welcome jwt

vigneshhari:
Created Another PR and Merged

PR #42, opened 2020-03-21
Add api endpoint - facility

abinmn:
- Add DRF to installed apps
- Create endpoints related to `facility` model.


vigneshhari:
is_active is an internal field and should not be exposed via API

abinmn:
The current implementation return choice fields(district,facility_type) as integers. Is this is the desired behavior or should it return data in human-readable form? (Though the choice mapping can be obtained by OPTIONS method).

PR #62, opened 2020-03-22
Adds lat long to facility

afzalIbnSH:


afzalIbnSH:
I could not add the option to the form.
```
from django.contrib.gis import forms

class MyGeoForm(forms.Form):
    point = forms.PointField(widget=
        forms.OSMWidget(attrs={'map_width': 800, 'map_height': 500}))
```
is how to do it but I couldn't use this in the current ModelField

afzalIbnSH:
Added option to the API.
Below is how a valid location to the API should look like;
```
  "location": {
      "latitude": 49.8782482189424,
      "longitude": 24.452545489
  }
```

shanavas786:
This requires postgis to be installed and configured. 
@bodhish @vigneshhari does current environment support it ?

vigneshhari:
Heroku does have postgis built in

afzalIbnSH:
> This requires postgis to be installed and configured.
> @bodhish @vigneshhari does current environment support it ?

I had coordinated with @vigneshhari prior. He has already added postgis to the heroku app.

shanavas786:
Great.

@afzalIbnSH I suggest to use 
https://github.com/caioariede/django-location-field package

Makes things a lot easier

afzalIbnSH:
> Great.
> 
> @afzalIbnSH I suggest to use
> https://github.com/caioariede/django-location-field package
> 
> Makes things a lot easier

We're moving to a react app and the app devs would just use the API right. It accepts
```
  "location": {
      "latitude": 49.8782482189424,
      "longitude": 24.452545489
  }
```
and I think that's super intuitive and more than meets our need.

config/settings/base.py
```diff
@@ -41,7 +41,7 @@
 # ------------------------------------------------------------------------------
 # https://docs.djangoproject.com/en/dev/ref/settings/#databases
 
-DATABASES = {"default": env.db("DATABASE_URL", default="postgres:///care")}
+DATABASES = {"default": env.db("DATABASE_URL", default="postgis:///care")}
```
biswaz: Do we really need this, since postgis is an extension on top of postgres

afzalIbnSH: I checked. This seems to be the way. The standard looks like the below
```
DATABASES = {
    'default': {
        'ENGINE': 'django.contrib.gis.db.backends.postgis',
        'NAME': 'YOUR_POSTGIS_DATABASE_NAME',
        'USER': 'YOUR_DATABASE_USER_NAME',
        'PASSWORD': 'YOUR_DATABASE_USER_PASSWORD',
        'HOST': 'localhost',
        'PORT': '5432'
    }
}
```

Since we're using url directly, what I've done seems to be the way.

afzalIbnSH:
@shanavas786 Switched to https://github.com/caioariede/django-location-field package. Thanks for the suggestion. I couldn't get to adding it to the creation form though. Gotta log off.

config/settings/base.py
afzalIbnSH: You might have to change the value of `DATABASE_URL` in heroku app env to do the same.

Issue #63, opened 2020-03-22
Create a model for Ambulance and expose API


bodhish:
# Vehicle Details 
- Vehicle Registration Number: string (validation: A-Z and 0-9 (Alpha numeric only))
- Insurance Valid till: date (validation- year should one of 2020 / 2021 / 2022)
- Name of Owner: string
- Cell Number of Owner: number (10 digit mobile number)
- Smart Phone: bool  
- Primary District Served: District.id
- Secondary District Served: District.id
- Third District Served: District.id
- Oxygen Yes / No
- Ventilator Yes / No
- Suction Machine Yes / No
- Defibrillator Yes / No

## Driver Details
Name of Driver 1
Cell Number of Driver 1
Smart Phone Yes / No

## Driver Details
Name of Driver 2
Cell Number of Driver 2
Smart Phone Yes / No

# Checklist
- [ ] Create the table and organize the data
- [ ] Add a CRUD API for front-end
- [ ] Add an API to get ambulance details given an array of registration number

anroopak:
I can take this

rejuls:
@anroopak you on it?

anroopak:
@rejuls Yes

bodhish:
A minor update.

Given an array of registration numbers return an array with details.

anroopak:
APIs - 
POST /ambulance/
```
{
            "facility": 1, // facility ID
             "drivers": [
                {
                    "name": "Roopak A N",
                    "phone_number": "+91",
                    "is_smart_phone": true
                }
            ],
            "vehicle_number": "KL13A1234", // without spaces
            "owner_name": "Name",
            "owner_phone_number": "+91",
            "owner_is_smart_phone": true,
            "primary_district": 13,
            "secondary_district": 11,
            "third_district": null,
            "has_oxygen": true,
            "has_ventilator": true,
            "has_suction_machine": false,
            "has_defibrillator": true,
            "insurance_valid_till_year": 2020,
           
        }
```

GET ambulance/?vehicle_numbers=KL13A1234,KL13B1234 // comma seperated
```{
    "count": 2,
    "next": null,
    "previous": null,
    "results": [
        {
            "id": 1,
            "drivers": [
                {
                    "id": 1,
                    "name": "Roopak A N",
                    "phone_number": "+91",
                    "is_smart_phone": true
                }
            ],
            "vehicle_number": "KL13B1234",
            "owner_name": "Name",
            "owner_phone_number": "+91",
[...]

PR #67, opened 2020-03-22
add ambulance and driver models and apis

anroopak:
Implements - https://github.com/coronasafe/care/issues/63


vigneshhari:
There seems to be a lot of .idea config files, could you untrack them ? and could you resolve the conflicts as well?

vigneshhari:
Left some comments at #63

Issue #71, opened 2020-03-23
Add pre-commit hooks to maintain code styles


anroopak:
Add some common code styles as pre-commit hooks. 

Proposed hooks - 

- trailing-whitespace
        name: trim trailing whitespace
        description: This hook trims trailing whitespace on files
        entry: trailing-whitespace-fixer
- check-merge-conflict
        name: check for merge conflict
        description: Prevent accidentally commiting files with merge conflicts.
- end-of-file-fixer
        name: fix end of files.
        description: Ensures that a file is either empty, or ends with one newline.
- seed-isort-config
        description: sorts the imports, hence consistency in the import statements
- isort
        description: additional to seed-isort-config
- black
       description: ensures that commits don't go through without specific coding style

biswaz:
Might slow down rapid prototyping, but I'm open to adding this now rather than later. We can make it non intrusive for now. Just ran flake8 now, its mostly isort errors only.

anroopak:
One good thing about adding this would be that when we have more devs, the merging would not pull us down. Currently flake8 just shows the errors. If we replace that with black, isort, seed-isort-config - these pre-commit hooks takes care of the file changes.

biswaz:
Could you mind setting this up?

anroopak:
https://github.com/fluxility/pre-commit-django-migrations/blob/master/.pre-commit-hooks.yaml
I would suggest that we add this also.

anroopak:
Yep, almost done with this.

vigneshhari:
@biswaz please check the issue and merge it ASAP , as we wait it gets harder

PR #72, opened 2020-03-23
add pre-commit-hooks - black, isort, django-migrations-check

anroopak:
Though this looks like a huge diff, the major diff is only in the pre-commit config files.

View only these files - 
.pre-commit-config.yaml
setup.cfg

Rest of the files have been changed due to these pre-commit hook changes

Related Issue: https://github.com/coronasafe/care/issues/71


tucosaurus:
The only issue I see with this is that it overwrites the git history for quite a number of files. A `git blame` would not be that helpful to get the context for a line of code. If the maintainers are fine with that, this is good to merge. @biswaz @vigneshhari

anroopak:
It does not overwrite, it appends to the git history. Git blame and still find the previous editors (I assume). And this is only one-time, after every developer starts using the hooks, editors would be the particular dev only.

vigneshhari:
@biswaz review and merge

biswaz:
@anroopak could you document usage, such as manually triggering the hooks etc (if possible) in the README

biswaz:
Also add precommit package to requirements file (assuming you are using that). Any other helpful commands from https://pre-commit.com/ wil help

anroopak:
@biswaz added documentation. `pre-commit` was already part of the initial requirements.

Issue #73, opened 2020-03-23
Create Doctors and Facility Beds Count API


vigneshhari:
Complete the API for facility-related models like HospitalDoctor and Facility Bed Counts.

anroopak:
Is this the response you want?
Like this - 
```
GET /facility/:id/stats/
{
    doctors: 123, 
    beds: 123
}
```

vigneshhari:
No To update , remove and create Doctor counts and Bed Counts

vigneshhari:
CRUD operation on the FacilityCapacity and HospitalDoctors Model

anroopak:
@vigneshhari which one do we need?
## Option 1
```
PATCH /facility/:id/
{
  ... // other existing fields
  hospital_doctors: [
        {
            "area": 1, // upserts based on the area
            "count": 12
        }
    ],
  capacity: [
      {
        "room_type": 1, //upserts based on the room_type
        "total_capacity": 100,
        "current_capacity": 80
      }
  ]
}
```

## Option 2
```
PATCH /facility/:id/hospital_doctors/
// Can allow array also - 
{
    "area": 1, // upserts based on the area
    "count": 12
}

PATCH /facility/:id/capacity/
// Can allow array also - 
{
    "room_type": 1, //upserts based on the room_type
    "total_capacity": 100,
    "current_capacity": 80
}
```

vigneshhari:
Option 2 , we need a create , update ,softdelete and list endpoint , the permission should be given only for users who have created the facility.

anroopak:
Ok. Picking this up.

PR #74, opened 2020-03-23
add apis for hospital capacity and doctors

anroopak:
Related Issue: https://github.com/coronasafe/care/issues/73


vigneshhari:
why is the id excluded from the serializer ?

anroopak-ab:
@vigneshhari the primary key is ignored. I am using `area` in doctors and `room_type` in capactiy to identify. 

Eg: API would look like this - /facility/{facility_id}/hospital_doctors/{area_id}
Hence ID of that entity would not matter, and hence ignored.

biswaz:
lgtm

PR #78, opened 2020-03-23
Disable browser suggestions in user signup form

### LINKED ISSUE #48: Disable suggestion for input fields.
On Mac OS, the suggestion model is stuck 🙄. 

IMG_20200322_113952

Kailaash-Balachandran:
Fixes #48 : 
1. Disable suggestion for input fields (except the first name, last name, and phone number)
2. Use Crispyform helper to render form fields.

### DIFF
--- a/care/templates/users/signup.html
+++ b/care/templates/users/signup.html
@@ -16,7 +16,7 @@ Already have an account? Then please <a href="{% url 'users:signin' %}" class="c
 <div class="max-w-md mt-4">
     <form class="signup" id="signup_form" method="post" action="">
         {% csrf_token %}
-        {{ form|crispy }}
+        {% crispy form %}
         {% if redirect_field_value %}
         <input type="hidden" name="{{ redirect_field_name }}" value="{{ redirect_field_value }}" />
         {% endif %}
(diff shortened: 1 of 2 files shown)


vigneshhari:
why was the placeholders removed while adding the autocomplete fields ?

Kailaash-Balachandran:
> why was the placeholders removed while adding the autocomplete fields ?

Placeholders were unused in the template (https://care.coronasafe.network/users/signup/staff/). 

Previously, the form was rendered with `{{ form|crispy }}`, which doesn't use the form helper methods. This PR proposes to use `{% crispy form %}` instead to render the helper methods. 

However, if placeholders are required, I can add them back.

vigneshhari:
@rejuls check if the UI is rendering fine , if so Merge.

rejuls:
@Kailaash-Balachandran did you test the form after the changes? I think with the crispy forms tag require the submit button to be passed from the form class and not inside the template. Can you check if data is being POST?

Kailaash-Balachandran:
> @Kailaash-Balachandran did you test the form after the changes? I think with the crispy forms tag require the submit button to be passed from the form class and not inside the template. Can you check if data is being POST?

@rejuls  Great catch. I've now added `self.helper.form_tag = False` in the helper. Also, tested.

PR #87, opened 2020-03-24
fix created_by field population in facility bulk_upsert

anroopak:
In the previous PR, the create_by field was not populated, hence the person who created it could not see the items. This fixes the issue.


anroopak:
@vigneshhari @gigincg @biswaz

Issue #99, opened 2020-03-25
Add tables for state, district and panchayat


bodhish:
- [x] Add colum for panchayat in facility. 
- [x] Seed data for kerala
- [x] Migrate Existing Data

vigneshhari:
We need to write custom migrations to update the current district field to the new table as well,  We need to have API's to get the districts under the state and so on ..

vigneshhari:
@bodhish can we get an initial dataset for the models?  @tomahawk-pilot do you have it ?

anroopak:
I can pickup :raised_back_of_hand:

vigneshhari:
@afzalIbnSH you can also pitch in and solve some parts parallelly so we can push to production faster, you guys can collaborate

anroopak:
When you say, I am assuming that it is the Local Body (Panchayath / Municipality / Corportation)
Do we need a Ward level info?

vigneshhari:
@bodhish Please confirm

bodhish:
we just need panchayat details.

anroopak:
@bodhish there are hospitals which are inside Municipalites and Corporations? I feel we can generalise at "Local Body" levels.

bodhish:
My understanding about the organization is completely wrong.

anroopak:
Strategy for migration

First PR
1. will create the models, existing district field will not be removed, new field will be added.
2. add districts
3. populate the existing tables with district objects. 

Second PR (after the first PR is merged and released)
1. Remove the earlier district column
2. Rename the newly added district column to existing `district` column

anroopak:
PRs raised. @vigneshhari Review please.
Only list and retrieve APIs are posted for the new entities. 
Filters based on District and Local Body names have been added to Facility and Ambulance.

anroopak:
Facility Create
/facility/
{
  ...existing fields
  "local_body": id, 
  "district": id,
   "state": id
}

List/Detail
{
   ...existing,
   // only state will not be null, others can be null considering the data the states give
   "local_body": id,
   "local_body_object": {
          "id",
          "name", 
          "district": id
       },
    "district": id
    "district_object": {
          "id",
          "name", 
          "state": id
       },
     "state": id,
    "state_object": {
          "id",
          "name"
       }
 
}

PR #100, opened 2020-03-25
add state, district, local body models

anroopak:
Related Issue: https://github.com/coronasafe/care/issues/99


vigneshhari:
We also have to populate all the data in Kerala, possibly in Malayalam too, write management commands for this and use a json data dump. Also when we port the existing data how will we figure out the local body date

afzalIbnSH:
> Also when we port the existing data how will we figure out the local body date
Put `Panchayath` as default ?

vigneshhari:
we still need the panchayath data right ?

vigneshhari:
how do we say that this hospital comes under X Panchayath

anroopak:
We currently cannot figure out the local body without information from data entry team. So we can match then against a default panchayath and then when data entry comes in, we can update it. 

Second way is that we match only district fields now, leave local body as blank for now. And once data is in, we can map them.

anroopak:
Ambulance currently uses District, should that also be changed to Panchayaths?

vigneshhari:
Everything that uses Districts should be changed to this model, for Ambulances lets keep with just the district Table.

vigneshhari:
leaving them blank is good for now

vigneshhari:
Tmrw i will bring someone from open street map so we can correctly map the structure of the location and get a reliable source of data to populate them as well, Okay ?

anroopak:
@vigneshhari added models and initial data migration. This is the first PR as part of the plan mentioned here - https://github.com/coronasafe/care/issues/99#issuecomment-604030314

vigneshhari:
@anroopak we will block it until we can populate all the Lower Level Government details, will get it by tmrw noon, @tomahawk-pilot , we will NULL the lower level govt for the migrated data, We also have to open up API's for getting districts from states, panchayaths from districts and states and so on for the React Guys to shift properly.

tomahawk-pilot:
https://github.com/rebuildearth/data/tree/master/data/india/kerala/lsgi_site_data

vigneshhari:
@anroopak check if we can use this data and it agrees with our database model

anroopak:
@vigneshhari Yes, this looks fine. We can go ahead and populate based on this JSON. That can be done seperately.

vigneshhari:
@anroopak write a management command that syncs the json to the database model then, so its easier for us

anroopak:
@vigneshhari Can we merge this?

biswaz:
Still needs a data migration/one time script to populate lsgb data to the d. I prefer a data migration.

PR #105, opened 2020-03-25
Feat/add state district panchayath apis

anroopak:
Merge only after: https://github.com/coronasafe/care/pull/100
Related Issue: https://github.com/coronasafe/care/issues/99


care/users/api/serializers/lsg.py
```diff
@@ -0,0 +1,26 @@
+from rest_framework import serializers
```
afzalIbnSH: Two points:
- Is `lsg` a good name? We plan to make this usable across India right? Is lsg the name used across? I checked "Karantaka lsg" and "Tamil Nadu lsg" and couldn't find them using this name. How about `address`?
- I suggest putting these in a separate app altogether since these are not really exclusive to a user but used by facility too.

care/users/api/serializers/user.py
```diff
@@ -1,15 +1,14 @@
...
 from config.serializers import ChoiceField
 
 User = get_user_model()
 
 
 class UserSerializer(serializers.ModelSerializer):
     user_type = ChoiceField(choices=User.TYPE_CHOICES)
-    district = ChoiceField(choices=DISTRICT_CHOICES)
     gender = ChoiceField(choices=GENDER_CHOICES)
     password = serializers.CharField(write_only=True)
```
afzalIbnSH: Add
```
district = DistrictSerializer(read_only=True)
district_id = serializers.IntegerField(write_only=True)
```
?

care/facility/api/serializers/ambulance.py
```diff
@@ -18,6 +19,19 @@ class Meta:
...
+        data["primary_district"] = (
+            DistrictSerializer().to_representation(instance.primary_district) if instance.primary_district else None
+        )
+        data["secondary_district"] = (
+            DistrictSerializer().to_representation(instance.secondary_district) if instance.secondary_district else None
+        )
+        data["third_district"] = (
+            DistrictSerializer().to_representation(instance.third_district) if instance.third_district else None
+        )
+        return data
```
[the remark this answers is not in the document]
anroopak: Problem with this is that there will be inconsistency. Django RF standards are to use model name - eg: `district`. Also we use that approach in other places, so using `_id` in only a few places will break the consistency.

care/users/api/serializers/lsg.py
anroopak: LSG (Local Self Governing Body) is the actual word used in the rule books. But yeah, we dont use it colloquially. 

Moving to different app? I retained it there because of District Choices, I don't see LSG growing out of this heirarchy, also sinces Users are at LSG level, we can retain there.

afzalIbnSH: Ok

afzalIbnSH: @vigneshhari @biswaz thoughts ?

vigneshhari: lsg is the standard name for storing local governing bodies

vigneshhari:
@anroopak good to merge ?

PR #114, opened 2020-03-26
Add LocalBody model to Django Admin

agzuniverse:
If there is no particular reason why this model is not available in Django Admin, this change would be useful for debugging and adding data using Django Admin.


anroopak:
This is being added in the following PR.

anroopak:
Considering that we are not merging the earlier said PR, this may be merged.  @vigneshhari

anroopak:
@agzuniverse Hey, this was solved in #105 please confirm :)

agzuniverse:
Yes it does. Closing now.

Issue #126, opened 2020-03-27
District Level Access User Group


gigincg:
Need to create a usergroup that has access to all the facilities in a certain district. Will be used to provide accounts to district collectors.

anroopak:
@bodhish planning to add a new user type.

anroopak:
@gigincg @bodhish should they be allowed to update, like the super users?
I am assuming yes, for now.

bodhish:
Yes, the use case is to allow district admin to have an overview of the entire system and correct details.

PR #138, opened 2020-03-27
add district admin level access

anroopak:
Related Issue: https://github.com/coronasafe/care/issues/126


vigneshhari:
In the signup form, we still accept the type value as integers, shouldn't we add checks so that district admins can only be created by super admins, else we can change a value in the signup form and we get to be a district admin.

anroopak:
> In the signup form, we still accept the type value as integers, shouldn't we add checks so that district admins can only be created by super admins, else we can change a value in the signup form and we get to be a district admin.

Agreed. But this change has to be in a different PR, right? Should this PR be blocked?

Issue #128, opened 2020-03-27
Map patients to facilities


afzalIbnSH:
Create a many to many mapping as:
- A facility will have multiple patients
- More importantly, we need history of a patient's move between facilities. Use a flag to denote the current facility a patient is in.

afzalIbnSH:
DB Proposal:

FacilityAdmission

- facility_id, int referencing Facility, required
- patient_id, int, referencing PatientRegistration, required
- admission_date, timestamp, required
- is_active, boolean, "If the patient is currently in the facility"
- discharge_date, timestamp

Constraints:
- partial uniqueness => patient_id where is_active = True
- check ( (is_active) or (discharge is not null) )

afzalIbnSH:
APIs:
- Admit patient to facility => `PUT /facility/{id}/patient/{id}/`
- Discharge patient from facility => `DELETE /facility/{id}/patient/{id}/`
- List patients in a facility => `GET /facility/{id}/patient/`

Also,
- Return facility_id in `GET /patient/` and `GET /patient/{id}`. It should be the id of the facility that the patient is currently admitted to, if he is.

afzalIbnSH:
@anroopak @vigneshhari Please review the db and API proposals. I'll get to dev late night or early morning. Got some office stuff to do. Feel free to move it to your plates if this is required earlier.

anroopak:
@afzalIbnSH In the Patient Detail, do you want to show the history?

afzalIbnSH:
> In the Patient Detail, do you want to show the history?

@anroopak Only if there is such requirement? Or I can accept a query param, say, `history=<bool>` and return history only if true, to avoid unnecessary db load.

bodhish:
A few additions required. 
We cannont assume that the patient will be admitted in a facility. 
Their are 3 outcomes for a patient who visits a hospital

1. They may be asked to go home for home isolation
2. They would be asked to get admitted in the hospital
3. They would be reffered to another facility. 

We should have an option to capture these detials. 

Ex: A patient asked to be in home isolation will have to be followed up. (This can be a manuel process for now) We should store these information so that, the next time when they visit the same facility the doctor will have better data

Basic patient detailes has to captured for all the 3 cases. The outcome will be one of the 3.

afzalIbnSH:
@anroopak Added a separate endpoint for patient history:
- GET /patient/{id}/history/

afzalIbnSH:
@bodhish Update db:

Table => `patient_consultation`

[...]

PR #139, opened 2020-03-27
Issue 128

afzalIbnSH:
Issue: #128


care/facility/api/serializers/patient.py
```diff
@@ -2,13 +2,21 @@
...
 )
 
 
 class PatientSerializer(serializers.ModelSerializer):
     medical_history = fields.MultipleChoiceField(choices=MEDICAL_HISTORY_CHOICES)
+    admitted_at = serializers.SerializerMethodField(source="", read_only=True)
+
+    def get_admitted_at(self, obj):
+        try:
+            return PatientAdmission.objects.only("facility_id").get(patient_id=obj.id, is_active=True).facility_id
```
anroopak: Would it be better to return the whole `Facility` object?

care/facility/api/viewsets/patient.py
```diff
@@ -26,3 +32,59 @@ def get_serializer_class(self):
...
+        return Response(data=PatientAdmissionSerializer(queryset, many=True).data)
+
+
+class PatientAdmissionViewSet(viewsets.ModelViewSet):
+    permission_classes = (IsAuthenticated,)
+    queryset = PatientAdmission.objects.all()
+    serializer_class = PatientAdmissionSerializer
+    http_method_names = ["put", "delete", "get"]
+
+    def get_queryset(self):
```
anroopak: Do you want to use the `UserAccessMixin`?

anroopak: and then override on the facility_pk here?

care/facility/api/viewsets/patient.py
```diff
@@ -26,3 +32,59 @@ def get_serializer_class(self):
...
+        queryset = self.queryset.filter(facility__id=self.kwargs.get("facility_pk"), is_active=True)
+        if user.is_superuser:
+            return queryset
+
+        return queryset.filter(facility__created_by=user)
+
+    def update(self, request, *args, **kwargs):
+        data = {"facility_id": self.kwargs.get("facility_pk"), "patient_id": self.kwargs.get("pk")}
+        admission_date = request.data.get("admission_date")
+        if admission_date:
```
anroopak: In some case the entry may be made while discharging. We cannot expect everything to be done in a seamless way. So we should be allowing to enter the previous data, and hence `discharge_date`

care/facility/api/viewsets/patient.py
```diff
@@ -26,3 +32,59 @@ def get_serializer_class(self):
...
+            with transaction.atomic():
+                PatientAdmission.objects.create(**serializer.validated_data)
+        except IntegrityError:
+            admitted_at = PatientAdmission.objects.get(
+                patient_id=serializer.validated_data["patient_id"], is_active=True
+            ).facility_id
+            raise ValidationError(f"The patient is currenly admitted at facility {admitted_at}")
+        return Response(status=204)
+
+    def destroy(self, request, *args, **kwargs):
```
anroopak: How do we differentiate with a wrong entry and a discharged entry?

Issue #137, opened 2020-03-27
Create a system to store details on samples taken


bodhish:
Their should be an option to capture details of samples taken in each facility. 
This should be associated with a patient record.
Sample can be taken multiple times.
Possible outcomes for a sameple is [Positive, Negative, Resutl Awaiting]
Date of sample taken should be recorded

anroopak:
### Model: Sample
```
facility_id
patient_id - refers to PatientRegistration
result - enum
date_of_sample
date_of_result
```

### APIs
POST /sample/
```
Request:
{
  "facility": id,
  "patient": id,
  "date_of_sample": 
   "result": (int option), // supporting this feature so that previous data can be entered. if empty "Result Awaiting"
  "date_of_result":  // nullable field
}

Response
{
   "id":  // sample id,
    ... other fields
}
```

PATCH /sample/:id/
```
 "date_of_sample": 
   "result": (int option), 
  "date_of_result":  // nullable field
```

@bodhish please confirm

krishnanunnir:
Hey, I would like to help out

anroopak:
This was completed. Closing.

PR #140, opened 2020-03-27
add patient sample model and viewsets

anroopak:
Related Issue: https://github.com/coronasafe/care/issues/137


bodhish:
@anroopak can you check  #151 also

afzalIbnSH:
Is it better to map a sample to a consultation ? @anroopak

care/facility/models/patient_data.py
```diff
@@ -66,6 +70,14 @@ class PatientTeleConsultation(models.Model):
     created_by = models.ForeignKey(User, on_delete=models.SET_NULL, null=True)
 
 
+class PatientSample(FacilityBaseModel):
+    patient = models.ForeignKey(PatientRegistration, on_delete=models.PROTECT)
+    facility = models.ForeignKey("Facility", on_delete=models.PROTECT)
+    result = models.IntegerField(choices=SAMPLE_TEST_RESULT_CHOICES, default=SAMPLE_TEST_RESULT_MAP["Awaiting"])
```
afzalIbnSH: Should we map this to `PatientConsultation` too ?

anroopak: @bodhish @tomahawk-pilot  What are your thoughts? I am thinking of adding.

anroopak:
Updated PR: https://github.com/coronasafe/care/pull/160

Issue #103, opened 2020-03-25
Restrict users list to only superusers


afzalIbnSH:
Now all users can see all.

vigneshhari:
Just Username and Name and District , Nothing else, we might need this to add doctors to facility and staff to facility and so on later on

afzalIbnSH:
Okay

afzalIbnSH:
@tomahawk-pilot re-iterated "One user should not see other users" via slack. @vigneshhari Can you guys discuss and confirm what is to be done ?

afzalIbnSH:
@vigneshhari says;

> Normal users should see other users' id and name only, nothing else

Gonna do this

afzalIbnSH:
#159

PR #141, opened 2020-03-28
Refactor unit tests for Users API 

### LINKED ISSUE #106: Users API: Structure the unit tests
It's vaguely written. Refer https://github.com/coronasafe/care/blob/master/care/facility/tests/test_facility.py.

nabhanabdulla:
Fixes #106

All the unit tests of the Users API checks for a single functionality now. Couple of other scenarios for consideration are:

1. All the tests pass on running `pytest` but there's a warning being displayed at the end of the output
`/home/user/.local/lib/python3.6/site-packages/whitenoise/base.py:116: UserWarning: No directory at: /home/nabhanpv/Desktop/OpenSource/care/staticfiles/
        warnings.warn(u"No directory at: {}".format(root))`

2. Only refactoring of unit tests has been performed and all of the functionalities earlier tested for has retained. This includes the test for any user being able to read(list) other users as well. #103 seeks to restrict this to a certain type of users only. Should the tests be changed taking this into consideration once it's fixed or is it better to create a new issue to update the tests once #103 is resolved?

3. How about the tests are grouped into two classes TestSuperUser & TestNormalUser instead of a common TestUser class?


afzalIbnSH:
@nabhanabdulla 
- The warning seems to be related to location for static files. Can be ignored.
- Tests for #103 can be added later
- You don't need to test every functionality as done by super user/ normal user separately. Just do  `test_create`, `test_retrieve`, `test_update`, `test_destroy` and `test_list`. Do these with normal for normal user. Then test the special privileges a super user has i.e. `test_superuser_can_modify_others` and `test_superuser_can_delete_others`. Hope this helps. Also, as @abhiabhi94 correctly pointed out to me, these are integration tests rather than unit tests. You can maybe move this file to `care/users/tests/integration/test_users.py`. Abhyudai can then write actual unit tests in `care/users/tests/unit/test_users.py`

nabhanabdulla:
Ok, got it. I'll change the tests as suggested and move the file to the integration test folder

PR #145, opened 2020-03-28
Comment out Axes app to suppress warnings

agzuniverse:
Since Axes was removed temporarily in 1cf1123 , commenting out this line suppresses the warnings:

```
?: (axes.W002) You do not have 'axes.middleware.AxesMiddleware' in your settings.MIDDLEWARE.
?: (axes.W003) You do not have 'axes.backends.AxesBackend' or a subclass in your settings.AUTHENTICATION_BACKENDS.
	HINT: AxesModelBackend was renamed to AxesBackend in django-axes version 5.0.
```
Once Axes is integrated back in, this line can be uncommented to enable the app.

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -82,7 +82,7 @@ THIRD_PARTY_APPS = [
     "drf_extra_fields",
     "location_field.apps.DefaultConfig",
     "django_filters",
-    "axes",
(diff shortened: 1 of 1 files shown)


rejuls:
Causes issues with the migrations. Pls check in local.

PR #166, opened 2020-03-29
update Readme with instructions for setting up postgres and fix issues when creating superuser

ignujee:


biswaz:
Some feedback: 
PR name shouldn't be your name. 
Give a short description in the pr

farisshajahan:
Also if the changes are for an earlier PR, push your changes to the same branch as before, The changes will automatically pop up in the old PR

agzuniverse:
If we are going to give detailed instructions on setting up Postgres in the Readme itself, I feel it's probably better as a separate file. The currently linked stackoverflow answer in the Readme has most of these instructions already, but the `createsuperuser` part will be a valuable addition to our Readme right now.

Issue #175, opened 2020-03-29
Move tests from pytest to django unittest


abhiabhi94:
- `django` `unittest` seems to be more powerful for django based projects.
- Also, some of the tests written currently seem as integration tests. They should be transformed to unit tests.

vigneshhari:
Closing this issue now

Issue #112, opened 2020-03-26
Unit tests for the users application


abhiabhi94:
Write tests that try to cover models, forms, views and serializers for the users application

PR #177, opened 2020-03-29
Move tests from pytest to django unnittest

abhiabhi94:
Move all tests to `django` `unittest`

    Some tests that seemed integration tests have now be transformed to unit tests
    Some tests might fail because of the recent changes to the models

The modules used by `pytest` have not yet been deleted for others to view and compare current tests with the present one.
Once this request is ready to merge, we can delete them maybe.
More tests have to be added.
Addresses #175 and #112 #7


abhiabhi94:
I think we also need to maybe change the name of the file https://github.com/coronasafe/care/blob/master/config/settings/test.py since the command `python manage.py` is unable to detect all the tests from the base directory. I may be completely wrong.

vigneshhari:
@abhiabhi94 we can merge if you are okay with it , we will check each tests when we are free

abhiabhi94:
@vigneshhari Yes i am okay...just make sure I am not responsible for breaking something :-p

vigneshhari:
@abhiabhi94 i will merge the tests, but i changed Travis CI config to use PyTests , once the Django based tests are successful i will switch back, okays?

PR #190, opened 2020-03-30
fix patient sample and facility permissions

anroopak:


care/facility/api/viewsets/patient_sample.py
```diff
@@ -41,23 +43,18 @@ class PatientSampleViewSet(UserAccessMixin, viewsets.ModelViewSet):
...
-
     def get_serializer_class(self):
         serializer_class = self.serializer_class
         if self.action == "retrieve":
             serializer_class = PatientSampleDetailSerializer
         return serializer_class
 
+    def get_queryset(self):
+        queryset = super(PatientSampleViewSet, self).get_queryset()
+        if self.kwargs.get("patient_pk") is not None:
```
afzalIbnSH: Any reason this is not
`if not self.kwargs.get("patient_pk"):` ?

care/facility/api/viewsets/patient_sample.py
```diff
@@ -41,23 +43,18 @@ class PatientSampleViewSet(UserAccessMixin, viewsets.ModelViewSet):
...
-        Injects the patient data to serializer
-        """
-        try:
-            kwargs["data"]["patient"] = self.kwargs["patient_pk"]
-        except KeyError:
-            # In read serializers, "data" will not be present.
-            pass
-        return super().get_serializer(*args, **kwargs)
-
     def get_serializer_class(self):
```
afzalIbnSH:
```
            def get_serializer_class(self):
                if self.action == "retrieve":
                     return PatientSampleDetailSerializer
                return self.serializer_class
```
?

care/facility/api/viewsets/patient_sample.py
anroopak: Readability is more in this code. Both are executed same only.

care/facility/api/viewsets/patient_sample.py
anroopak: There is a convention that there should be only one `return` statement. 
https://hackerchick.com/religious-war-48293-single-vs-multiple/

care/facility/api/viewsets/patient_sample.py
afzalIbnSH: In my eyes it's redundant, unless we expect different set of values equating to `if not`, as in it could be either `None` or an empty list and we need to act differently on either cases. Personal preference I guess :)

care/facility/api/viewsets/patient_sample.py
afzalIbnSH: Bad example to quote, the author says he prefers multiple returns over multiple if else statements :D

PR #193, opened 2020-03-31
fix tests - facility - after moving to unittests

anroopak:


care/utils/tests/test_base.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+class TestBase(APITestCase):
+    """
+    Base class for tests, handles most of the test setup and tools for setting up data
+    """
+
+    @classmethod
+    def create_user(cls, district: District, username: str = "user", **kwargs):
+        username = kwargs.get("username", username)
```
abhiabhi94: `username` will either be passed as a keyword-based argument or not. Should we expect both?

anroopak: username will have a default value of "user". If nothing is passed, this default will be used. If it is passed, then the passed value will be used.

PR #206, opened 2020-04-01
add all query param to get facility list by every user

### LINKED ISSUE #171: Enhance Facility list API
Let the `GET api/v1/facility` API accept a couple of query params, `search_text` and `all`.

- `all` -> Boolean, if true should return every facility to all logged in users with just their id, name and district
- `search_text` -> str, Search on the facility's properties.

Make sure the API is paginated

sandeepsajan0:
Added a check in Viewset for `all` query param and return list of active facilities available with id, name, district and state fields only.
Usage:
`api/v1/facility?all=true` get request with a valid token for authorization.
Fixed #171


vigneshhari:
There is a separate endpoint to get all the users currently.

Issue #210, opened 2020-04-01
Add Admin Panel Filters


vigneshhari:
Add filters for district/local body/state to all models that contain those fields. Add searchable foreign key fields as well, since there are going to be many objects in the foreign key selection dropdown

sandeepsajan0:
I am working on it.

vigneshhari:
Since this is a production Issue, Please try to solve the issue by tonight itself.

sandeepsajan0:
https://github.com/coronasafe/care/pull/213#discussion_r401542513 need to add filters for every models there like Facility staff, Inventory, etc?

sandeepsajan0:
I think it is not completed yet?

PR #212, opened 2020-04-01
Add state name and id filtering to facility

agzuniverse:
Related issue: #210


vigneshhari:
The issue was, adding filters to admin panel , not to the existing views @agzuniverse

vigneshhari:
@sandeepsajan0  is working on that

PR #213, opened 2020-04-01
add admin filter for Facility model

### LINKED ISSUE #210: Add Admin Panel Filters

sandeepsajan0:
Added the admin panel filters for the Facility model.
Fixes #210

### DIFF
--- a/care/facility/admin.py
+++ b/care/facility/admin.py
@@ -1,4 +1,5 @@
 from django.contrib import admin
+from django.contrib.admin import SimpleListFilter
 
 from care.facility.models.patient_sample import PatientSample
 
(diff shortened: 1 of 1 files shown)


care/facility/admin.py
```diff
@@ -25,8 +26,57 @@ class BuildingAdmin(admin.ModelAdmin):
...
+            return queryset
+        return queryset.filter(state__name=self.value())
+
+
 class FacilityAdmin(admin.ModelAdmin):
     search_fields = ["name"]
+    list_filter = [DistrictFilter, LocalBodyFilter, StateFilter]
 
 
 class FacilityStaffAdmin(admin.ModelAdmin):
```
sandeepsajan0, line 82: need to add similar filters with every model here?

PR #218, opened 2020-04-02
Fix tests

afzalIbnSH:
@anroopak @vigneshhari @biswaz I have made sure all the tests run with pytest pass and configured travis to run only those tests. Let's make sure these tests don't break going forward and only merge PRs that doesn't. Meanwhile @abhiabhi94 shall fix the tests written with django unit tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,4 +8,4 @@ addons:
       - libsqlite3-mod-spatialite
 install:
   - pip install -q -r requirements.txt
-script: pytest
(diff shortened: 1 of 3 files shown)


afzalIbnSH:
@vigneshhari We now have circle ci too? Why do we have both travis and circle ci?. Anyways I couldn't find the code for circle ci in our repo. Can you change the test command from `pytest` to `pytest -v care/facility/tests/test_facility.py care/facility/tests/test_patient.py care/users/tests/test_users.py`

Issue #211, opened 2020-04-01
Fix failing tests


anroopak:
There are quite a lot of tests which are failing because we moved from pytest to UnitTests. 
Also there were a lot of changes in the application by the time we ported. 
Need someone to work on this.

afzalIbnSH:
I think Abhyudai is already working on this. @abhiabhi94 Can you push your PR against this issue. The earlier you can the merrier.

abhiabhi94:
Till the time, all texts are fixed, I would suggest to run only `pytest` tests. We can add some sort of regex when running tests that does not run file ending with `_api.py`. There is a file `test_models.py` inside `care/users/` while also need not be run currently. The above argument is based upon the assumption that none else apart from me has added tests based on `django` `unittest`.

abhiabhi94:
I think this can be closed now.

PR #222, opened 2020-04-02
Test for facility api are passing as of when this commit was made

abhiabhi94:
Some tests from patient history API are failing(3 of them). If anyone of you among  @vigneshhari @anroopak @afzalIbnSH or someone else has made the changes, please ask them to change the file `care/facility/tests/test_facility_patient_history_api.py` accordingly.

Addresses #211 partially


care/facility/tests/test_facility_api.py
```diff
@@ -206,86 +228,140 @@ def test_facility_list_retrieval(self):
...
-            location=Point(24.452545, 49.878248),
-            oxygen_capacity=10,
-            phone_number="9998887776",
-        )
 
-    def setUp(self):
-        """This is run before every class method"""
-        self.client.login(username=self.su_data["username"], password=self.su_data["password"])
+    def get_detail_representation(self, facility: Facility) -> dict:
+        return {
```
anroopak, line 269: I don't think these were required in this class because it is not used. You could have just did a `pass`.

PR #228, opened 2020-04-02
add details to patient regn, user models

anroopak:


care/facility/models/patient_data.py
```diff
@@ -85,6 +84,13 @@ class PatientRegistration(models.Model):
 
     disease_status = models.IntegerField(choices=DISEASE_STATUS_CHOICES, default=1, blank=True)
 
+    number_of_aged_dependents = models.IntegerField(
+        default=0, verbose_name="Number of people aged above 60 living with the patient"
+    )
+    number_of_chronic_diseased_dependents = models.IntegerField(
+        default=0, verbose_name="Number of people who have chronic diseases living with the patient"
+    )
+
```
abhiabhi94, line 93: Shouldn't the appropriate changes be made also for the `tests`?

anroopak: Once we port to unittests, we will fix these. I am spending some time this weekend to fix the tests. after that no bypassses allowed

PR #237, opened 2020-04-03
add search_text param with search vector and search rank

### LINKED ISSUE #171: Enhance Facility list API

sandeepsajan0:
Add a search vector on Facility `name`, `district__name`, `state__name` and annotate result with search rank. 
Added `seach_text` query param with endpoint `api/v1/facility`
Fixes #171

### DIFF
--- a/care/facility/api/viewsets/facility.py
+++ b/care/facility/api/viewsets/facility.py
@@ -1,3 +1,4 @@
+from django.contrib.postgres.search import SearchQuery, SearchRank, SearchVector
 from django.db import transaction
 from django_filters import rest_framework as filters
 from dry_rest_permissions.generics import DRYPermissionFiltersBase, DRYPermissions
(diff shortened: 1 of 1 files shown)


anroopak:
@sandeepsajan0 Your comments on this 

https://docs.djangoproject.com/en/3.0/ref/contrib/postgres/search/#searchvectorfield
May be a cron which runs every 1 hours, or a trigger which can run only when the fields of search vector changes to rebuild the search vector index.

We can take this up in a second issue. Need not be part of this issue / PR.

afzalIbnSH:
I'll merge this and fix the build failure in a separate PR.

PR #248, opened 2020-04-03
Allow null entries for countries travelled field in patients model

agzuniverse:
This fixes a failing test.


care/facility/models/patient_data.py
```diff
@@ -92,7 +92,9 @@ class PatientRegistration(models.Model):
     past_travel = models.BooleanField(
         default=False, verbose_name="Travelled to Any Foreign Countries in the last 28 Days"
     )
-    countries_travelled = models.TextField(default="", blank=True, verbose_name="Countries Patient has Travelled to")
+    countries_travelled = models.TextField(
+        default="", null=True, blank=True, verbose_name="Countries Patient has Travelled to"
```
abhiabhi94, line 96: why is default value required, when you have already set `blank=True` for form and `null=True` for database.

anroopak:
@agzuniverse Cant we change something so that null will be handled as empty string?

agzuniverse:
@anroopak I guess the seralizer can be changed so a `null` in the request will be changed to empty string, or we can just stop sending the `contries_travelled` field in the req so the default value gets stored as `""`.

But is there any advantage of storing that value as an empty string over storing it as `null`?

care/facility/models/patient_data.py
agzuniverse: Thanks, I've removed the default field and updated the commit.

abhiabhi94:
> @agzuniverse Cant we change something so that null will be handled as empty string?

This is what `null=True` does to the database?
https://docs.djangoproject.com/en/3.0/ref/models/fields/#null

care/facility/models/patient_data.py
abhiabhi94: I think you will have to migrate again for the change to be transferred to the database. I can't see the new migration file

agzuniverse:
@anroopak do merge this if there are no issues, the migration may cause problems if someone else introduces changes to the model in the meantime

vigneshhari:
@agzuniverse in the migrations file, i still see the default values as "" . So is there any advantage for changing the code?

agzuniverse:
@vigneshhari sorry my bad, I've fixed the migration now.

PR #249, opened 2020-04-04
Fixes all facility api tests

abhiabhi94:


care/facility/tests/test_facility_api.py
```diff
@@ -91,7 +91,13 @@ def test_user_can_create_facility(self):
         self.assertEqual(response.status_code, status.HTTP_201_CREATED)
         # test response data
         self.assertDictEqual(
-            response.json(), {**self.get_detail_representation(facility_data), "id": mock_equal},
+            response.json(),
+            {
+                **self.get_detail_representation(facility_data),
+                "id": mock_equal,
+                "modified_date": mock_equal,
```
anroopak, line 98: @abhiabhi94 can't we move the `created_date` and `modified_date` to `get_detail_representation` ?

abhiabhi94: no, because when trying to send a `POST` request to the API, it wants the data to be in `JSON` format. The `mock_equal` object is not `JSON` serializable.
The function that this will affect is `test_superuser_can_update_ones_facility`.

PR #251, opened 2020-04-04
Fixing patient tests

agzuniverse:
Requires #248 to be merged first.


agzuniverse:
Seems like the tests are fixed now, so this PR is no longer needed.

Issue #253, opened 2020-04-04
List patients api should return facility name


bodhish:
Its hard to find facility from patient list for district admin

PR #259, opened 2020-04-04
#253 Facility name in patient list

SarathMK:
Fix for #253
Added `facility_name` in patient list API


vigneshhari:
A facility object has been added to the response already

PR #277, opened 2020-04-06
Fix choices

abhiabhi94:
#### Moves all choices to enum
- create a base class that just needs to be passed the `dict` of choices.
- the base class also has a function for generating the list of choice tuples to be used for models.

##### Advantages
- easy to use
- the function for generating a list of the choice tuples tries to stop duplication of the same logic.
- other people don't need to remember or look-up mapping for choices when writing code.

#### Remove a bit of duplication of code in the test for users API

Also fixes #254

PR #282, opened 2020-04-08
Revert "Fix choices"

anroopak:
Reverts coronasafe/care#277

@abhiabhi94 I dont think the samples flow, etc has been touched in this change. Reverting as it is causing issues.


abhiabhi94:
@anroopak Do you mean this function? https://github.com/coronasafe/care/blob/master/care/facility/models/patient_sample.py#L76
Is there something else that needs to change?
Sorry, I will fix it and it also means we need to add tests for this part.

PR #416, opened 2020-06-26
fixes the detailed consultation view for district admin

Ajuajmal:
#### Error
----
Screenshot from 2020-06-27 04-08-26
717dd4b0f.png)

### Fix
----

- for the object `PatientConsultation` does not have `district` and `state`  attributes 
- but, by using the `Patient` attribute we can get the values for above
- so used used `Patient` attribute for accessing the above values

#### District Admin - Detailed Patient Consultation View Before this commit 
----
Screenshot from 2020-06-27 03-12-02

#### District Admin - Detailed Patient Consultation View After this commit
----
Screenshot from 2020-06-27 03-14-33


vigneshhari:
Good Work :)

PR #530, opened 2021-04-19
avoid investigation update in closed consultation

divyagar:


anroopak (commented):
Implementation looks good, but always add a description to the PR so that the reviewer knows what are the changes being implemented.

vigneshhari (approved):

PR #533, opened 2021-04-20
initial commit for post covid form.

divyagar:


care/facility/api/serializers/post_covid_data.py
```diff
@@ -0,0 +1,176 @@
...
+    IV = "IV"
+    ORAL = "Oral"
+
+
+ModeOfTransmissionChoices = [(e.value, e.name) for e in ModeOfTransmission]
+
+
+# Validator serializer for vitals_at_admission
+class VitalsAtAdmissionSerializer(serializers.Serializer):
+    pr = serializers.IntegerField(allow_null=True)
```
anroopak, line 19: Would it be better to come up with full names? BP is a common term, but names like `pr`, `rr` are not common terms and we cannot expect the next developers to understand the meaning.

care/facility/api/viewsets/post_covid_data.py
```diff
@@ -0,0 +1,79 @@
...
+        elif self.request.user.user_type >= User.TYPE_VALUE_MAP["StateLabAdmin"]:
+            return self.queryset.filter(patient__facility__state=self.request.user.state)
+        elif self.request.user.user_type >= User.TYPE_VALUE_MAP["DistrictLabAdmin"]:
+            return self.queryset.filter(patient__facility__district=self.request.user.district)
+        filters = Q(patient__facility__users__id__exact=self.request.user.id)
+        filters |= Q(assigned_to=self.request.user)
+        return self.queryset.filter(filters).distinct("id")
+
+    def create(self, request, *args, **kwargs):
+        print("create function")
```
anroopak, line 33: remove the print statement.

care/facility/api/viewsets/post_covid_data.py
```diff
@@ -0,0 +1,79 @@
...
+        with transaction.atomic():
+            if kwargs.get("patient_external_id") is None:
+                raise ValidationError({"error": "Patient id should be provided"})
+
+            request.data["patient_id"] = PatientRegistration.objects.get(
+                external_id=kwargs.get("patient_external_id")).id
+            patient = PatientRegistration.objects.get(id=request.data["patient_id"])
+
+            # data = {
```
anroopak, line 42: Remove unwanted comments.

care/facility/api/viewsets/post_covid_data.py
```diff
@@ -0,0 +1,79 @@
...
+            request.data["consultation"] = consultation.id
+
+            facilities = []
+            for id in request.data["treatment_facility"]:
+                facilities.append(Facility.objects.get(external_id=id).id)
+
+            request.data["treatment_facility"] = facilities
+            return super(PostCovidDataViewSet, self).create(request, *args, **kwargs)
+
+    # def perform_create(self, serializer):
```
anroopak, line 61: Remove unwanted comments.

care/facility/models/post_covid_data.py
```diff
@@ -0,0 +1,206 @@
+from django.db import models
+from django.contrib.postgres.fields import JSONField, ArrayField
+from care.facility.models.base import BaseModel
+from care.facility.models.patient_consultation import PatientConsultation
+import enum
+
+
+class PostCovidTime(enum.Enum):
```
anroopak, line 8: Why should we use this? Can't we just use an integer? Like `number_of_weeks`.

care/facility/models/post_covid_data.py
```diff
@@ -0,0 +1,206 @@
...
+    UNDER_THREE_WEEKS = 1
+    THREE_TO_TWELVE = 2
+    OVER_TWELVE_WEEKS = 3
+
+
+PostCovidTimeChoices = [(e.value, e.name) for e in PostCovidTime]
+
+
+class CovidCategory(enum.Enum):
+    MILD = 1
```
anroopak, line 18: Better to use string names - will be helpful for the Frontend devs.

care/facility/models/post_covid_data.py
```diff
@@ -0,0 +1,206 @@
...
+        null=True
+    )
+    date_of_onset_symptoms = models.DateField(
+        default=None,
+        null=True,
+        verbose_name="Date of first appearance of symptoms"
+    )
+    post_covid_time = models.IntegerField(
+        choices=PostCovidTimeChoices,
+        default=1,
```
anroopak, line 64: Becomes difficult to understand and hence debug. Use `PostCovidTime.UNDER_THREE_WEEKS`. I had difficulty finding the value for `1` while review, we cannot expect the next developers to know the number values and use it.

care/facility/models/post_covid_data.py
```diff
@@ -0,0 +1,206 @@
...
+    steroids = models.BooleanField(default=False, null=True)
+    steroids_drugs = ArrayField(
+        JSONField(default=dict, null=False),
+        default=list,
+        null=True,
+        blank=True
+    )
+
+    anticoagulants = models.BooleanField(default=False, null=True)
+    anticoagulants_drugs = ArrayField(
```
anroopak, line 137: We could just use `JSONField(default=list)`, instead of using an ArrayField and a JSON field inside that.

anroopak (changes requested):

vigneshhari:
@divyagar please resolve all the issues pointed out and fix the merge conflicts.

PR #535, opened 2021-04-25
Made requested changes

divyagar:
#533


care/facility/api/serializers/post_covid_data.py
```diff
@@ -0,0 +1,175 @@
...
+            "icu_admission",
+            "oxygen_requirement",
+            "probable_diagnosis"
+        ]
+        extra_kwargs = {i: {'required': True} for i in required_fields}
+
+    def validate_vitals_at_admission(self, vitals_at_admission):
+        if vitals_at_admission == None:
+            return vitals_at_admission
+
```
abhiabhi94, line 94: You may probably replace this with:
```suggestion
        if vitals_at_admission is not None:
```
and similarly for other validators as well.

care/facility/api/viewsets/post_covid_data.py
```diff
@@ -0,0 +1,53 @@
...
+            return self.queryset.filter(patient__facility__state=self.request.user.state)
+        elif self.request.user.user_type >= User.TYPE_VALUE_MAP["DistrictLabAdmin"]:
+            return self.queryset.filter(patient__facility__district=self.request.user.district)
+        filters = Q(patient__facility__users__id__exact=self.request.user.id)
+        filters |= Q(assigned_to=self.request.user)
+        return self.queryset.filter(filters).distinct("id")
+
+    def create(self, request, *args, **kwargs):
+        with transaction.atomic():
+            if kwargs.get("patient_external_id") is None:
```
abhiabhi94, line 34: You probably want
```suggest
if kwargs.get("patient_external_id", None) is None:
```

care/facility/api/viewsets/post_covid_data.py
```diff
@@ -0,0 +1,53 @@
...
+            patient = PatientRegistration.objects.get(id=request.data["patient_id"])
+
+            consultation = PatientConsultation(facility_id=patient.facility.id, patient_id=patient.id)
+            consultation.save()
+
+            request.data["consultation"] = consultation.id
+
+            facilities = []
+            for id in request.data["treatment_facility"]:
+                facilities.append(Facility.objects.get(external_id=id).id)
```
abhiabhi94, line 48: This could be probably replaced by a leaner `queryset` and list comprehension which is computationally faster.
```suggestion
            facilities = [
                Facility.objects.get(external_id=i).values('id")[0] 
                for i in request.data["treatment_facility"
            ]
```

abhiabhi94 (commented):
In additon to the other suggestions, it is generally advisable to use the `gettext_lazy` function when sending/displaying raw strings to the user. This enables those strings to be marked and used if and when the translation is available.

```python
from django.utils.translation import gettext_lazy as _

verbose_name = _('This is my verbose name')
```

care/facility/api/viewsets/post_covid_data.py
```diff
@@ -0,0 +1,53 @@
...
+        return self.queryset.filter(filters).distinct("id")
+
+    def create(self, request, *args, **kwargs):
+        with transaction.atomic():
+            if kwargs.get("patient_external_id") is None:
+                raise ValidationError({"error": "Patient id should be provided"})
+
+            request.data["patient_id"] = PatientRegistration.objects.get(
+                external_id=kwargs.get("patient_external_id")).id
+            patient = PatientRegistration.objects.get(id=request.data["patient_id"])
```
afzalIbnSH, line 39: Why are there 2 queries here ? It should just be `patient = PatientRegistration.objects.get(external_id=kwargs["patient_external_id"])`

care/facility/api/viewsets/post_covid_data.py
```diff
@@ -0,0 +1,53 @@
...
+            consultation = PatientConsultation(facility_id=patient.facility.id, patient_id=patient.id)
+            consultation.save()
+
+            request.data["consultation"] = consultation.id
+
+            facilities = []
+            for id in request.data["treatment_facility"]:
+                facilities.append(Facility.objects.get(external_id=id).id)
+
+            request.data["treatment_facility"] = facilities
```
afzalIbnSH, line 50: Just do
```
request.data["treatment_facility"] = Facility.objects \
    .values_list('id', flat=True) \
    .filter(external_id__in=request.data["treatment_facility"])
```
More than being simpler, this does just one query instead of how you're currently making a query each.

afzalIbnSH (changes requested):

care/facility/api/viewsets/post_covid_data.py
afzalIbnSH, line 50: Do a single query by using `filter`. Use `.values_list('id', flat=True)` to let django take care of giving you just ids as a flattened queryset. Converting it to a standard Python list with `list()` ensures there are no serialization problems (Just to be safe, may not be necessary).
```suggestion
            request.data["treatment_facility"] = list(
                Facility.objects
                    .values_list('id', flat=True)
                    .filter(external_id__in=request.data["treatment_facility"])
            )
```

divyagar: Done

care/facility/api/viewsets/post_covid_data.py
divyagar: done

care/facility/api/serializers/post_covid_data.py
divyagar: done

PR #540, opened 2021-04-26
add data and script to load state, districts data

anroopak:
Adds the State-District JSON and a management command to load the data.

### DIFF
--- a/care/users/management/commands/load_state_data.py
+++ b/care/users/management/commands/load_state_data.py
@@ -1,0 +1,44 @@
+import csv
+import json
+import os
+from typing import Optional
+
+from django.core.management import BaseCommand, CommandParser
+
+from care.users.models import District, State
+
+states_to_ignore = [s.lower() for s in ["Kerala", "Lakshadweep (UT)"]]
+
+
+class Command(BaseCommand):
+    """
+    Management command to load state and district data from JSON
+    Usage: python manage.py load_state_data ./data/india/states-and-districts.json
+    """
+
+    help = "Loads Local Body data from a folder of JSONs"
+
+    def add_arguments(self, parser: CommandParser) -> None:
+        parser.add_argument("json_file_path", help="path to the folder of JSONs")
+
+    def handle(self, *args, **options):
+        json_file_path = options["json_file_path"]
+
+        data = []
+        with open(json_file_path, "r") as json_file:
+            data = json.load(json_file)
+
+        for item in data:
+            state_name = item["state"].strip()
+            if state_name.lower() in states_to_ignore:
+                print(f"Skipping {state_name}")
+                continue
+
+            districts = [d.strip() for d in item["districts"].split(",")]
+
+            state, is_created = State.objects.get_or_create(name__iexact=state_name, defaults={"name": state_name})
+            print(f"{state_name} - {is_created}")
+
+            for d in districts:
+                _, is_created = District.objects.get_or_create(state=state, name__iexact=d, defaults={"name": d})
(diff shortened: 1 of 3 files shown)


care/users/management/commands/load_state_data.py
```diff
@@ -0,0 +1,44 @@
...
+        json_file_path = options["json_file_path"]
+
+        data = []
+        with open(json_file_path, "r") as json_file:
+            data = json.load(json_file)
+
+        for item in data:
+            state_name = item["state"].strip()
+            if state_name.lower() in states_to_ignore:
+                print(f"Skipping {state_name}")
```
afzalIbnSH, line 34: We don't have a logger setup ?

care/users/management/commands/load_state_data.py
```diff
@@ -0,0 +1,44 @@
...
+                continue
+
+            districts = [d.strip() for d in item["districts"].split(",")]
+
+            state, is_created = State.objects.get_or_create(name__iexact=state_name, defaults={"name": state_name})
+            print(f"{state_name} - {is_created}")
```
afzalIbnSH, line 40: Something like `f"{'Created' if is_created else 'Retrieved'} {state_name}"` could potentially be more helpful. It'll log "Created Kerala" for instance.

care/users/management/commands/load_state_data.py
```diff
@@ -0,0 +1,44 @@
...
+
+            for d in districts:
+                _, is_created = District.objects.get_or_create(state=state, name__iexact=d, defaults={"name": d})
+                print(f"{d} - {is_created}")
```
afzalIbnSH, line 44: Same as above

setup.cfg
```diff
@@ -26,4 +26,4 @@ django_settings_module = config.settings.test
 ignore_errors = True
 
 [isort]
-known_third_party = allauth,celery,crispy_forms,dateparser,dateutil,django,django_filters,django_rest_passwordreset,djangoql,djqscsv,drf_extra_fields,drf_yasg,dry_rest_permissions,environ,fernet_fields,freezegun,hardcopy,location_field,multiselectfield,partial_index,phonenumber_field,phonenumbers,pytz,ratelimit,requests,rest_framework,rest_framework_nested,rest_framework_simplejwt,sentry_sdk,simple_history
+known_third_party = allauth,boto3,celery,crispy_forms,dateparser,dateutil,django,django_filters,django_rest_passwordreset,djangoql,djqscsv,drf_extra_fields,drf_yasg,dry_rest_permissions,environ,fernet_fields,freezegun,hardcopy,location_field,multiselectfield,partial_index,phonenumber_field,phonenumbers,pytz,pywebpush,ratelimit,requests,rest_framework,rest_framework_nested,rest_framework_simplejwt,sentry_sdk,simple_history
```
afzalIbnSH, line 29: Looks like this PR adds `boto3 ` and `pywebpush` but it's not clear why.

vigneshhari: we use both these libraries now

care/users/management/commands/load_state_data.py
vigneshhari: We don't store any kind of logs as of now, We can start that if needed.

care/users/management/commands/load_state_data.py
vigneshhari: Will Change it to that

care/users/management/commands/load_state_data.py
vigneshhari: Will Change it to that

PR #541, opened 2021-04-27
Remove recovery from patient disease status

divyagar:
Please check [this](https://github.com/coronasafe/care_fe/issues/988) issue for information.


care/facility/models/patient_base.py
```diff
@@ -72,9 +72,8 @@ class DiseaseStatusEnum(enum.IntEnum):
     SUSPECTED = 1
     POSITIVE = 2
     NEGATIVE = 3
-    RECOVERY = 4
-    RECOVERED = 5
-    EXPIRED = 6
+    RECOVERED = 4
+    EXPIRED = 5
```
vigneshhari, line 76: Never use an old choice integer for something else, use another integer

care/facility/management/commands/clean_patient_disease_status.py
```diff
@@ -0,0 +1,29 @@
...
+            try:
+                patient.disease_status = new
+                patient.save()
+            except Exception:
+                failed.append({"id": patient.id, "disease_status": patient.disease_status})
+
+            print(patient.disease_status)
+    return failed
+
+
```
vigneshhari, line 20: we can just do patient.filter(disease_status=5).update(disease_status=4) right ?

vigneshhari (commented):
Changes Required

care/facility/management/commands/clean_patient_disease_status.py
divyagar: > we can just do patient.filter(disease_status=5).update(disease_status=4) right ?

Yes we can but then we will not know updation of which rows are failed.

vigneshhari: there will be one single mysql query, what failure are we talking about here ?

vigneshhari (changes requested):
Changes Requested

abhiabhi94:
Hey @vigneshhari, I think this change to 'RECOVERY' to `RECOVERED` should ideally be handled by a custom migration. This way, we won't necessarily have to run the command at every end(production as well as for all developers).

If you agree, I can do that and then this one only needs to handle the addition of another value to the model.

vigneshhari:
@abhiabhi94 you can take over this issue then, @divyagar after he is done with his work review it.

vigneshhari:
@abhiabhi94 close this PR when you start working on the issue

abhiabhi94:
Do we need to add an extra field `EXPIRED`?

vigneshhari:
https://github.com/coronasafe/care_fe/issues/988
This is the issue

PR #543, opened 2021-04-27
Change disease status from recovery to recovered

abhiabhi94:
- inside the PatientRegistration model.

closes #541

### DIFF
--- a/care/facility/migrations/0224_change_disease_status_from_recover_to_recovered.py
+++ b/care/facility/migrations/0224_change_disease_status_from_recover_to_recovered.py
@@ -1,0 +1,17 @@
+from django.db import migrations, models
+
+from care.facility.models.patient_base import DiseaseStatusEnum
+
+
+def change_recovery_to_recovered(apps, schema_editor):
+    PatientRegistration = apps.get_model('facility', 'PatientRegistration')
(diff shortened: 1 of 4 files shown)


vigneshhari:
The Recovery Option should be deleted as well.

vigneshhari:
Let's close the other PR and move all the changes here.

abhiabhi94:
> The Recovery Option should be deleted as well.

If we remove the `RECOVER` option, how will the migrations and the test use that value? I think we can probably alias them through a method inside the model's `save` method maybe?

vigneshhari:
@abhiabhi94 we can remove it after the migration is performed right ? as another migration?

abhiabhi94:
> @abhiabhi94 we can remove it after the migration is performed right ? as another migration?

not really, because it will create issues at other places when the migration is run. If the value `RECOVERED` is not present inside the codebase, when the migrations is run at some other place(consider the migration was first run successfully on my end and then, after removing the thing,  you too try to run it at your end), it will throw an error.

I think aliases it should be sufficiently enough for this use case.

abhiabhi94:
The tests for the current change are passing as of now at my end. 

`black`, `isort`, `autoflake` etc. hooks seem to have made some changes that might make the `diff` a bit more than it should have been.

If this look good enough, feel free to merge.

PR #550, opened 2021-04-29
Add instruction for setting up using docker.

abhiabhi94:
- Move all instructions for setting up the project to CONTRIBUTING.rst


abhiabhi94:
I couldn't think of a way to run the `pre-commit` hooks from the docker container without installing `pre-commit` on the host machine.

Rest everything else can be setup using `docker`.

Any suggestions/ideas are welcome.

PR #571, opened 2021-05-10
Add created by field in daily rounds for telemedicine purpose

divyagar:


divyagar:
@vigneshhari Can a telemedicine doctor update the consultation created by another telemedicine doctor? If yes, then we would have to set flag on the basis of `facility of doctor` instead of `assignment`.

vigneshhari:
there can be only one telemedicine doctor assigned at any point in time. always check if the current logged in user is the assigned user, that logic should be enough.

divyagar:
`add a flag for consultation and Daily Round and cascade changes in the daily round into the patient consultation model as well.`.
@vigneshhari  Can you clarify what you meant here?

PR #572, opened 2021-05-11
Add username validation

divyagar:
Fixes coronasafe/care_fe#1031


vigneshhari (approved):
I am assuming you tested this.

PR #577, opened 2021-05-12
Change offset to page number in paginated apis

divyagar:
Fixes coronasafe/care_fe#1045

The default page size is 15.
Some of the classes have different page size. I have overridden page size of following classes:
Facility - 14
Patient consultation - 5.
test_sample - 10 (can be changed using page_size parameter. Ex - `http://localhost:8000/api/v1/patient/:patient_id/test_sample/?page_size=5&page=2` ).

I have looked at all the pages. Still, if I have missed any route please let me know.

### DIFF
--- a/care/facility/api/viewsets/facility.py
+++ b/care/facility/api/viewsets/facility.py
@@ -10,6 +10,7 @@ from rest_framework.decorators import action
 from rest_framework.permissions import IsAuthenticated
 from rest_framework.response import Response
 from simple_history.utils import bulk_create_with_history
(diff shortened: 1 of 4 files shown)


vigneshhari:
Page Number pagination was already supported with the old version. The front only needed to be implemented

divyagar:
Can you enable the page number pagination then? I would work on frontend accordingly.

vigneshhari:
@divyagar the backend already supports page number based pagination, nothing has to be changed here.

PR #597, opened 2021-06-24
Update test_patient_consultation_api.py

albernsrya:
#BUG RISK
Multiple keys
expression binds the same key multiple times. This may lead to unintentional behaviour as the key will always map the last value provided to it. It is recommended to have unique keys.


vigneshhari:
The Tests are all broken hence we are in the process of rewriting the whole thing once again.

PR #617, opened 2021-09-09
Add is_antenatal to filters and csv export

Apurva-Nagar:
- Allows Shifting requests to be filtered based on the is_antenatal parameter.
- Added is_antenatal information to CSV export
- Added is_antenatal to Patient API filter set


care/facility/api/viewsets/shifting.py
```diff
@@ -69,6 +69,7 @@ class ShiftingFilterSet(filters.FilterSet):
     assigned_to = filters.NumberFilter(field_name="assigned_to__id")
     created_by = filters.NumberFilter(field_name="created_by__id")
     last_edited_by = filters.NumberFilter(field_name="last_edited_by__id")
+    is_antenatal = filters.BooleanFilter(field_name="patient__is_antenatal")
```
vigneshhari, line 72: Add a default to None , rather than false , also create the required migration files, add a filter to the list api as well

vigneshhari (changes requested):

care/facility/api/viewsets/shifting.py
Apurva-Nagar: Can you please specify which list API you are referring to? Thanks.

vigneshhari: The shifting list API

bodhish:
@vigneshhari waiting for your review 🎉

vigneshhari (approved):

PR #619, opened 2021-09-14
Blockwise external result

shushil-kumar:
changes involve:
1. creating the requires model, filterset and viewset for Block
2. modifying user and LocalBody models to include Block
3. adding a new user type "BlockAdmin"
4. adding required validations for the creation of BlockAdmin
5. modification to the logic of ExternalResults to show Block level data for BlockAdmins
6. modifications to the load_data script to load Block data as well
7. creation of blocks_balia.json to store block data


vigneshhari:
Closing PR since features are not likely to be merged in ( too many conflicts )

PR #640, opened 2021-10-29
Asset Model: added camera asset type and camera details JSON field

Apurva-Nagar:
Modified the Asset Model to add Camera as one of the Asset Types and added field camera in meta field to store details about the camera (URL, preset) in JSON

Changes related to issue https://github.com/coronasafe/care_fe/issues/1942 on Care FE


care/facility/models/asset.py
```diff
@@ -42,6 +45,8 @@ class Status(enum.Enum):
     support_name = models.CharField(max_length=1024, blank=True, null=True)
     support_phone = models.CharField(max_length=14, validators=[phone_number_regex], default="")
     support_email = models.EmailField(blank=True, null=True)
+    # Camera Details
+    camera = JSONField(default=dict, validators=[JSONFieldSchemaValidator(CAMERA_DETAILS)])
```
vigneshhari, line 49: We already have a meta JSON field in this model right? we can reuse that instead of having type-specific attributes.

vigneshhari (changes requested):

care/facility/models/asset.py
gigincg: @vigneshhari What about validation? Is there a possibility of having polymorphic validation to the meta field, based on asset type or so?

vigneshhari: the `meta` json can have an optional key `camera` with a given data structure which can be enforced, we can go one level deep by overriding the `save` function and adding validation to only save when the `camera` key is present in `meta` as well.

care/facility/api/serializers/asset.py
```diff
@@ -53,7 +51,19 @@ def validate(self, attrs):
...
 
+    def create(self, validated_data):
+        camera_details = validated_data["meta"]["camera"]["url"] and validated_data["meta"]["camera"]["preset"]
+
+        if validated_data["asset_type"] == Asset.AssetType.CAMERA.value and not camera_details:
+            raise ValidationError({"Camera Details": "Cannot be blank when asset type is Camera"})
+        return super().create(validated_data)
+
     def update(self, instance, validated_data):
+        camera_details = validated_data["meta"]["camera"]["url"] and validated_data["meta"]["camera"]["preset"]
```
vigneshhari, line 62: since the `camera` attribute is specific to the asset type camera, there can be cases where the key `camera` is not 
 present, this will produce a key error, move the check inside the if block or use the built-in get method to get the value or add an exception handling block.

care/facility/models/json_schema/asset.py
```diff
@@ -1,6 +1,6 @@
-CAMERA_DETAILS = {
+META = {
     "$schema": f"http://json-schema.org/draft-07/schema#",
     "type": "object",
-    "properties": {"url": {"type": "string"}, "preset": {"type": "number"}},
+    "properties": {"camera": {"url": {"type": "string"}, "preset": {"type": "number"}}},
```
vigneshhari, line 4: JSON schema has an option to keep the camera key as optional, mark the camera field as optional.

vigneshhari (changes requested):

vigneshhari:
Merge all migrations as well, we only need the migration for the field change, we don't need the migrations that create and delete another field.

care/facility/api/serializers/asset.py
vigneshhari: Since we are repeating the same validation in create and update, it also makes sense to make it into a function and abstracting it out.

vigneshhari (changes requested):

care/facility/models/json_schema/asset.py
Apurva-Nagar: The latest draft of JSON schema only seems to have `required` - https://json-schema.org/draft/2020-12/json-schema-validation.html#rfc.section.6.5.3

Aren't non-required keys optional by default?

vigneshhari: My bad! I thought it was the other way around.

vigneshhari:
Merge the migrations and we should be good to merge this PR.

PR #645, opened 2021-11-13
Fix : Patient CSV export duration check

### LINKED ISSUE #644: Patient CSV Export only works for `created_date`
**Describe the bug**
1. In Patient CSV Exports, A `created_date` filter is required with any export request.
2. Only the `created_date` field value is checked for 7 day export limit, other date fields are ignored.

**To Reproduce**
1. Export Patients with `date_declared_positive` as any window less than 7 days, with and without `created_date` filter applied.

**Expected behavior**
1. Any of the set date filters (at least one) should be accepted.
2. All filters should be checked to stay within 7 day period.

shaileshaanand:
fixes #644 

Fix a bug where only `created_date` was checked for 7 day restriction and other fields were ignored.

### DIFF
--- a/care/facility/api/viewsets/patient.py
+++ b/care/facility/api/viewsets/patient.py
@@ -242,8 +242,8 @@ class PatientViewSet(
                 slice_obj = temp.form.cleaned_data.get(field)
                 if slice_obj:
                     days_difference = (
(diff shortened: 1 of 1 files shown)


vigneshhari:
Good Catch!

PR #655, opened 2021-12-21
Add ventilator fields to daily rounds

shaileshaanand:
As required in https://github.com/coronasafe/care_fe/issues/1973


care/facility/models/daily_round.py
```diff
@@ -315,6 +318,19 @@ class InsulinIntakeFrequencyType(enum.Enum):
...
         default=None, null=True, validators=[MinValueValidator(0), MaxValueValidator(5000)],
     )
+    ett_tt = models.IntegerField(
+        null=True,
+        default=None,
+        verbose_name="ETT/TT in mmid",
+        validators=[MinValueValidator(3), MaxValueValidator(10)],
+    )
+    intubation_start_date = models.DateTimeField(null=True, blank=True, default=None)
+    intubation_end_date = models.DateTimeField(null=True, blank=True, default=None)
```
gigincg, line 328: @shaileshaanand the intubation start and end date need not be there, because we can understand this from the Consultation Updates where the patient is intubated and then the end date would be the date when the patient is no longer intubated. 

Or in a nutshell, we won't need these 2 feels

Also, we need to add `cpk_mb` field as well. if I'm not wrong

shaileshaanand: Made the necessary changes.

PR #669, opened 2022-01-25
Modified JSON schema of daily rounds

sam9111:
Added "concentration" of type number and "conc_unit" of type string to INFUSION schema
Added "calories" of type number to FEED schema


gigincg:
@sam9111 add a description for the PR😅

gigincg (changes requested):
If you make concentration and conc unit required, you won't be able to make it optional later

gigincg:
Calories also can't be a required field. There's a possibility of the nurses not knowing the right value. This would mean they'll enter wrong or placebo values 

@sam9111

sam9111:
will create a new PR for this one

PR #670, opened 2022-01-25
Create Asset Integrations Classes

gigincg:


care/utils/assetintegration/onvif.py
```diff
@@ -0,0 +1,53 @@
...
+            self.meta = json.loads(meta)
+            self.meta = meta
+            self.name = meta["camera_type"]
+            self.host = meta["camera_address"]
+            self.port = meta["camera_port"] or 80
+            self.username = meta["camera_access_key"].split(":")[0]
+            self.password = meta["access_credentials"].split(":")[1]
+            self.middleware_hostname = meta["middleware_hostname"]
+        except KeyError:
+            print("Error: Invalid Onvif Asset")
```
vigneshhari, line 26: A bit more information on why it's an invalid asset would be great, this should be raising an exception.

care/utils/assetintegration/onvif.py
```diff
@@ -0,0 +1,53 @@
...
+        return "http://{}{}".format(self.middleware_hostname, endpoint)
+
+    def api_post(self, url, data=None):
+        req = request.post(url, json=data)
+
+    def api_get(self, url, data=None):
+        req = request.get(url, data=data)
+
+    def handle_action(self, action):
+        self.validate_camera(self)
```
vigneshhari, line 38: validate_camera does not seem to be a method in this class, no need to pass self explicitly since the first parameter is always self.

vigneshhari (changes requested):
Changes Requested

care/facility/api/viewsets/asset.py
```diff
@@ -117,6 +157,22 @@ def set_default_user_location(self, request, *args, **kwargs):
         except:
             raise Http404
 
+    # Asset Integration API
+    @action(detail=False, methods=["POST"])
```
vigneshhari, line 161: use the swagger auto schema and a dummy serializer to create the swagger spec for this API

vigneshhari, line 161: detail should be true right? the asset can be identified, authorized, and fetched by Django that way.

care/facility/api/viewsets/asset.py
```diff
@@ -117,6 +157,22 @@ def set_default_user_location(self, request, *args, **kwargs):
...
+        """
+        This API is used to operate assets. API accepts the asset_id and action as parameters.
+        """
+        if "asset_id" not in request.data:
+            raise ValidationError({"asset_id": "is required"})
+        if "action" not in request.data:
+            raise ValidationError({"action": "is required"})
+        asset_id = request.data["asset_id"]
+        action = request.data["action"]
+        asset: Asset = get_object_or_404(Asset.objects.filter(external_id=asset_id))
```
vigneshhari, line 172: No Authorisation filter.

care/facility/api/viewsets/asset.py
```diff
@@ -117,6 +157,22 @@ def set_default_user_location(self, request, *args, **kwargs):
...
+        AssetClasses(asset.asset_class).handle_action(action)
```
vigneshhari, line 173: Handle exceptions raised by this class. Document exceptions raised by the class in the class def as well.

care/facility/api/viewsets/asset.py
```diff
@@ -117,6 +157,22 @@ def set_default_user_location(self, request, *args, **kwargs):
...
+        return Response({"message": "Success"})
```
vigneshhari, line 174: Return 202 with no content.

vigneshhari (changes requested):
Changed requested in API

care/utils/assetintegration/onvif.py
gigincg: Validation has been moved to __init__, so removing the call

care/utils/assetintegration/onvif.py
gigincg: Switched to "Error: Invalid Onvif Asset; Missing required fields"

care/utils/assetintegration/onvif.py
```diff
@@ -0,0 +1,53 @@
...
+                "x": action.data["x"],
+                "y": action.data["y"],
+                "z": action.data["z"],
+                "speed": action.data["speed"],
+                "meta": self.meta,
+            }
+            self.api_post(self.get_url("absoluteMove"), data=request_data)
+
+        elif action.type == self.OnvifActions.GOTO_PRESET.value:
+            action.data["preset"]
```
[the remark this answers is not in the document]
gigincg: Added the API gotoPreset and getStatus as well

care/facility/api/viewsets/asset.py
gigincg: Done. Do check the Dummy once

gigincg: Fiex

care/facility/api/viewsets/asset.py
gigincg: Returning 200 with the JSON response

PR #684, opened 2022-02-22
Modified JSON schema of daily rounds

sam9111:
Added "concentration" of type number and "conc_unit" of type string to INFUSION schema
Added "calories" of type number to FEED schema


gigincg:
@sam9111 This can be merge, but the migration needs to be redone. Can you delete the migration, merge develop and then create a new migration?

Everything else looks good

PR #714, opened 2022-04-09
bed revision v1

arpancodes:
- [x] Remove: 
   - `AdmittedTo` in Consultation
   - `AdmittedTo` in DailyRound
   - `Bed` in DailyRound

- [x] Add Beds in Consultation
   - Add the Bed Relation (ConsultationBed Through Table)

- [ ] Switch Beds UI/API
   - API to Switch Beds of a Consultation (can be done with update view?)

### DIFF
--- a/care/facility/api/serializers/bed.py
+++ b/care/facility/api/serializers/bed.py
@@ -94,7 +94,7 @@ class ConsultationBedSerializer(ModelSerializer):
 
     def validate(self, attrs):
         user = self.context["request"].user
-        if "consultation" in attrs and "bed" in attrs:
+        if "consultation" in attrs and "bed" in attrs and "start_date" in attrs:
             consultation = get_object_or_404(PatientConsultation.objects.filter(external_id=attrs["consultation"]))
             bed = get_object_or_404(Bed.objects.filter(external_id=attrs["bed"]))
             facilities = get_facility_queryset(user)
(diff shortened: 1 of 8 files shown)


care/facility/api/serializers/bed.py
```diff
@@ -74,3 +76,36 @@ def validate(self, attrs):
...
+    bed = UUIDField(write_only=True, required=True)
+
+    class Meta:
+        model = ConsultationBed
+        exclude = ("deleted", "external_id")
+        read_only_fields = TIMESTAMP_FIELDS
+
+    def validate(self, attrs):
+        user = self.context["request"].user
+        if "consultation" in attrs and "bed" in attrs:
```
vigneshhari, line 97: These are required fields right?

care/facility/models/bed.py
```diff
@@ -39,3 +40,9 @@ class AssetBed(BaseModel):
     asset = models.ForeignKey(Asset, on_delete=models.PROTECT, null=False, blank=False)
     bed = models.ForeignKey(Bed, on_delete=models.PROTECT, null=False, blank=False)
     meta = JSONField(default=dict)
+
+
+class ConsultationBed(BaseModel):
+    consultation = models.ForeignKey(PatientConsultation, on_delete=models.PROTECT, null=False, blank=False)
+    bed = models.ForeignKey(Bed, on_delete=models.PROTECT, null=False, blank=False)
```
vigneshhari, line 47: We need a start date and end date for each consultation bed assignment, it should also be helpful if we can have a current bed association from the consultation for ease of queries. 
This start and end date would be used to identify when the patient has entered/ exited specific bed types

vigneshhari (changes requested):

vigneshhari:
Switching beds should be done in the POST/Create call for the consultation bed association end point, the backend should also perform the required logic of ending any existing consultation bed associations and moving any current pointers to the consultation bed model.

PR #718, opened 2022-04-18
updated `/add_user` endpoint to return status code instead of user object

### LINKED ISSUE #707: Update `/add_user` endpoint to return status code instead of user object
**Describe the bug**
Update `/add_user` endpoint to return status code instead of user object

**To Reproduce**
1. call `/add_user` api
2. inspect the success response from the api

**Expected behavior**
The API should respond with a success status if the user is created, else return error messages.

sainAk:
Fixes #707

### DIFF
--- a/care/users/api/viewsets/users.py
+++ b/care/users/api/viewsets/users.py
@@ -163,10 +163,7 @@ class UserViewSet(
         if User.objects.filter(username=username).exists():
             raise ValidationError({"username": "User with Given Username Already Exists"})
         user = serializer.create(serializer.validated_data)
-
-        response_data = UserCreateSerializer(user).data
-        # response_data["password"] = password
-        return Response(data=response_data, status=status.HTTP_201_CREATED)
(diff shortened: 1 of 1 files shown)


care/users/api/viewsets/users.py
```diff
@@ -158,15 +158,9 @@ def add_user(self, request, *args, **kwargs):
         serializer = UserCreateSerializer(
             data={**request.data, "password": password}, context={"created_by": request.user},
         )
-        serializer.is_valid(raise_exception=True)
-        username = request.data["username"]
-        if User.objects.filter(username=username).exists():
```
vigneshhari, line 163: you sure this validation is being performed regardless?

sainAk: yes it does as UserCreateSerializer is a modelserializer for User model where username field is unique

iamsdas (approved):

Marmik2003 (approved):

vigneshhari (approved):

PR #731, opened 2022-04-26
Validations based of the latest entry

### LINKED ISSUE #730: Validation issues in `ConsultationBed`
- [ ] Switching to the same bed consecutively shouldn't be permitted
- [ ] If `start_date` is the same as before for the new entry, `end_date` is not set. It's better to reject the request if the `start_date` is not greater than the previous `start_date`.

image

arpancodes:
Fixes #730 
- Add validation to prevent duplicate consecutive bed entries.
- Add validation to ensure the new `start_date` is not past the last `start_date`.


care/facility/api/serializers/bed.py
```diff
@@ -108,6 +108,14 @@ def validate(self, attrs):
             start_date = attrs["start_date"]
             end_date = attrs.get("end_date", None)
             existing_qs = ConsultationBed.objects.filter(consultation=consultation, bed=bed)
+            latest_qs = ConsultationBed.objects.filter(consultation=consultation).latest("id")
```
vigneshhari, line 111: This logic will crash if there is no latest bed. you need to add a block that only runs if the latest bed exists.

vigneshhari (changes requested):

vigneshhari (approved):
Need more checks, but this is a good start.

PR #736, opened 2022-05-11
Prevent creation of bed with the same name

iamsdas:
Prevent the creation of a bed, if a bed with the same name exists in that location.
Fixes: [#2282](https://github.com/coronasafe/care_fe/issues/2282)

### DIFF
--- a/care/facility/models/bed.py
+++ b/care/facility/models/bed.py
@@ -7,6 +7,7 @@ Leaving scope to build rooms and wards to being even more organization.
 import enum
 
 from django.contrib.postgres.fields.jsonb import JSONField
+from django.core.exceptions import ValidationError
 from django.db import models
 
 from care.facility.models.asset import Asset, AssetLocation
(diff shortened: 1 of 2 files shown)


vigneshhari (changes requested):
This validation should be written in the validate method, these validations can also be written at the model level which might make more sense.

iamsdas:
@vigneshhari
I updated the PR to add the validations in the model level (using the `clean` method) and made some minor changes so that DRF can handle these Django validation errors.

care/facility/models/bed.py
```diff
@@ -35,6 +36,14 @@ class BedType(enum.Enum):
     assets = models.ManyToManyField(Asset, through="AssetBed")
     location = models.ForeignKey(AssetLocation, on_delete=models.PROTECT, null=False, blank=False)
 
+    def clean(self) -> None:
+        if Bed.objects.filter(location=self.location, name=self.name).exists():
+            raise ValidationError({"name": "Bed with same name already exists in location."})
+
+    def save(self, *args, **kwargs) -> None:
+        self.clean()
```
vigneshhari, line 44: validate makes more sense here since you are not really cleaning anything.

vigneshhari (approved):

vigneshhari:
resolve the conflicts and we are good to merge.

vigneshhari (approved):

PR #752, opened 2022-05-18
added localstack s3 buckets for local environment

Marmik2003:
1. added localstack in docker-compose.local.yaml
2. added bucket-setup.sh to create buckets upon running container

### DIFF
--- a/docker-compose.local.yaml
+++ b/docker-compose.local.yaml
@@ -27,5 +27,39 @@ services:
     volumes:
       - .:/app
 
+  redis:
+    container_name: care_redis
+    image: redis:alpine
+    restart: always
+    ports:
+      - "6379:6379"
+
+  celery:
+    container_name: care_celery
+    build: 
+      context: .
+      dockerfile: docker/DevDockerfile
+    env_file:
+      - ./docker/.local.env
+    entrypoint: ["bash", "devCelery.sh"]
+
+    depends_on:
+      - db
+      - backend
+      - redis
+
+  localstack:
+    image: localstack/localstack:latest
+    environment: 
+      - AWS_DEFAULT_REGION=ap-south-1
+      - EDGE_PORT=4566
+      - SERVICES=s3
+    ports: 
+      - '4566-4583:4566-4583'
+    volumes: 
+      - "${TEMPDIR:-/tmp/localstack}:/tmp/localstack"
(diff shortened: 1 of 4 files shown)


Marmik2003:
fixes #751

sainAk (approved):

docker-compose.local.yaml
```diff
@@ -27,5 +27,39 @@ services:
...
+    image: localstack/localstack:latest
+    environment: 
+      - AWS_DEFAULT_REGION=ap-south-1
+      - EDGE_PORT=4566
+      - SERVICES=s3
+    ports: 
+      - '4566-4583:4566-4583'
+    volumes: 
+      - "${TEMPDIR:-/tmp/localstack}:/tmp/localstack"
+      - "/var/run/docker.sock:/var/run/docker.sock"
```
vigneshhari, line 61: We don't need to mount docker.

vigneshhari (changes requested):

docker-compose.local.yaml
```diff
@@ -27,5 +27,39 @@ services:
...
+      - "6379:6379"
+
+  celery:
+    container_name: care_celery
+    build: 
+      context: .
+      dockerfile: docker/DevDockerfile
+    env_file:
+      - ./docker/.local.env
+    entrypoint: ["bash", "devCelery.sh"]
```
vigneshhari, line 44: Since this is the local version, we should mount local volumes as well.

vigneshhari (changes requested):

docker-compose.local.yaml
Marmik2003: done

docker-compose.local.yaml
Marmik2003: done

docker-compose.local.yaml
```diff
@@ -27,5 +27,39 @@ services:
...
+    image: redis:alpine
+    restart: always
+    ports:
+      - "6379:6379"
+
+  celery:
+    container_name: care_celery
+    build:
+      context: .
+      dockerfile: docker/DevDockerfile
```
vigneshhari, line 41: This does not contain a beat instance, create another container for beat or add the beat functionality to this container.

vigneshhari (changes requested):

vigneshhari (approved):
LGTM. Have not verified.

PR #772, opened 2022-05-20
fix: Inventory unexpectedly marked as Low Stock

### LINKED ISSUE #725: Inventory unexpectedly marked as Low Stock
Creating a PPE Inventory, without having a Minimum Quantity marks it as Low Stock. This shouldn't be the case

sainak:
If the FacilityInventoryMinQuantity object is not present for an item it uses the value defined in the [FacilityInventoryItem](https://github.com/coronasafe/care/blob/master/care/facility/models/inventory.py#L54)
so the fix would be to ignore the min value in FacilityInventoryItem and retire the min_quantity field

fixes #725, closes coronasafe/care_fe#2437


vigneshhari:
@sainAk the min_quantity can stay and maybe FE can ignore the flag if the min_quantity value is not present, or we can add that logic into our flagging system.

sainAk:
> @sainAk the min_quantity can stay and maybe FE can ignore the flag if the min_quantity value is not present, or we can add that logic into our flagging system.

Sure I'll apply the changes

vigneshhari:
@sainAk in FE or BE ?

sainAk:
Its possible to do it on the BE but the min_quantity will still be redundant atlest in `FacilityInventoryLogSerializer`

sainAk:
what we can do is rename the min_quantity field to default_min_quantity in the response and decide the is_low based on the `FacilityInventoryMinQuantity`

vigneshhari:
yea sure

PR #771, opened 2022-05-20
add migration to change meta.camera_address to meta.local_ip_address

iamsdas:
Add data migration to change `meta.camera_address` to `meta.local_ip_address` in Asset table.

fixes: #759

PR #776, opened 2022-05-21
Add new class for HL7MonitorAsset

iamsdas:
Add new class for HL7MonitorAsset

fixes: #758
dependent on: https://github.com/coronasafe/care/pull/771

### DIFF
--- a/care/utils/assetintegration/onvif.py
+++ b/care/utils/assetintegration/onvif.py
@@ -17,7 +17,7 @@ class OnvifAsset(BaseAssetIntegration):
             self.meta = json.loads(meta)
             self.meta = meta
             self.name = meta["camera_type"]
-            self.host = meta["camera_address"]
+            self.host = meta["local_ip_address"]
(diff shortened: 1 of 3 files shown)


iamsdas:
@gigincg could you confirm if everything required for the class is present?

care/utils/assetintegration/onvif.py
```diff
@@ -17,7 +17,7 @@ def __init__(self, meta):
             self.meta = json.loads(meta)
             self.meta = meta
             self.name = meta["camera_type"]
-            self.host = meta["camera_address"]
+            self.host = meta["local_ip_address"]
```
vigneshhari, line 20: Is this change in onvif intended?

iamsdas: yup, `camera_address` was not an accurate name and has been replaced in [#771](https://github.com/coronasafe/care/pull/771).

vigneshhari (approved):

PR #777, opened 2022-05-21
fix issues when qr_code_id is a blank string and show message for unique integrity error

sainAk:
fixes coronasafe/care_fe#2422


Marmik2003 (approved):
Looks good to me

vigneshhari (approved):
Approved without testing

PR #787, opened 2022-05-26
added DELETE endpoint for file_upload

Marmik2003:
I think this code is intact for deleting, but not working


Marmik2003:
API endpoint for https://github.com/coronasafe/care_fe/issues/2288

Issue #792, opened 2022-05-27
Getting a response of 500 when attempting to create asset with asset_class specified.


rithviknishad:
1. Create new asset with `asset_class` attribute included with a valid value (eg: "ONVIF" / "HL7MONITOR")

**Expected behavior**
201 Response

**Screenshots**

![image](https://user-images.githubusercontent.com/25143503/170700869-5dd427dd-d532-4904-9171-eacffc3f3bf9.png)

However, endpoint works fine when `asset_class` is not included.

![image](https://user-images.githubusercontent.com/25143503/170701025-da06d0a1-1d8d-453a-bff4-30aaa5a20e62.png)

rithviknishad:
@vigneshhari @iamsdas still getting 500 as response :/
Am I doing something wrong or...?

![image](https://user-images.githubusercontent.com/25143503/171834317-b5c65b96-dec5-469f-997f-4c221bb5f892.png)

iamsdas:
@rithviknishad have you pulled the latest changes from the master branch? It seems to be working fine for me locally (using the Swagger UI). Also, check if you are providing the necessary input fields.

Here's the request body I used for testing:
```
{
  "asset_type": "INTERNAL",
  "location_object": {
    "facility": {
      "name": "facility #1"
    },
    "name": "al 1",
    "description": "string",
    "location_type": 10
  },
  "location": "18b21291-1c87-4bfc-8c02-f60c5ccc285f",
  "name": "asset name",
  "description": "string",
  "asset_class": "ONVIF",
  "is_working": true,
  "not_working_reason": "string",
  "serial_number": "string",
  "warranty_details": "string",
  "meta": {},
  "vendor_name": "string",
  "support_name": "string",
  "support_phone": "+919999999999",
  "support_email": "user@example.com",
  "qr_code_id": null
}
```
![Screenshot_20220603_155057](https://user-images.githubusercontent.com/26461855/171836380-7d3229b9-5fd5-46c4-8e83-6546336933e5.png)

rithviknishad:
@iamsdas I'm using the staging API via proxy mentioned in `care_fe/package.json`

I'm yet to setup a local care BE on my machine.

![image](https://user-images.githubusercontent.com/25143503/171849660-749941b9-314e-494f-b7df-696037c63ac8.png)

- What I'm performing is:
```sh
curl 'https://careapi.coronasafe.in/api/v1/asset/' \
        -X POST \
[...]

PR #793, opened 2022-05-27
change model field type of asset_class

iamsdas:
Update field type of `asset_class` from integer field to char field.
fixes: #792


Marmik2003 (approved):

gigincg:
@vigneshhari Should we create an Integer Mapping for the Asset Classes and keep it as an Integer Field

cc: @iamsdas

vigneshhari (approved):

PR #797, opened 2022-05-28
Added API Route for User Password Change

siddnikh:
The API route is /api/v1/password_change/
Need to update documentation and frontend.


config/settings/local.py
```diff
@@ -104,3 +104,7 @@
 AUDIT_LOG_ENABLED = True
 
 DISABLE_RATELIMIT = True
+
+# Uncomment when setting up on mac
+# GDAL_LIBRARY_PATH = '/opt/homebrew/opt/gdal/lib/libgdal.dylib'
+# GEOS_LIBRARY_PATH = '/opt/homebrew/opt/geos/lib/libgeos_c.dylib'
```
Marmik2003, line 110: i think adding a flag like GDAL_MANUAL_PATH would be better rather than commenting and uncommenting...

```suggestion
GDAL_MANUAL_PATH = env.bool("GDAL_MANUAL_PATH", default=False)
if GDAL_MANUAL_PATH:
    GDAL_LIBRARY_PATH = env("GDAL_LIBRARY_PATH", default=None)
    GEOS_LIBRARY_PATH = env("GEOS_LIBRARY_PATH", default=None)
```

Marmik2003 (changes requested):

config/settings/local.py
siddnikh: Yeah, this is a better implementation. Let's go with this.

Marmik2003 (approved):
LGTM

care/users/api/viewsets/change_password.py
```diff
@@ -0,0 +1,52 @@
...
+
+User = get_user_model()
+
+class ChangePasswordSerializer(serializers.Serializer):
+    model = User
+
+    """
+    Serializer for password change endpoint.
+    """
+    username = serializers.CharField(required=True)
```
vigneshhari, line 16: Why is the username required ?

care/facility/migrations/0293_auto_20220526_2249.py
```diff
@@ -0,0 +1,24 @@
+# Generated by Django 2.2.11 on 2022-05-26 17:19
+
+import django.core.validators
```
vigneshhari, line 3: Why is this migration part of this PR ?

config/settings/local.py
```diff
@@ -104,3 +104,7 @@
 AUDIT_LOG_ENABLED = True
 
 DISABLE_RATELIMIT = True
+GDAL_MANUAL_PATH = env.bool("GDAL_MANUAL_PATH", default=False)
```
vigneshhari, line 107: Changes Irrelevant to PR, Remove.

vigneshhari (changes requested):

care/users/api/viewsets/change_password.py
```diff
@@ -0,0 +1,51 @@
...
+
+User = get_user_model()
+
+class ChangePasswordSerializer(serializers.Serializer):
+    model = User
+
+    """
+    Serializer for password change endpoint.
+    """
+    old_password = serializers.CharField(required=True)
```
vigneshhari, line 16: Why is this field not `write_only=True`? ideally, both these fields do not need `write_only` tags since they are not used to render any content.

care/users/api/viewsets/change_password.py
```diff
@@ -0,0 +1,51 @@
...
+        serializer_class = ChangePasswordSerializer
+        model = User
+        permission_classes = (IsAuthenticated,)
+
+        def get_object(self):
+            obj = self.request.user
+            return obj
+
+        def update(self, request, *args, **kwargs):
+            self.object = self.get_object()
```
vigneshhari, line 32: do we need a function call here? can't we just do `self.request.user` ?

vigneshhari (changes requested):

care/users/api/viewsets/change_password.py
```diff
@@ -0,0 +1,51 @@
...
+        """
+        An endpoint for changing password.
+        """
+        serializer_class = ChangePasswordSerializer
+        model = User
+        permission_classes = (IsAuthenticated,)
+
+        def get_object(self):
+            obj = self.request.user
+            return obj
```
sainAk, line 29:
```suggestion
            return self.request.user
```

care/users/api/viewsets/change_password.py
```diff
@@ -0,0 +1,51 @@
...
+            self.object = self.request.user
+            serializer = self.get_serializer(data=request.data)
+
+            if serializer.is_valid():
+                # Check old password
+                if not self.object.check_password(serializer.data.get("old_password")):
+                    return Response({"old_password": ["Wrong password entered. Please check your password."]}, status=status.HTTP_400_BAD_REQUEST)
+                # set_password also hashes the password that the user will get
+                self.object.set_password(serializer.data.get("new_password"))
+                self.object.save()
```
sainAk, line 41: should be handled in serializer

care/users/api/viewsets/change_password.py
```diff
@@ -0,0 +1,51 @@
...
+                response = {
+                    'status': 'success',
+                    'code': status.HTTP_200_OK,
+                    'message': 'Password updated successfully',
+                    'data': []
+                }
+
+                return Response(response)
```
sainAk, line 49: We prefer http status codes in care
```suggestion
                return Response({"message": "Password updated successfully"})
```

sainAk (changes requested):

care/users/api/viewsets/change_password.py
```diff
@@ -0,0 +1,47 @@
...
+    model = User
+
+    """
+    Serializer for password change endpoint.
+    """
+    old_password = serializers.CharField(required=True)
+    new_password = serializers.CharField(required=True)
+
+    def check_old_password(self, instance, validated_data):
+        if instance.check_password(validated_data.get("old_password")):
```
vigneshhari, line 20: can we just return this value, do we need an if-else here? also, why is this in the serializer?

care/users/api/viewsets/change_password.py
```diff
@@ -0,0 +1,47 @@
...
+
+class ChangePasswordView(UpdateAPIView):
+        """
+        An endpoint for changing password.
+        """
+        serializer_class = ChangePasswordSerializer
+        model = User
+        permission_classes = (IsAuthenticated,)
+
+        def get_object(self):
```
vigneshhari, line 32: why is this method present?

vigneshhari (changes requested):

PR #807, opened 2022-05-31
fix: regression from #715

sainAk:
the `admitted_to` field id no longer present on `PatientConsultation`, the fix was to get the value from `PatientConsultation__current_bed__bed__bed_type`


Marmik2003 (approved):

sainAk:
superseded by #924

PR #810, opened 2022-06-01
Fix error format for external test

Ashesh3:
Required for https://github.com/coronasafe/care_fe/pull/2614

Just a minor change to how list of errors are sent to the client for the external test page.


care/facility/api/viewsets/patient_external_test.py
```diff
@@ -133,7 +133,7 @@ def bulk_upsert(self, request, *args, **kwargs):
             valid = serialiser_obj.is_valid()
             current_error = prettyerrors(serialiser_obj._errors)
             if current_error and (not valid):
-                errors[counter] = current_error
+                errors.append(current_error)
```
vigneshhari, line 136: Do you lose the context on what row has the error?

vigneshhari (changes requested):

Ashesh3:
@vigneshhari Any updates on this? :D

vigneshhari (approved):

Issue #658, opened 2021-12-28
Integrate with Middleware for Periodic Sync of Asset Information


gigincg:
Create CRON Jobs that would periodically update Patient Information after fetching data from the middleware

_This is a draft issue. Device Integrations that affect Patient Information needs to be triaged_

bodhish:
@gigincg Middleware should update care in specific intervals.

gigincg:
https://github.com/coronasafe/care/pull/811 Should Close this

PR #811, opened 2022-06-01
Finalize MiddlewareAuthentication for required endpoints

### LINKED ISSUE #671: Daily Rounds API for Middleware
Create a Daily Round API that would be hit from the Middleware running at the spokes

- [x] #667

sainAk:
- [x] Create reusable permissions component for assets
- [x] add endpoint to get patient connected to asset `/api/v1/consultation/patient_from_asset/`
- [x] Write Tests

closes #671


care/facility/models/daily_round.py
```diff
@@ -462,3 +515,14 @@ def has_object_write_permission(self, request):
...
         return self.has_object_read_permission(request)
+
+    def has_object_asset_read_permission(self, request):
+        return False
+
+    def has_object_asset_write_permission(self, request):
+        consultation = PatientConsultation.objects.get(
+            external_id=request.parser_context["kwargs"]["consultation_external_id"]
+        )
+        return consultation.current_bed.bed.bed_set.filter(
```
vigneshhari, line 526: Lets use the AssetBed Class to query, using set can be confusing for the most part.

vigneshhari (approved):

vigneshhari:
We also need an endpoint that returns the current patient/consultation, only accessible by an asset, the data should be of the patient that is connected to the asset via a bed. Only the external_id needs to be returned

care/facility/api/viewsets/patient_consultation.py
```diff
@@ -17,28 +25,59 @@ class PatientConsultationFilter(filters.FilterSet):
...
         applied_filters |= Q(assigned_to=self.request.user)
         applied_filters |= Q(patient__assigned_to=self.request.user)
         return self.queryset.filter(applied_filters)
 
+    @swagger_auto_schema(
+        responses={200: PatientConsultationIDSerializer},
+    )
+    @action(detail=False, methods=["GET"])
+    def patient_from_asset(self, request):
+        try:
```
vigneshhari, line 73: get the bed separately, validate, and then use that instead of using set.

care/facility/api/viewsets/patient_consultation.py
```diff
@@ -17,28 +25,59 @@ class PatientConsultationFilter(filters.FilterSet):
...
+            consultation = (
+                PatientConsultation.objects.select_related("patient")
+                .order_by("-id")
+                .filter(current_bed__bed__in=request.user.asset.bed_set.all())
+                .only("external_id", "patient__external_id")
+                .first()
```
vigneshhari, line 79: Will there be more than one ?

care/facility/api/viewsets/patient_consultation.py
```diff
@@ -17,28 +25,59 @@ class PatientConsultationFilter(filters.FilterSet):
...
+            )
+        except PatientConsultation.DoesNotExist as e:
+            raise NotFound({"detail": "No consultation found for this asset"}) from e
+        return Response(PatientConsultationIDSerializer(consultation).data)
```
vigneshhari, line 83: What happens if there are no records from the database, send a different code maybe ?

vigneshhari (changes requested):

care/facility/api/viewsets/patient_consultation.py
sainAk: I think 404 is appropriate response code, but if you want something else please mention it.

care/facility/api/viewsets/patient_consultation.py
sainAk: yes, there can be multiple consultation with same current beds as it's an FK.

care/facility/api/viewsets/patient_consultation.py
vigneshhari: NVM this is good

gigincg:
Solves https://github.com/coronasafe/care/issues/658

vigneshhari:
Move tests to a different PR.

vigneshhari (approved):
Approving without testing.

PR #815, opened 2022-06-03
Allow DELETE method for bed endpoint

Ashesh3:
- [ ] Frontend change: https://github.com/coronasafe/care_fe/pull/2645

Basically allows bed to be deleted. Adds the `DestroyModelMixin` to the `BedViewSet` and custom define the `destroy` method


Marmik2003 (approved):
looks ok

vigneshhari (approved):
LGTM, Untested.

PR #821, opened 2022-06-05
Add a "working" filter for Assets

siddnikh:
Made in order to solve: https://github.com/coronasafe/care_fe/issues/2055


care/facility/api/serializers/asset.py
```diff
@@ -30,7 +30,7 @@ class AssetSerializer(ModelSerializer):
     id = UUIDField(source="external_id", read_only=True)
     status = ChoiceField(choices=Asset.StatusChoices, read_only=True)
     asset_type = ChoiceField(choices=Asset.AssetTypeChoices)
-
+    is_working = ChoiceField(choices=Asset.AssetWorkingChoices)
```
vigneshhari, line 33: Does it make more sense for the BE to return True/False rather than text? It is usually a pattern that is followed to decouple texts from the BE to allow easier translations.

siddnikh: sure, I'll change it to text.

care/facility/models/asset.py
```diff
@@ -62,10 +64,8 @@ class Status(enum.Enum):
         choices=AssetClassChoices, default=None, null=True, blank=True, max_length=20
     )
     status = models.IntegerField(choices=StatusChoices, default=Status.ACTIVE.value)
-    current_location = models.ForeignKey(
-        AssetLocation, on_delete=models.PROTECT, null=False, blank=False
-    )
-    is_working = models.BooleanField(default=None, null=True, blank=True)
+    current_location = models.ForeignKey(AssetLocation, on_delete=models.PROTECT, null=False, blank=False)
+    is_working = models.CharField(choices=AssetWorkingChoices, default=None, null=True, blank=True, max_length=11)
```
vigneshhari, line 68: Change this back to the boolean field, if there are only two options then it does not make sense to not use boolean fields. also characters are not a logical value for a field named `is_working`

care/facility/api/serializers/asset.py
vigneshhari: move to boolean values.

care/facility/api/viewsets/asset.py
```diff
@@ -93,6 +93,7 @@ class AssetFilter(filters.FilterSet):
     location = filters.UUIDFilter(field_name="current_location__external_id")
     asset_type = CareChoiceFilter(choice_dict=inverse_asset_type)
     status = CareChoiceFilter(choice_dict=inverse_asset_status)
+    is_working = CareChoiceFilter(choice_dict=Asset.AssetWorkingChoices)
```
vigneshhari, line 96: move to boolean values.

care/facility/models/asset.py
```diff
@@ -51,6 +51,8 @@ class Status(enum.Enum):
         ACTIVE = 50
         TRANSFER_IN_PROGRESS = 100
 
+    AssetWorkingChoices = [("WORKING", "WORKING"), ("NOTWORKING", "NOTWORKING")]
```
vigneshhari, line 54: This is not required.

vigneshhari (changes requested):

care/facility/models/asset.py
siddnikh: done.

vigneshhari (approved):

PR #822, opened 2022-06-06
Remove Postgis

cp-Coder:
## Updates
- [x] Removed location field
- [x] Removed unwanted packages
- [x] Added latitude and longitude fields
- [x] Added migrations to copy data from location field


sainAk:
This change would be a 2 step process
1. Migrating the location field to normal fields 
2. removing references of the location filed from older migrations (0015, 0017)

vigneshhari:
we need to remove references from old migrations as well.

cp-Coder:
> we need to remove references from old migrations as well.

@sainAk Mentioned this issue and told me to do it in new PR. I have already mentioned it above. #827

vigneshhari (approved):

PR #826, opened 2022-06-06
add workflow to build and deploy docs

### LINKED ISSUE #825: Create pipelines to automatically build documentation
_Originally posted by @vigneshhari in https://github.com/coronasafe/care/issues/763#issuecomment-1147457139_

sainAk:
closes #825

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -16,6 +16,11 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
+      - uses: actions/setup-python@v3
+        with:
+          python-version: "3.10"
+          cache: "pip"
+
       - name: Install dependencies
         run: |
           pip install -r requirements/docs.txt
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,53 @@
+name: Build Docs
+
+on:
+  push:
+    paths:
+      - "docs/**"
+  pull_request:
+    paths:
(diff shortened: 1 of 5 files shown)


.github/workflows/docs.yml
```diff
@@ -0,0 +1,59 @@
+name: Build Docs
+
+on:
+  push:
```
vigneshhari, line 4: we only need to build/deploy on the master branch

vigneshhari (changes requested):

Marmik2003 (approved):
Looks good, i think it is ok if we don't remove sphinx from local requirements as some developers might also want to change the docs after few major changes!

sainAk:
> Looks good, i think it is ok if we don't remove sphinx from local requirements as some developers might also want to change the docs after few major changes!

I've refrenced the docs.txt requirements in local.txt so sphinx will also get installed in local environments

.github/workflows/docs.yml
```diff
@@ -0,0 +1,62 @@
...
+
+      - name: Upload sphinx documentation
+        uses: actions/upload-artifact@v3
+        with:
+          name: sphinx-docs
+          path: ./docs/_build
+          retention-days: 30
+
+  deploy-docs:
+    if: github.ref == 'refs/heads/master'
```
vigneshhari, line 43: Do we need this, this job only runs on pushes to master right?

vigneshhari (approved):

PR #830, opened 2022-06-07
DRY Permissions added for assetLocation viewset

### LINKED ISSUE #818: Bug: read-only user can edit/update location management of a facility
**Describe the bug**
Read-Only users should not be allowed to make changes to the facility, but it can edit and add locations and also the bed in a location

**To Reproduce**

1. Login as a StaffReadOnly user (username:readonly, password:Welcome@123)
2. Go to a particular facility page
3. Click on "Location Management"
4. You will be able to Add/Edit location and Bed for those locations

**Expected behavior**
Read-Only user should be prohibited from updating/editing location data

ishanExtreme:
### Issue Fixed
fixes #818 

### Updates
**Read only users can now only read the location list and will not be able to perform any create/edit on it**
- Added new DRY permission class *models/mixins/permissions/asset.py*
- Updated AssetLocation model to inherit from DRYPermission class
- Updated AssetLocationViewset to use DryPermission and isAuthenticated Permission

### DIFF
--- a/care/facility/api/viewsets/asset.py
+++ b/care/facility/api/viewsets/asset.py
@@ -4,10 +4,17 @@ from django.shortcuts import get_object_or_404
 from django_filters import rest_framework as filters
 from drf_yasg.utils import swagger_auto_schema
 from rest_framework import filters as drf_filters
+from dry_rest_permissions.generics import DRYPermissions
+from rest_framework.permissions import IsAuthenticated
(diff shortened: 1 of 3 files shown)


sainAk (approved):
Try to not change the formatting of code, It generates diff noise which makes reviews harder, otherwise LGTM

vigneshhari (approved):

PR #835, opened 2022-06-07
Added tests workflow

skks1212:
Added django test workflow

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,26 @@
+name: Django CI
+
+on:
+  pull_request:
+    branches:
+      - '**'
+  push:
+    branches:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -41,3 +41,8 @@ makemigrations:
 test:
 	make up
 	docker exec care bash -c "python manage.py test --keepdb"
+
+test_coverage:
+	make up
```
sainAk, line 46:
```suggestion
test_coverage: up
```

sainAk:
@skks1212 @Marmik2003 we can integrate the test step in the deployment workflow 
https://github.com/coronasafe/care/blob/master/.github/workflows/deployment.yaml#L54-L56

PR #836, opened 2022-06-07
update pre-commit hooks and dependencies

### LINKED ISSUE #417: pre-commit fails while trying to commit the changes that we made, [un-used imports on various files]
### pre-commit failures

-  imported but unused [ on various files ]
- invalid escape sequence
- undefined names, E722 do not use bare 'except' ..etc

> currently for commiting changes we need to use --no-verify flag to get override these  

Screenshot from 2020-06-27 23-52-32

Screenshot from 2020-06-27 23-54-22

sainAk:
use common code style everywhere

removed 
- pycodestyle
- pylint

closes #417


vigneshhari:
@sainAk the app requires write access to the code, so the app won't be installed. we can keep tabs on it.

PR #848, opened 2022-06-10
update sentry sdk

### LINKED ISSUE #844: Update sentry sdk and specify deployment environment for each deployment
- [ ] The sentry sdk is quiet old and we are missing out on may new features in sentry so we need to update it to the latest supported version for django 2.2
- [ ] Also there is no seperation between deployments so all issues end up in the same environment, we need to specify the environment names to enforce separation.
- [ ] Another thing worth mentioning is that we are being spammed by requests with unknown hosts which triggers `django.security.DisallowedHost` events on sentry these events are not worth tracking and they only consume events bandwidth, it's possible to ignore these events from the deployment itself.

sainAk:
added SENTRY_ENVIRONMENT to set environment
and SENTRY_TRACES_SAMPLE_RATE to set trace sample rate
ignore django.security.DisallowedHost events

closes #844


Marmik2003 (approved):

skks1212 (approved):

vigneshhari (approved):

vigneshhari:
Try to exclude unrelated changes in the PR, there are settings reconfigurations and some precommit changes in a PR about sentry!.

config/settings/deployment.py
```diff
@@ -0,0 +1,196 @@
...
+SENTRY_DSN = env("SENTRY_DSN")
+SENTRY_LOG_LEVEL = env.int("DJANGO_SENTRY_LOG_LEVEL", logging.INFO)
+
+sentry_logging = LoggingIntegration(
+    level=SENTRY_LOG_LEVEL,
+    event_level=logging.ERROR,  # Capture info and above as breadcrumbs  # Send errors as events
+)
+sentry_sdk.init(
+    dsn=SENTRY_DSN,
+    environment=env("SENTRY_ENVIRONMENT", default="deployment-unknown"),
```
vigneshhari, line 178: @sainAk make sure @mathew-alex knows about this change, update all concerned deployment documents as well.

vigneshhari (approved):

PR #851, opened 2022-06-11
fix test warnings and parallel runs

sainAk:
we are going to drop most of the templates so `django-coverage-plugin` wont be required


Marmik2003 (approved):

skks1212 (approved):
Great stuff!

PR #853, opened 2022-06-11
Adding field validator for non-empty field

### LINKED ISSUE #849: create a field validator for validating non empty fields
create a field validator named `not_empty` so it can be added directly to the fields in serializer that are required to have some text

_context: https://github.com/coronasafe/care/issues/838#issuecomment-1152619471_

cp-Coder:
# Updates

- [x] Fixes #849
- [x] Added `notEmptyValidator` to validate the object
- [x] Modified serializers to validate the objects


care/utils/validation/not_empty.py
```diff
@@ -0,0 +1,21 @@
+from django.utils.translation import gettext as _
+from rest_framework.utils.representation import smart_repr
+from rest_framework import serializers
+
+class notEmptyValidator:
```
vigneshhari, line 5: follow class names patterns by [PEP 0008](https://peps.python.org/pep-0008/#class-names)

care/utils/validation/not_empty.py
```diff
@@ -0,0 +1,21 @@
...
+    Validator for checking whether the field has empty value or not
+    """
+    message = _('{field} cannot be empty')
+
+    def __init__(self, field, message=None):
+        self.field = field
+        self.message = message or self.message
+
+    def __call__(self, attrs):
+        if not attrs[self.field].strip():
```
vigneshhari, line 16: can there be a condition where attrs does not have the given field?

care/utils/validation/not_empty.py
```diff
@@ -0,0 +1,21 @@
+from django.utils.translation import gettext as _
+from rest_framework.utils.representation import smart_repr
+from rest_framework import serializers
```
vigneshhari, line 3: import ValidationError instead of importing the whole package

vigneshhari (changes requested):

cp-Coder:
I don't think it is relevant now.

Issue #850, opened 2022-06-11
A patient can be switched to same bed at which he/she before was


patelaryan7751:
# Bugs

## A patient can be switched to the same bed at which he/she before was

https://user-images.githubusercontent.com/59426397/173169660-b5446de4-637c-41af-a51d-c067cd92da2e.mp4

aparnacoronasafe:
Multiple people can be assigned to the same bed.

Eg: https://care.coronasafe.in/facility/015ad911-39b7-4104-b304-bf8178c33830/patients
2 patients are assigned to Bed10- 10bedicu.

Add rule that if a bed is occupied by any patient, that bed must not appear as an option in the dropdown in "Assign bed" window.

Issue #847, opened 2022-06-10
After a patient is discharged from CARE we can still switch the bed


patelaryan7751:
# Bugs 

## After a patient is discharged from CARE we can still switch the bed

https://user-images.githubusercontent.com/59426397/173101600-299382df-67c6-44a3-a120-6e8234e12555.mp4

PR #854, opened 2022-06-11
update patient bed switching logic

iamsdas:
Prevent patients from switching to the same bed during consultation while the previous one hasn't ended.
fixes: #850 
fixes: #847 (both are similar/related issues)

### DIFF
--- a/care/facility/api/serializers/bed.py
+++ b/care/facility/api/serializers/bed.py
@@ -125,14 +125,15 @@ class ConsultationBedSerializer(ModelSerializer):
     def create(self, validated_data):
         consultation = validated_data["consultation"]
         bed = validated_data["bed"]
-        current_bed = consultation.current_bed.bed if consultation.current_bed else None
+
         existing_beds = ConsultationBed.objects.filter(
-            consultation=consultation,
-            bed=current_bed,
-            end_date__isnull=True,
-        )
+            consultation=consultation, end_date__isnull=True)
+        if existing_beds.filter(bed=bed).exists():
+            raise ValidationError({"Bed:": ["Bed already in use by patient"]})
         existing_beds.update(end_date=validated_data["start_date"])
+
         obj = super().create(validated_data)
-        consultation.current_bed = obj  # This needs better logic, when an update occurs and the latest bed is no longer the last bed consultation relation added.
+        # This needs better logic, when an update occurs and the latest bed is no longer the last bed consultation relation added.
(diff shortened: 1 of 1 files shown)


iamsdas:
before merging, do confirm if this was actually a bug and not some design consideration.

iamsdas:
Added commit for preventing bed switching after the patient has been discharged from care.

care/facility/api/serializers/bed.py
```diff
@@ -125,14 +125,20 @@ def validate(self, attrs):
     def create(self, validated_data):
         consultation = validated_data["consultation"]
         bed = validated_data["bed"]
-        current_bed = consultation.current_bed.bed if consultation.current_bed else None
+
+        if consultation.patient.is_active is False:
```
vigneshhari, line 129:
```suggestion
        if not consultation.patient.is_active:
```

care/facility/api/serializers/bed.py
```diff
@@ -125,14 +125,20 @@ def validate(self, attrs):
...
+            raise ValidationError(
+                {"Patient:": ["Patient is already discharged from CARE"]})
```
vigneshhari, line 131:
```suggestion
                {"patient:": ["Patient is already discharged from CARE"]})
```

care/facility/api/serializers/bed.py
```diff
@@ -125,14 +125,20 @@ def validate(self, attrs):
...
+
         existing_beds = ConsultationBed.objects.filter(
-            consultation=consultation,
-            bed=current_bed,
-            end_date__isnull=True,
-        )
+            consultation=consultation, end_date__isnull=True)
+
+        if existing_beds.filter(bed=bed).exists():
+            raise ValidationError({"Bed:": ["Bed already in use by patient"]})
```
vigneshhari, line 137:
```suggestion
            raise ValidationError({"bed:": ["Bed already in use by patient"]})
```

vigneshhari (changes requested):

iamsdas:
Added a commit to prevent switching to occupied beds (regardless of consultation).
Added it to address this [issue/comment](https://github.com/coronasafe/care/issues/850#issuecomment-1159448522).
This does not affect already existing multi-occupancies of beds. Should I add some kind of migration to handle them? (not sure how that would work though)

vigneshhari (approved):

Issue #657, opened 2021-12-28
Asset Integration Classes for Multi Para Monitors


gigincg:
Create Asset Integration Classes for AssetType Multi Para Monitors

PR #858, opened 2022-06-16
add asset integration class for multipara monitors

iamsdas:
fixes: #657


iamsdas:
@gigincg could you check if any fields need to be modified/added?

iamsdas:
@gigincg could you check if any fields need to be modified/added?

gigincg:
Duplicate of HL7 Monitor

PR #864, opened 2022-06-20
Added an endpoint to search diseases from ICD11

khavinshankar:
Added an endpoint (`/api/v1/icd`) to search for diseases from International Classification of Diseases (ICD), Refer: [care_fe#2016](https://github.com/coronasafe/care_fe/issues/2016)

image
image

References:
https://icd.who.int/ct11/icd11_mms/en/release
https://icd.who.int/icdapi

### DIFF
--- a/care/facility/api/serializers/icd.py
+++ b/care/facility/api/serializers/icd.py
@@ -1,0 +1,17 @@
+from rest_framework import serializers
+
+from care.facility.models.icd import ICDDisease
+
+
+class ICDSerializer(serializers.Serializer):
+    id = serializers.CharField(max_length=50, read_only=True)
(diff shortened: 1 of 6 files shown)


vigneshhari:
I need more context, is the API inaccessible from the front-end, i generally do not prefer using the BE as a proxy if it can be performed directly from the FE.

khavinshankar:
Yes, we can access this API from front-end but every time we want to access any endpoint, we have to get the bearer token first (authentication is needed). If it is done in the front-end, we have to do the authentication for every client, so thought of going with backend.

@vigneshhari So what are your thoughts Vignesh, should I move this to frontend?

khavinshankar:
continued in #947

PR #877, opened 2022-06-23
Update HL7Monitor asset integration

iamsdas:
fixes: #801 
related to: https://github.com/coronasafe/care_fe/issues/2159

### DIFF
--- a/care/facility/api/viewsets/asset.py
+++ b/care/facility/api/viewsets/asset.py
@@ -33,6 +33,7 @@ from care.facility.models.asset import (
 )
 from care.users.models import User
 from care.utils.assetintegration.asset_classes import AssetClasses
+from care.utils.assetintegration.base import BaseAssetIntegration
 from care.utils.cache.cache_allowed_facilities import get_accessible_facilities
 from care.utils.filters.choicefilter import CareChoiceFilter, inverse_choices
 from care.utils.queryset.asset_location import get_asset_location_queryset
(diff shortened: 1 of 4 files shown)


care/facility/api/viewsets/asset.py
```diff
@@ -188,10 +188,10 @@ def operate_assets(self, request, *args, **kwargs):
                 raise ValidationError({"asset_id": "is required"})
             if "action" not in request.data:
                 raise ValidationError({"action": "is required"})
-            asset_id = request.data["asset_id"]
             action = request.data["action"]
             asset: Asset = self.get_object()
-            result = AssetClasses(asset.asset_class).handle_action(action)
+            asset_class = AssetClasses[asset.asset_class].value(asset.meta)
```
vigneshhari, line 193: Handle unknown asset class in this method

vigneshhari:
@gigincg please review.

care/facility/api/viewsets/asset.py
iamsdas: The error details are only returned when there is some validation error (post request body has missing/invalid parameters) with a status of 400. In the case of all other errors, only a message of "Internal server error" is returned with a status of 500. So, will there be any point in raising a separate new error? In any case, it is all inside a try-except block.

vigneshhari: ideally, internal server errors are reserved for errors that are unknown in nature. since these are known exceptions we try to handle as much as possible.

PR #902, opened 2022-06-30
Added discharge_reason and discharge_notes in Patient Consultation

khavinshankar:
Refer: https://github.com/coronasafe/care_fe/issues/2385


care/facility/api/viewsets/patient.py
```diff
@@ -343,9 +343,17 @@ def discharge_patient(self, request, *args, **kwargs):
         last_consultation = PatientConsultation.objects.filter(patient=patient).order_by("-id").first()
         current_time = localtime(now())
         if last_consultation:
+            reason = request.data.get("discharge_reason")
+            notes = request.data.get("discharge_notes", "")
+            if not reason:
+                raise serializers.ValidationError(
+                    {"discharge_reason": "discharge reason is mandatory"}
+                )
+            last_consultation.discharge_reason = reason
```
vigneshhari, line 352: Shouldn't we perform some validation to ensure that the value is within the choices we specified?

care/facility/api/serializers/patient_consultation.py
```diff
@@ -51,9 +51,13 @@ class PatientConsultationSerializer(serializers.ModelSerializer):
         queryset=User.objects.all(), required=False, allow_null=True
     )
 
+    discharge_reason = serializers.ChoiceField(
```
vigneshhari, line 54: This should be a read-only field, same goes with the discharge notes field.

vigneshhari (changes requested):

vigneshhari (approved):

PR #906, opened 2022-07-02
updated message when token is not valid

Marmik2003:
fixes https://github.com/coronasafe/care_fe/issues/2964

### DIFF
--- a/config/authentication.py
+++ b/config/authentication.py
@@ -16,6 +16,18 @@ class CustomJWTAuthentication(JWTAuthentication):
     def authenticate_header(self, request):
         return ""
 
+    def get_validated_token(self, raw_token):
+        try:
+            return super().get_validated_token()
+        except Exception as e:
+            print(e)
+        raise InvalidToken(
+            {
+                "detail": "Invalid Token, please relogin to continue",
+                "messages": [],
+            }
+        )
+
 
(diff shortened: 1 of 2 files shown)


sainAk (changes requested):
This auth class is used for middlewares its not used by the front-end to change the message, You need to override  `get_validated_token` in `CustomJWTAuthentication` where you need to catch `InvalidToken` and re raise it with a different messege.

sainAk (changes requested):
Override this auth class
https://github.com/coronasafe/care/blob/master/config/settings/base.py#L302

config/authentication.py
```diff
@@ -16,6 +16,18 @@ class CustomJWTAuthentication(JWTAuthentication):
...
+        try:
+            return super().get_validated_token()
+        except Exception as e:
+            print(e)
+        raise InvalidToken(
+            {
+                "detail": "Invalid Token, please relogin to continue",
+                "messages": [],
+            }
+        )
```
sainAk, line 29:
```suggestion
            return super().get_validated_token(raw_token)
        except InvalidToken as e:
            raise InvalidToken({
                "detail": "Invalid Token, please relogin to continue",
                "messages" : e.detail.get("messages", [])
            }) from e
```

sainAk (approved):

vigneshhari (approved):

PR #909, opened 2022-07-04
update asset integration actions

iamsdas:
- minor bug fixes
- sort out misplaced actions between hl7 and onvif.

### DIFF
--- a/care/utils/assetintegration/base.py
+++ b/care/utils/assetintegration/base.py
@@ -1,6 +1,7 @@
 import json
 
 import requests
+from django.conf import settings
 from rest_framework.exceptions import APIException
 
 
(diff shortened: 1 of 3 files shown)


care/utils/assetintegration/base.py
```diff
@@ -14,7 +14,7 @@ def handle_action(self, action):
         pass
 
     def get_url(self, endpoint):
-        return "http://{}/{}".format(self.middleware_hostname, endpoint)
+        return "https://{}/{}".format(self.middleware_hostname, endpoint)
```
vigneshhari, line 17: Are you sure this would work in a local env?

iamsdas: Maybe not, but using http only was giving me unexpected errors while testing with the dev camera.

vigneshhari: move it to an env variable then

iamsdas: Instead of an environment variable, I added the variable to the asset metadata with it defaulting to https as it makes more sense to have the connection type set per asset.

vigneshhari: ideally, we would not want any data to be sent over non HTTPS in production (especially medical data)

iamsdas: Added a check to prevent sending insecure requests in production mode.

vigneshhari (approved):

PR #915, opened 2022-07-08
Added custom filter to filter out patient based on single date

### LINKED ISSUE #913: Add custom filter to filter out patients based on a single query date params
### Overview
Filter out the patient by using a date parameter passed which gives only the patients that are present in the hospital at that date i.e 
"AdmittedDate <= FilterDate <= DischargeDate"

ishanExtreme:
### Changes
fixes #913 
Added Custom filter "admit_date" which can accept a single date and will filter user if "Admission Date <= Filter Date <= Discharge Date"

fixes this([#1789](https://github.com/coronasafe/care_fe/issues/1789)) issue at frontend.


skks1212 (approved):

vigneshhari (changes requested):
The logic does not seem right to me, the filter naming also seems wrong

PR #916, opened 2022-07-08
Fixed bed update error

skks1212:
closes [https://github.com/coronasafe/care_fe/issues/3048](https://github.com/coronasafe/care_fe/issues/3048)

### DIFF
--- a/care/facility/models/bed.py
+++ b/care/facility/models/bed.py
@@ -40,7 +40,7 @@ class Bed(BaseModel):
         return self.name
 
     def validate(self) -> None:
-        if Bed.objects.filter(location=self.location, name=self.name).exists():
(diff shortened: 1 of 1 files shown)


care/facility/models/bed.py
```diff
@@ -40,7 +40,7 @@ def __str__(self):
         return self.name
 
     def validate(self) -> None:
-        if Bed.objects.filter(location=self.location, name=self.name).exists():
+        if Bed.objects.filter(location=self.location, name=self.name).exclude(pk=self.pk).exists():
```
vigneshhari, line 43: Finally, something that the todo app taught!

vigneshhari (approved):

care/facility/models/bed.py
skks1212: 😅

PR #929, opened 2022-07-16
Dummy data seeding

Marmik2003:


vigneshhari:
The dummy data has a lot of real names and phone numbers, let's make it anonymized.
This method does not account for migrations that might happen later, we need a more dynamic seeder.

vigneshhari:
Also delete the commit that adds the file for the sake of maintaining anonymity

PR #935, opened 2022-07-19
add schema for validating meta field of assets

iamsdas:
fixes: #799


iamsdas:
We will be needing another migration to handle the existing invalid meta objects. Before I work on that, can someone confirm the correctness of this implementation?

vigneshhari:
ignore the existing values since we cant really do anything in those cases.

care/facility/models/asset.py
```diff
@@ -89,6 +91,10 @@ class Meta:
     def __str__(self):
         return self.name
 
+    def save(self, *args, **kwargs) -> None:
+        self.clean_fields()
```
vigneshhari, line 95: Why is this required? also where is this defined?

iamsdas: The validators defined using `validators=[JSONFieldSchemaValidator(ASSET_META)]` in the model fields are not run by default while saving the model. So to enforce the schema, we need to manually call these validators. This is done using the `clean_fields` method (it's inbuilt in the Django ORM).

vigneshhari: We did not require this in the daily round models, not sure about this thou.

iamsdas: It turns out this validation is handled by the `ModelSerializer`. So all the changes made using the API calls should be validated and only the direct/non-api changes won't be validated. I hadn't run the migration initially while testing, making me overthink the issue.

So yeah, we don't really need it.

vigneshhari (approved):

PR #939, opened 2022-07-23
Fixed: Added in_use field to track used bed

### LINKED ISSUE #936: Issue: In Use beds present in bedList
# Issue

## Problem

- [x] Related to coronasafe/care_fe#3149
- [x] Hide the used beds from available beds available for consultation

## Solution

- [ ] Add the `assigned_to` field that will store the patient object

cp-Coder:
# Updates

- [x] Fixes #936
- [x] Added `in_use` field to Bed model to track used bed
- [x] Added migration function to update `in_use` field for previous bed models
- [x] Updated serializer to update `in_use` field during bed update


vigneshhari:
do we need an `in_use` attribute ? `end_date == None` serves the same purpose right?

cp-Coder:
@vigneshhari There is a issue coronasafe/care_fe#3217. To fix this issue, I have added a field  in Bed Model

vigneshhari:
@cp-Coder you can use the end_date is null to check if a bed is in use.

cp-Coder:
@vigneshhari end_date is present in ConsultationBed. In the issue, user should be able to see only those beds (in the list present in the switch bed form) which are not in use. If in_use field is not present, then I have to query every bed of that facility in ConsultationBed model too and check whether it is still in use or not. This will increase the no of query to the database.

Issue #938, opened 2022-07-21
Permission Revamp: Write Access only for home_facility for users below DistrictAdmin. Read Access for all linked facilities. Exception for Doctors Notes, Users of type Doctor can make Doctor Notes on Patients in Linked Facilities


gigincg:

gigincg:
Solved in #922

PR #946, opened 2022-07-25
update permissions for write access

iamsdas:
fixes: #938

### DIFF
--- a/care/facility/models/mixins/permissions/facility.py
+++ b/care/facility/models/mixins/permissions/facility.py
@@ -57,7 +57,10 @@ class FacilityPermissionMixin(BasePermissionMixin):
             or request.user.user_type == User.TYPE_VALUE_MAP["StaffReadOnly"]
         ):
             return False
-        if request.user.user_type >= User.TYPE_VALUE_MAP["DistrictAdmin"]:
+        if (
+            request.user.user_type >= User.TYPE_VALUE_MAP["DistrictAdmin"]
+            or request.user.user_type == User.TYPE_VALUE_MAP["Staff"]
+        ):
             return self.has_object_read_permission(request)
         return request.user.home_facility == self
 
--- a/care/facility/models/mixins/permissions/facility.py
+++ b/care/facility/models/mixins/permissions/facility.py
@@ -57,9 +57,9 @@ class FacilityPermissionMixin(BasePermissionMixin):
             or request.user.user_type == User.TYPE_VALUE_MAP["StaffReadOnly"]
         ):
             return False
-        if request.user.user_type < User.TYPE_VALUE_MAP["Staff"]:  # todo Temporary
-            return False
-        return self.has_object_read_permission(request)
+        if request.user.user_type >= User.TYPE_VALUE_MAP["DistrictAdmin"]:
+            return self.has_object_read_permission(request)
+        return request.user.home_facility == self
 
     def has_object_update_permission(self, request):
         return super().has_object_update_permission(request) or self.has_object_write_permission(request)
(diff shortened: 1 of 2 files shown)


Marmik2003 (approved):

care/facility/models/mixins/permissions/patient.py
```diff
@@ -148,7 +149,10 @@ def has_object_update_permission(self, request):
...
             request.user.user_type == User.TYPE_VALUE_MAP["DistrictReadOnlyAdmin"]
             or request.user.user_type == User.TYPE_VALUE_MAP["StateReadOnlyAdmin"]
-            or request.user.user_type == User.TYPE_VALUE_MAP["StaffReadOnly"]
         ):
             return False
-        return self.has_object_read_permission(request)
+        return (
+            self.has_object_read_permission(request)
+            and request.user.user_type == User.TYPE_VALUE_MAP["Doctor"]
+            or request.user.user_type >= User.TYPE_VALUE_MAP["DistrictAdmin"]
```
vigneshhari, line 157: This overrides the check for dist admins right? `False and False or True is True`

care/facility/models/mixins/permissions/patient.py
```diff
@@ -125,8 +125,9 @@ def has_write_permission(request):
             return False
         return (
             request.user.is_superuser
-            or request.user.verified
-            and request.user.user_type >= User.TYPE_VALUE_MAP["Staff"]
+            or request.user.user_type >= User.TYPE_VALUE_MAP["DistrictAdmin"]
+            or request.user.user_type == User.TYPE_VALUE_MAP["Doctor"]
+            and request.user.verified
```
vigneshhari, line 130: lets remove verified wherever present

care/facility/models/mixins/permissions/facility.py
```diff
@@ -57,9 +57,9 @@ def has_object_write_permission(self, request):
             or request.user.user_type == User.TYPE_VALUE_MAP["StaffReadOnly"]
         ):
             return False
-        if request.user.user_type < User.TYPE_VALUE_MAP["Staff"]:  # todo Temporary
-            return False
-        return self.has_object_read_permission(request)
+        if request.user.user_type >= User.TYPE_VALUE_MAP["DistrictAdmin"]:
+            return self.has_object_read_permission(request)
```
vigneshhari, line 61: staff also needs to be checked with the `self.has_object_read_permission(request)` since staff can also have access to a facility

vigneshhari (changes requested):

care/facility/models/mixins/permissions/facility.py
iamsdas: Staff doesn't have write permission, right? According to the issue description, all users below the DistrictAdmin level can only edit their home facility (including the staff).  So I am checking the admin's access inside the if condition. Rest are being checked for their home facility.

care/facility/models/mixins/permissions/patient.py
iamsdas: Yup, I messed up the logic 😅

care/facility/models/mixins/permissions/facility.py
vigneshhari: you can keep the existing permissions, staff can edit facilities that they have access to

vigneshhari:
Why the force pushes ?

iamsdas:
> Why the force pushes ?

I thought that it would have been better to have a single commit for smaller changes.

care/facility/models/mixins/permissions/facility.py
iamsdas: Updated it. Staff now have write permission to all their linked facilities. Also wanted to confirm if similar roles like the WardAdmin or LocalBodyAdmin require these additional write permissions.

CC: @gigincg

vigneshhari:
All commits will end up being squashed to one, so no worries! 
Don't let the dark side tempt you! Use the force wisely!!

vigneshhari (approved):

Issue #898, opened 2022-06-27
Create Constraint not to allow multiple patients to be allotted to the same bed


gigincg:
> For migrating existing data, if there is a conflict, the patient who has been admitted early on should be discharged at the time of the conflicting entry

cc: @aparnacoronasafe @vigneshhari

gigincg:
Resolved in https://github.com/coronasafe/care/pull/854

PR #950, opened 2022-07-26
Added constraints in models to allow one patient per bed

siddnikh:
Solves #898


vigneshhari:
Why are there no migrations?

PR #953, opened 2022-07-29
add ONVIF action to get streaming token

iamsdas:
Add ONVIF asset action to get streaming token from middleware server.

Note: API is yet to be implemented in the middleware.


vigneshhari (approved):

vigneshhari:
Merge when needed.

Issue #965, opened 2022-08-05
Clear current_bed on Discharge


gigincg:
The current_bed for a patient is reflecting for a new patient_consultation. This needs to be fixed

PR #971, opened 2022-08-09
clear current_bed on discharge

iamsdas:
fixes: #965

Clear `current_bed` attribute of patient consultation when discharged.


vigneshhari (approved):

vigneshhari:
shouldn't we set the end date for the consultation bed as well ?

iamsdas:
@vigneshhari It is already being updated over [here](https://github.com/coronasafe/care/blob/d56b6b0034a23ec70cc66a5a73aa11b4d8a5f45f/care/facility/api/viewsets/patient.py#L365). (We are filtering and updating it if exists)

Issue #973, opened 2022-08-10
PATCH, PUT, DELETE Api needed for Consultaion Files 


patelaryan7751:
# PATCH, PUT, DELETE Api needed for Consultaion Files for 

 - https://github.com/coronasafe/care_fe/issues/3367 
 - https://github.com/coronasafe/care_fe/issues/2288

PR #981, opened 2022-08-17
add update and delete apis for consultation files

iamsdas:
## Proposed Changes

- Create APIs for updating and deleting the consultation files.
- Improve error handling for invalid requests (initially giving internal server errors).

@coronasafe/code-reviewers
fixes: #973

### DIFF
--- a/care/facility/api/viewsets/file_upload.py
+++ b/care/facility/api/viewsets/file_upload.py
@@ -14,6 +14,7 @@ from care.facility.api.serializers.file_upload import (
     FileUploadCreateSerializer,
     FileUploadListSerializer,
     FileUploadRetrieveSerializer,
+    FileUploadUpdateSerializer,
     check_permissions,
 )
 from care.facility.models.file_upload import FileUpload
@@ -37,8 +38,10 @@ class FileUploadViewSet(
             return FileUploadRetrieveSerializer
         elif self.action == "list":
             return FileUploadListSerializer
-        else:
+        elif self.action =="create":
             return FileUploadCreateSerializer
+        else:
+            return FileUploadUpdateSerializer
 
     def get_queryset(self):
         if "file_type" not in self.request.GET:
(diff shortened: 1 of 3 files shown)


care/facility/models/file_upload.py
```diff
@@ -47,14 +47,13 @@ class FileCategory(enum.Enum):
...
-                if len(parts) > 1:
-                    internal_name = internal_name + "." + parts[-1]
-            self.internal_name = internal_name
-            return super().save(*args, **kwargs)
+        internal_name = str(uuid4()) + str(int(time.time()))
+        if self.internal_name:
+            parts = self.internal_name.split(".")
+            if len(parts) > 1:
+                internal_name = internal_name + "." + parts[-1]
+        self.internal_name = internal_name
```
vigneshhari, line 55: internal name should not be changed at any point

care/facility/api/viewsets/file_upload.py
```diff
@@ -18,7 +24,7 @@ class FileUploadFilter(filters.FilterSet):
 
 
 class FileUploadViewSet(
-    CreateModelMixin, RetrieveModelMixin, ListModelMixin, GenericViewSet,
+    CreateModelMixin, RetrieveModelMixin, ListModelMixin, UpdateModelMixin, DestroyModelMixin, GenericViewSet,
```
vigneshhari, line 27: destroy should destroy the file from AWS

vigneshhari, line 27: The update should use a different serializer and not update anything to AWS.

vigneshhari (changes requested):

care/facility/api/viewsets/file_upload.py
iamsdas: In the issue it was written that we needed soft deletion for the files, in that case we shouldn't delete the file from AWS right?

vigneshhari: soft delete in the database might be what he meant, the file needs to be deleted still ( Because it makes 0 sense to keep a file with no reference and containing actual confidential data ).
cc @gigincg

iamsdas: updated it

vigneshhari: @iamsdas had a conversation with the team, lets do the soft delete itself, they want it for audit reasons.

vigneshhari (approved):

Issue #986, opened 2022-08-21
Pressure Sore: Enhancement (PUSH score)


arpancodes:
Refer: https://github.com/coronasafe/care_fe/issues/3432
Need to add 
- [ ] Length - number
- [ ] Width - number
- [ ] Exudate Amount - string, ENUM: None, Light, Moderate, Heavy
- [ ] Tissue Type - String, ENUM: Closed, Epithelial tissue, Granulation Tissue, Slough, Necrotic tissue
- [ ] Description - String
- [ ] Push score - number, calculated on the backend 

scores should be based on the following table:
![image](https://user-images.githubusercontent.com/42417893/185783545-2c7ab666-667d-4442-bbc7-65f40d360d4c.png)

aparnacoronasafe:
Clarification:

A suggestive design would be something like this
<img width="734" alt="Screenshot 2022-08-26 at 8 18 31 PM" src="https://user-images.githubusercontent.com/67050716/186932068-ae57c78e-6717-40b5-834f-ed5c93369a8d.png">

PR #987, opened 2022-08-21
add fields for push score

iamsdas:
## Proposed Changes

Add fields required for push score calculation in daily rounds:
- Length - number
- Width - number
- Exudate Amount - ENUM
- Tissue Type - ENUM
- Description - String
- Push score - number, calculated on the backend

fixes: #986 

@coronasafe/code-reviewers


care/facility/models/daily_round.py
```diff
@@ -418,6 +439,15 @@ class InsulinIntakeFrequencyType(enum.Enum):
         default=list, validators=[JSONFieldSchemaValidator(NURSING_PROCEDURE)]
     )
 
+    length = models.FloatField(default=0.0, validators=[MinValueValidator(0)])
```
vigneshhari, line 442: As per https://github.com/coronasafe/care_fe/issues/3432, each section must have these values, so we might have to dump these as a json. storing these as columns only allows us to store one instance of the value.

vigneshhari: @aparnacoronasafe Discuss with @iamsdas to understand exactly what needs to be stored and how it needs to be represented.  ( I might be wrong and this might be correct )

vigneshhari (changes requested):

iamsdas:
Closing this PR, as will make a new one from the beginning.

care/facility/models/daily_round.py
aparnacoronasafe: @iamsdas @vigneshhari Yes, there will be multiple windows with these parameters, one for each pressure sore. Added clarification on the issue

PR #990, opened 2022-08-27
Add investigation field in patient consultation

skks1212:
Added a new investigation JSON field to patient consultation model as requested in [https://github.com/coronasafe/care_fe/issues/3435](https://github.com/coronasafe/care_fe/issues/3435)


vigneshhari:
Is this in contrast to the investigations module we already have?

skks1212:
> Is this in contrast to the investigations module we already have?

Nope this is different, it will be used to provide nurses information on what investigations should be taken, how and when.

vigneshhari (approved):

PR #1003, opened 2022-09-05
add dry permissions for facility cover_image

sainak:
Fix was to add object level dry permissions for cover_image route and relaxing model level permissions as we weren't sending district, state or lsb fields

The code was also refactored to improve api documentation

### DIFF
--- a/care/facility/api/serializers/facility.py
+++ b/care/facility/api/serializers/facility.py
@@ -1,17 +1,18 @@
-from distutils import extension
 import boto3
-
-from django.contrib.auth import get_user_model
 from django.conf import settings
+from django.contrib.auth import get_user_model
 from rest_framework import serializers
 
-from care.facility.api.serializers.facility_capacity import FacilityCapacitySerializer
 from care.facility.models import FACILITY_TYPES, Facility, FacilityLocalGovtBody
 from care.facility.models.facility import FEATURE_CHOICES
-from care.facility.models.patient_base import reverse_choices
-from care.users.api.serializers.lsg import DistrictSerializer, LocalBodySerializer, StateSerializer, WardSerializer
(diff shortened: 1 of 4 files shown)


care/facility/models/mixins/permissions/facility.py
```diff
@@ -34,6 +38,11 @@ def has_write_permission(request):
         except Exception:
             return False
 
+    @staticmethod
+    def has_cover_image_permission(request):
+        # Returning true here as the permission is validated at object level for this action
+        return True
+
```
sainak, line 45: @vigneshhari  is there any edge case that I'm missing?

sainak:
https://github.com/coronasafe/care_fe/issues/2171

rithviknishad:
coronasafe/care_fe#3528

care/facility/api/serializers/facility.py
```diff
@@ -75,7 +79,7 @@ class FacilitySerializer(FacilityBasicInfoSerializer):
     #     "latitude": 49.8782482189424,
     #     "longitude": 24.452545489
     # }
-    read_cover_image_url = serializers.CharField(read_only=True)
```
vigneshhari, line 78: lets keep this as a Charfield

vigneshhari: Lets store file name and domain for the bucket separately

vigneshhari (changes requested):

care/facility/api/serializers/facility.py
sainak: converting it to URLfield will fix the api schema

vigneshhari: My bad, i thought it was a model change

PR #1006, opened 2022-09-07
Deprecates text field `warranty_details` of Asset

rithviknishad:
Follow up of #984 

Marks text field warranty_details as deprecated.

**Reason:**
More specific fields related to warranty was added in #984


care/facility/models/asset.py
```diff
@@ -61,6 +61,9 @@ class Status(enum.Enum):
     not_working_reason = models.CharField(max_length=1024, blank=True, null=True)
     serial_number = models.CharField(max_length=1024, blank=True, null=True)
     warranty_details = models.TextField(null=True, blank=True, default="")
+    """
+    Deprecated as fields that are more specific related to warranty are present.
+    """
```
sainak, line 66:
```suggestion
    warranty_details = models.TextField(null=True, blank=True, default="") # Deprecated
```

vigneshhari (approved):

PR #1013, opened 2022-09-08
delete cached asset object only on update

sainak:
## Proposed Changes

delete cached asset object only on update

closes #1012 
fixes https://github.com/coronasafe/care_fe/issues/3561

@coronasafe/code-reviewers

## Merge Checklist

- [ ] Add specs that demonstrate bug / test a new feature.
- [ ] Update [product documentation](https://docs.coronasafe.network/coronasafe-care-documentation/architecture/architecture-and-layering-of-care).
- [ ] Ensure that UI text is kept in I18n files.
- [ ] Prep screenshot or demo video for changelog entry, and attach it to issue.
- [ ] Request for Peer Reviews
- [ ] Completion of QA

PR #1017, opened 2022-09-09
Clear asset cache only when updating an existing asset

Ashesh3:
Fixes https://github.com/coronasafe/care_fe/issues/3561

This fixes a bug where it would try to clear the cache for a newly created asset on `save`. Now the cache is only deleted if the `save` was triggered on a pre-existing asset object.

@coronasafe/code-reviewers


sainak:
duplicate of #1013

PR #1019, opened 2022-09-09
return file extension in file_upload list and retrieve views

sainak:
## Proposed Changes

return file extension in file_upload list and retrieve views

@coronasafe/code-reviewers

## Merge Checklist

- [ ] Add specs that demonstrate bug / test a new feature.
- [ ] Update [product documentation](https://docs.coronasafe.network/coronasafe-care-documentation/architecture/architecture-and-layering-of-care).
- [ ] Ensure that UI text is kept in I18n files.
- [ ] Prep screenshot or demo video for changelog entry, and attach it to issue.
- [ ] Request for Peer Reviews
- [ ] Completion of QA

### DIFF
--- a/care/facility/api/serializers/file_upload.py
+++ b/care/facility/api/serializers/file_upload.py
@@ -49,7 +49,12 @@ def check_permissions(file_type, associating_id, user):
                     if user == sample.consultation.assigned_to:
                         return sample.id
             if sample.testing_facility:
-                if has_facility_permission(user, Facility.objects.get(external_id=sample.testing_facility.external_id)):
+                if has_facility_permission(
+                    user,
+                    Facility.objects.get(
+                        external_id=sample.testing_facility.external_id
+                    ),
+                ):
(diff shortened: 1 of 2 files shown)


vigneshhari (changes requested):
Let's not give the internal name out, our goal here was to break the relationship between the file name and the patient in the unlikely case our buckets are attacked. 
Let's serialize only the extension.

sainak:
> Let's not give the internal name out, our goal here was to break the relationship between the file name and the patient in the unlikely case our buckets are attacked. Let's serialize only the extension.

yeah I thought returning the extension would be a better solution
do you have any suggestions for implementing that?
I have 2 implementations in mind 
- making the query return the extension
- make a model method to return extension or file type

vigneshhari:
the model method sounds good enough. make sure you handle the edge cases.

vigneshhari (approved):

PR #1025, opened 2022-09-17
Added medicine_administration field

skks1212:
A JSON field to host data for Medicine Administration Record in Daily rounds, a new proposed feature as mentioned in [https://github.com/coronasafe/care_fe/issues/3437](https://github.com/coronasafe/care_fe/issues/3437)


care/facility/models/daily_round.py
```diff
@@ -435,6 +435,10 @@ class InsulinIntakeFrequencyType(enum.Enum):
         default=list, validators=[JSONFieldSchemaValidator(NURSING_PROCEDURE)]
     )
 
+    medicine_administration = JSONField(
```
vigneshhari, line 438: can we have a JSON schema for this as well? having a schemaless field might get hard for standardization at some point.

vigneshhari (changes requested):

care/facility/models/daily_round.py
skks1212: The JSON schema for this is going to be very complex. It has still not been finalized. Once it will, I will add the schema in the backend as another PR.

vigneshhari: okays, make an issue and tag it here even if this PR is merged

PR #1029, opened 2022-09-19
Expose upload_completed for FileUpload

Ashesh3:
## Proposed Changes

- This PR exposes the already present `upload_completed` attribute for the `FileUpload` model, this is now used to keep track for successful file uploads.
- The frontend will be updated to accordingly hide any incomplete file uploads and only show the ones with `upload_completed = True`
- This also allows `upload_completed` to be modified via a PATCH request, the frontend will now send a separate PATCH to set the said field to `True` when the file upload is successful.
- A migration is also included which sets `upload_completed` to `True` for all existing files.

### Associated Issue
  - https://github.com/coronasafe/care_fe/issues/3582
 
@coronasafe/code-reviewers


care/facility/migrations/0319_file_upload_completed.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('facility', '0318_merge_20220918_1241'),
+    ]
+
+    operations = [
+        migrations.RunSQL(
```
vigneshhari, line 13: can we use the orm here?

vigneshhari (changes requested):

care/facility/migrations/0319_file_upload_completed.py
Ashesh3: Should be done now, Reference used: https://docs.djangoproject.com/en/4.0/ref/migration-operations/#runpython

vigneshhari (approved):

PR #1038, opened 2022-09-27
Added constraints to models to not allot multiple patients the same bed

siddnikh:
## Proposed Changes

- Added constraint in ConsultationBed model so no 2 patients can be allotted the same bed. 
- Added constraint in PatientConsultation model so no 2 patients can be allotted the same bed provided both of them are admitted.

### Associated Issue
  Issue #898


care/facility/models/bed.py
```diff
@@ -65,3 +65,10 @@ class ConsultationBed(BaseModel):
     start_date = models.DateTimeField(null=False, blank=False)
     end_date = models.DateTimeField(null=True, blank=True, default=None)
     meta = JSONField(default=dict, blank=True)
+
+    class Meta:
+        constraints = [
```
vigneshhari, line 70: Let's take a case where a patient starts off in a regular bed, moves to an ICU, and then moves back to the normal bed, this is practically possible right?

vigneshhari: The correct solution should check if there is any patient currently using the bed.

vigneshhari (changes requested):

care/facility/models/bed.py
siddnikh: So, do you suggest I modify the api route when assigning patients to bed *along with* the added constraints?

siddnikh:
@vigneshhari I tried adding a patient to the same bed, except we don't get that bed listed when we're trying to add another patient.

gigincg:
Already Resolved in https://github.com/coronasafe/care/pull/854

cc: @vigneshhari

Issue #989, opened 2022-08-23
Check availability of username


aparnacoronasafe:
Build a test to check availability of username to be placed in the User creation page

Ashesh3:
@siddnikh Hey, are you currently working on this PR?

siddnikh:
@Ashesh3 hi! No, I'm not. Please go ahead.

rithviknishad:
@siddnikh use closing keywords when linking PRs with issues when the PR achieves the requirements in the issue.

PR #1043, opened 2022-09-30
Added API View to check availability of username

siddnikh:
## Proposed Changes
- Added checkusername API View

### Associated Issue
  - #989


care/users/api/viewsets/users.py
```diff
@@ -286,3 +286,23 @@ def pnconfig(self, request, *args, **kwargs):
...
+    Checks availability of username, returns 200 if available, and 400 otherwise.
+    """
+
+    permission_classes = (AllowAny, )
+    serializer_class = CheckUsernameSerializer
+
+    def get(self, request):
+        serializer = self.serializer_class(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        username = serializer.validated_data['username']
```
sainak, line 304: we could use query parameters here.
get method don't support body according to specs.

care/users/api/viewsets/users.py
```diff
@@ -286,3 +286,23 @@ def pnconfig(self, request, *args, **kwargs):
...
+        user = User.objects.filter(username=username) # since users are soft deleted, there might be inactive/deleted users that had the same username
+        if user.exists():
+            return Response(status=status.HTTP_400_BAD_REQUEST)
```
vigneshhari, line 307: use 409 conflict.

vigneshhari (changes requested):

sainak (approved):

care/users/api/viewsets/users.py
```diff
@@ -286,3 +286,17 @@ def pnconfig(self, request, *args, **kwargs):
                 setattr(user, field, request.data[field])
         user.save()
         return Response(status=status.HTTP_200_OK)
+
+class CheckUsernameAPIView(GenericAPIView):
+    """
+    Checks availability of username by getting as query, returns 200 if available, and 409 otherwise.
+    """
+
+    permission_classes = (AllowAny, )
```
vigneshhari, line 295: Can we switch this to authenticated users only? Having an open endpoint to guess usernames might be a potential attack vector. Since users creation is an authenticated process for us, it makes sense in that way as well.

care/users/api/viewsets/users.py
```diff
@@ -286,3 +286,17 @@ def pnconfig(self, request, *args, **kwargs):
...
+
+    def get(self, request):
+        username = request.query_params.get('username')
+        user = User.objects.filter(username=username) # since users are soft deleted, there might be inactive/deleted users that had the same username
```
vigneshhari, line 299: You can delete the comment, since deleted users are also users in our system since audit logs are tied to their username, so we should NOT allow the username to be reused in any case.

vigneshhari (changes requested):

PR #1055, opened 2022-10-07
Adds `archive_reason` to `FileUpload` and validate if `is_archived`

### LINKED ISSUE #1054: Field for having a reason why the file is archived in the consultation files api
**Describe the bug**
Field for having a reason why the file is archived in the consultation files api as per https://github.com/coronasafe/care_fe/issues/3366#issue-1334061185
a field for **is_archived** has been created before in this PR https://github.com/coronasafe/care/pull/1052
a filed for **reason why it is archived** need to be created the field should be mandatory while archieving the file there should be madatorily a reason why the file is being arachived without reason user cannot archive the file

This issue is for the PR https://github.com/coronasafe/care_fe/pull/3716

rithviknishad:
Closes #1054 
Required by https://github.com/coronasafe/care_fe/issues/3366

## Proposed Changes

- Adds non-mandatory text field `archive_reason`
- Validate `archive_reason` to be mandatory if `is_archived` is set.
- Raises validation error if an update is attempted after being archived.

image

 
@coronasafe/code-reviewers

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete


vigneshhari:
We should also not let them update the archive reason. I believe that was the conclusion from the last call with @aparnacoronasafe , Do confirm before you make this change.

aparnacoronasafe:
> We should also not let them update the archive reason. I believe that was the conclusion from the last call with @aparnacoronasafe , Do confirm before you make this change.

I agree with vig

rithviknishad:
@vigneshhari I've made the changes required.

vigneshhari (approved):

PR #1076, opened 2022-10-24
fix(Patient): Updated review_time logic

### LINKED ISSUE #1075: update review_time logic
- When a patient is discharged set `review_time` to null
- Also when `review_interval` is -1 in log update or consultation, set `review_time` to null

ref: https://github.com/coronasafe/care_fe/pull/3815#issuecomment-1288552508

cp-Coder:
# Updates

- [x] Fixes #1075
- [x] Updated `review_time` logic

### DIFF
--- a/care/facility/api/serializers/patient_consultation.py
+++ b/care/facility/api/serializers/patient_consultation.py
@@ -152,8 +152,8 @@ class PatientConsultationSerializer(serializers.ModelSerializer):
             if "review_interval" in validated_data:
                 review_interval = validated_data.pop("review_interval")
                 instance.review_interval = review_interval
+                instance.save()
                 if review_interval >= 0:
-                    instance.save()
                     patient.review_time = localtime(now()) + timedelta(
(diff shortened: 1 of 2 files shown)


khavinshankar (approved):

care/facility/api/serializers/patient_consultation.py
```diff
@@ -149,12 +151,14 @@ def update(self, instance, validated_data):
...
                 review_interval = validated_data.pop("review_interval")
+                instance.review_interval = review_interval
                 if review_interval >= 0:
-                    instance.review_interval = review_interval
                     instance.save()
                     patient.review_time = localtime(now()) + timedelta(
                         minutes=review_interval
                     )
+                else:
+                    patient.review_time = None
```
khavinshankar, line 161: for `else` block the instance is not saved
bring `instance.save()` outside conditional block

khavinshankar (changes requested):

PR #1083, opened 2022-10-28
Added is_occupied read-only field in Bed

### LINKED ISSUE #1082: Add `is_occupied` read_only field to bed
Currently, there is no way to filter unoccupied beds in the front end, As of now we get an error response after submitting to the backend

khavinshankar:
Fixes #1082 

## Proposed Changes

- Added is_occupied read-only field in Bed
This is required to filter beds based on weather it is occupied or not

### Associated Issue
  - https://github.com/coronasafe/care_fe/pull/3690 (requirement mentioned here)
  - https://github.com/coronasafe/care_fe/issues/3842 (feature needed for this issue)

### Architecture changes
  - none
 
@coronasafe/code-reviewers

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete

### DIFF
--- a/care/facility/api/serializers/bed.py
+++ b/care/facility/api/serializers/bed.py
@@ -1,6 +1,6 @@
 from django.shortcuts import get_object_or_404
 from rest_framework.exceptions import ValidationError
-from rest_framework.serializers import ModelSerializer, UUIDField
+from rest_framework.serializers import BooleanField, ModelSerializer, UUIDField
 
 from care.facility.api.serializers import TIMESTAMP_FIELDS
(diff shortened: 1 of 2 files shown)


vigneshhari:
I would rather create a variable and store it in the database, this method fires a query per object retrieval, which can be DB intensive when multiple objects are rendered. @khavinshankar can you create a separate ticket for this. 
I will approve this on the condition that it will be improved later.

vigneshhari (approved):

PR #1093, opened 2022-11-08
Added is_occupied field in Bed

### LINKED ISSUE #1084: Create `is_occupied` property in Bed model
As of now, is_occupied is calculated for every fetch query, make a property in Bed model and store it in DB, so that it is not calculated on every fetch operation.

_Originally posted by @vigneshhari in https://github.com/coronasafe/care/issues/1083#issuecomment-1294737068_

khavinshankar:
## Proposed Changes

- added is_occupied field in Bed
- update is_occupied field when consultation bed is created (switch bed), patient consultation is created, and when the patient is discharged

### Associated Issue
  - Fixes #1084 

### Architecture changes
  - nil
 
@coronasafe/code-reviewers

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete


vigneshhari:
@rithviknishad @sainak review.

care/facility/migrations/0327_bed_is_occupied.py
```diff
@@ -0,0 +1,28 @@
...
+    ]
+
+    def update_previous_beds(apps, schema_editor):
+        Bed = apps.get_model("facility", "Bed")
+        ConsultationBed = apps.get_model("facility", "ConsultationBed")
+        for bed in Bed.objects.all():
+            bed.is_occupied = ConsultationBed.objects.filter(
+                bed=bed, end_date__isnull=True
+            ).exists()
+            bed.save()
```
rithviknishad, line 19: We could optimize the custom migration like this right?

```suggestion
        Bed.objects.all().update(
          is_occupied=F("consultationbed__end_date__isnull")
        )
```

care/facility/api/serializers/bed.py
```diff
@@ -179,10 +179,18 @@ def create(self, validated_data):
         if occupied_beds.filter(bed=bed).exists():
             raise ValidationError({"bed:": ["Bed already in use by patient"]})
 
+        Bed.objects.filter(
```
sainak, line 182: now that we have the value directly stored in the model we can directly get the object and check/update the status

care/facility/migrations/0327_bed_is_occupied.py
rithviknishad: or even:

```suggestion
        Bed.objects.filter(consultationbed__end_date__isnull=True).update(is_occupied=True)
```

sainak, line 19:
```suggestion
        Bed = apps.get_model("facility", "Bed")
        ConsultationBed = apps.get_model("facility", "ConsultationBed")
        objs = []
        for consultation_bed in ConsultationBed.objects.filter(end_date__isnull=True).select_related("bed"):
            consultation_bed.bed.is_occupied=True
            objs.append(consultation_bed.bed)
        Bed.objects.bulk_update(objs, ['is_occupied'])
           
```

sainak (changes requested):

Issue #1105, opened 2022-11-23
Compress/Resize Facility cover images


khavinshankar:
**Is your feature request related to a problem? Please describe.**
Currently, the facility cover images are huge and take too much time to load. 

**Describe the solution you'd like**
The images can be resized to a smaller fixed height and width.

**Additional context**
The resized images will not have any drawbacks as in the frontend we are anyways showing them within a container with max width and height less than 700px

ARH-MNAJS:
@vigneshhari I have understood the changes needed from the issue description and past PRs. If this issue is still valid i would like to work on it

bodhish:
@ARH-MNAJS when you request to assign its great if you can respond with your approach so that we are super clear what you have in your mind.

ARH-MNAJS:
@bodhish Sure here's a high-level approach to solve the issue effectively. I am planning to implement the changes with both frontend and backend solutions. The focus will be on ensuring images are appropriately resized/cropped before upload (for efficiency and user control) and validating them on the server (for safety and consistency).
### Front-End Image Cropping & Resizing
 - **Implement a Crop & Resize UI:** On the facility cover image upload form, introduce a simple image cropping tool. When a user selects an image, show a preview and allow them to crop to the desired aspect ratio (likely a square, since that seems to be the expected format)
 - **Enforce Recommended Aspect Ratio:** By using the cropping tool, we can enforce a square (1:1) aspect ratio for all cover images at the time of selection. This means the user will see a square crop box (which they can scale and move) on their image.
 - **Client-Side Resizing/Compression:** Once the user is happy with the crop, resize the image on the client side to a reasonable maximum size (for example, 700px or 720px width/height). This can be done in the browser (again via a canvas or a library) before uploading. By doing this:
    - We significantly reduce the file size that needs to be uploaded, saving bandwidth and time.
    - The user’s browser, not the server, does the heavy work of resizing, which reduces server load and speeds up the process.
    - We can also compress the image (e.g., adjust JPEG quality) to ensure it’s under the desired size threshold (1MB or even lower if possible) before sending.
 - **Preview to User:** The front-end should show the user a preview of how the image will look after crop and resize. This way, there are no surprises – the user knows exactly what will be uploaded and how it will appear as a cover image.
### Back-End Validation and Processing

> If someone bypasses the front-end (or for older browsers), the back-end will still handle large images. But for the vast majority of cases, the image will already be in the right format when it reaches the server.

[...]

PR #1106, opened 2022-11-23
Resize facility cover image before upload

khavinshankar:
## Proposed Changes

- resize facility cover images before uploading to s3
- set a maximum size limit of 1MB which is checked after resizing

### Associated Issue
  - Fixes https://github.com/coronasafe/care/issues/1105

### Architecture changes
  - nil
 
@coronasafe/code-reviewers

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete

### DIFF
--- a/care/facility/api/serializers/facility.py
+++ b/care/facility/api/serializers/facility.py
@@ -1,6 +1,9 @@
+from io import BytesIO
+
 import boto3
 from django.conf import settings
 from django.contrib.auth import get_user_model
+from PIL import Image
 from rest_framework import serializers
 
 from care.facility.models import FACILITY_TYPES, Facility, FacilityLocalGovtBody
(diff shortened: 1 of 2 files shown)


care/facility/api/serializers/facility.py
```diff
@@ -132,18 +135,41 @@ class Meta:
         # Check DRYpermissions before updating
         fields = ("cover_image", "read_cover_image_url")
 
+    def resize_image(self, image, image_extension="PNG"):
+        image_format = (
+            "JPEG" if image_extension.lower() == "jpg" else image_extension.upper()
+        )
+
+        img = Image.open(image)
+        img.thumbnail((720, 720))
```
vigneshhari, line 144: can we only allow images of this ratio to be uploaded in the FE? ( Square ) Anything not in this ration would look weird right?

care/facility/api/serializers/facility.py
```diff
@@ -132,18 +135,41 @@ class Meta:
         # Check DRYpermissions before updating
         fields = ("cover_image", "read_cover_image_url")
 
+    def resize_image(self, image, image_extension="PNG"):
+        image_format = (
+            "JPEG" if image_extension.lower() == "jpg" else image_extension.upper()
```
vigneshhari, line 140: why are we moving to uppercase?

vigneshhari (changes requested):

vigneshhari:
This PR will be marked stale and closed within the next 2 days.

gigincg:
@khavinshankar

sainak:
can we add a basic image tool to crop and resize the image, and then just check the size at the backend
@vigneshhari

Issue #1111, opened 2022-11-27
Add a filter to filter out archive files


cp-Coder:
# Feature Request

## Is your feature request related to a problem? Please describe

- It is difficult to only see the non-archive files for any patient

## Describe the solution you'd like

- Add a filter that can filter out the archive files

## Describe alternatives you've considered

- Didn't consider any other alternative

## Additional context

- N/A

PR #1112, opened 2022-11-27
feat(file_upload): added a filter to filter out the archive file

cp-Coder:
# Feature Request

## Proposed Changes

- Added a filter to filter out the archive files

## Associated Issue

- Fixes https://github.com/coronasafe/care/issues/1111

@coronasafe/code-reviewers

### DIFF
--- a/care/facility/api/viewsets/file_upload.py
+++ b/care/facility/api/viewsets/file_upload.py
@@ -22,12 +22,20 @@ from care.facility.models.file_upload import FileUpload
 
 class FileUploadFilter(filters.FilterSet):
     file_category = filters.CharFilter(field_name="file_category")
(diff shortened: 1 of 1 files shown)


care/facility/api/viewsets/file_upload.py
```diff
@@ -22,12 +22,20 @@
 
 class FileUploadFilter(filters.FilterSet):
     file_category = filters.CharFilter(field_name="file_category")
+    archive = filters.CharFilter(field_name="is_archived")
```
vigneshhari, line 25: This should be a boolean filter.

vigneshhari (changes requested):

vigneshhari (approved):

PR #1125, opened 2022-12-22
Added validations for create asset

### LINKED ISSUE #1124: Add Validations for create asset
Add validations to ensure that the `warranty_amc_end_of_validity` is not in the past and the `last_serviced_on` is not in the future when creating or updating an asset.
Related to [#4326](https://github.com/coronasafe/care_fe/issues/4326) in care_fe.

Pranshu1902:
Fixes #1124 
Added validations to ensure that the `warranty_amc_end_of_validity` is not in the past and the `last_serviced_on` is not in the future when creating or updating an asset.

### DIFF
--- a/care/facility/api/serializers/asset.py
+++ b/care/facility/api/serializers/asset.py
@@ -68,12 +68,12 @@ class AssetSerializer(ModelSerializer):
 
         # validate that warraty date is not in the past
         if "warranty_amc_end_of_validity" in attrs:
-            if attrs["warranty_amc_end_of_validity"] < datetime.now().strftime('%Y-%m-%d'):
+            if datetime.strptime(attrs["warranty_amc_end_of_validity"], '%Y-%m-%d') < datetime.now():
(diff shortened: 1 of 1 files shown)


care/facility/api/serializers/asset.py
```diff
@@ -65,6 +66,16 @@ def validate(self, attrs):
             del attrs["location"]
             attrs["current_location"] = location
 
+        # validate that warraty date is not in the past
+        if "warranty_amc_end_of_validity" in attrs:
+            if attrs["warranty_amc_end_of_validity"] < datetime.now().strftime('%Y-%m-%d'):
```
vigneshhari, line 71: Can you switch this to use actual DateTime comparison, this seems to be a string comparison

vigneshhari (changes requested):

care/facility/api/serializers/asset.py
Pranshu1902: Done

vigneshhari (approved):

PR #1144, opened 2023-01-05
Use configured middleware address for middleware authentication

sainak:
## Proposed Changes

Use configured middleware address for middleware authentication
 
@coronasafe/code-reviewers

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete

### DIFF
--- a/care/facility/api/serializers/facility.py
+++ b/care/facility/api/serializers/facility.py
@@ -119,6 +119,14 @@ class FacilitySerializer(FacilityBasicInfoSerializer):
         ]
         read_only_fields = ("modified_date", "created_date")
 
+    def validate_middleware_address(self, value):
+        value = value.strip()
+        if not value or not value.startswith("http"):
+            raise serializers.ValidationError("Invalid URL")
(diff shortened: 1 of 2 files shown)


care/facility/api/serializers/facility.py
```diff
@@ -119,6 +119,15 @@ class Meta:
         ]
         read_only_fields = ("modified_date", "created_date")
 
+    def validate_middleware_address(self, value):
+        if not value:
+            return value
```
vigneshhari, line 124: why is this not raising an error ?

vigneshhari (approved):

vigneshhari:
Are we no longer pointing to local if a middleware address is not present?

sainak:
> Are we no longer pointing to local if a middleware address is not present?

This change would give more info to the admin about whats going wrong.
we were pointing it to local for testing, we can still set the value to localhost in the configurations

vigneshhari (approved):

PR #1157, opened 2023-01-13
Restrict Admin from unlinking users from other district

### LINKED ISSUE #1152: User facility unlink permission issue.
**Describe the bug**
In user's section a DistrictAdmin can unlink linked facilities of other DistricAdmins from **other states**.

**To Reproduce**

1. Signin as a DistrictAdmin
2. Go to user's section
3. Find any other user from a different state and unlink facilities from its list.
4. See result

**Expected behavior**
A district admin of a particular state should be allowed to unlink facilities of users from its own district only.

**Recording**

https://user-images.githubusercontent.com/39593226/200363539-1789bde4-fd47-4ae5-ae5e-ecee240619e3.mp4

Pranshu1902:
Fixes #1152 
Ensure that the admin's district is same as that of the facility user's district being deleted and also able to home facilities only for users within same district.


care/users/api/viewsets/users.py
```diff
@@ -286,6 +286,9 @@ def delete_facility(self, request, *args, **kwargs):
             )
         if user.home_facility == facility:
             raise ValidationError({"facility": "Cannot Delete User's Home Facility"})
+        # ensure that district admin only able to delete in the same district
+        if user.district.name != requesting_user.district.name:
```
vigneshhari, line 290: use id for equality checks

vigneshhari (changes requested):

care/users/api/viewsets/users.py
Pranshu1902: Done

siddnikh:
LGTM

siddnikh (approved):
LGTM

care/users/api/viewsets/users.py
```diff
@@ -307,6 +307,9 @@ def delete_facility(self, request, *args, **kwargs):
             )
         if user.home_facility == facility:
             raise ValidationError({"facility": "Cannot Delete User's Home Facility"})
+        # ensure that district admin only able to delete in the same district
+        if user.district.id != requesting_user.district.id:
```
sainak, line 311:
```suggestion
        if user.district_id != requesting_user.district_id:
```

sainak (changes requested):

sainak:
@Pranshu1902 add tests

Pranshu1902:
@sainak @vigneshhari please review and merge the PR. I have added tests cases

rithviknishad (approved):

sainak:
@Pranshu1902 the test are not catching the changes, try to assert the error message response

Pranshu1902:
@sainak done

care/users/api/viewsets/users.py
```diff
@@ -280,6 +286,7 @@ def delete_facility(self, request, *args, **kwargs):
             )
         if user.home_facility == facility:
             raise ValidationError({"facility": "Cannot Delete User's Home Facility"})
+
```
sainak, line 289:
```suggestion
```

sainak (approved):

care/users/api/viewsets/users.py
```diff
@@ -253,6 +253,12 @@ def clear_home_facility(self, request, *args, **kwargs):
         if not self.has_user_type_permission_elevation(requesting_user, user):
             raise ValidationError({"home_facility": "Cannot Access Higher Level User"})
 
+        # ensure that district admin only able to delete in the same district
+        if user.district_id != requesting_user.district_id:
```
vigneshhari, line 257: A state admin need not have the same district as the requesting user right? Superadmin would not need any checks  as well.

vigneshhari (changes requested):

care/users/api/viewsets/users.py
Pranshu1902: @vigneshhari done, now this will only check for users equal to or below District Admin's position

sainak (approved):

rithviknishad (approved):

PR #1158, opened 2023-01-14
Add asset class to Asset filters

Ashesh3:
## Proposed Changes

Add `asset_class` to `AssetFilter`

### Associated Pulls/Issues
  - https://github.com/coronasafe/care_fe/pull/4596
  - https://github.com/coronasafe/care_fe/issues/4038
 
@coronasafe/code-reviewers

### DIFF
--- a/care/utils/assetintegration/onvif.py
+++ b/care/utils/assetintegration/onvif.py
@@ -6,7 +6,7 @@ from care.utils.assetintegration.base import BaseAssetIntegration
 
 
 class OnvifAsset(BaseAssetIntegration):
-    _name = "ONVIF"
+    _name = "onvif"
 
     class OnvifActions(enum.Enum):
         GET_CAMERA_STATUS = "get_status"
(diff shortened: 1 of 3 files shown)


care/utils/assetintegration/onvif.py
```diff
@@ -6,7 +6,7 @@
 
 
 class OnvifAsset(BaseAssetIntegration):
-    _name = "onvif"
```
vigneshhari, line 9: Can you confirm if these needs a data migration, i belive these are stored in the database and changing them would need a data migration as well.

vigneshhari (changes requested):

care/utils/assetintegration/onvif.py
Ashesh3: Hey, after reviewing this, these changes were not required and since `asset_class` is a CharField, using a `filters.CharFilter` works without having to make any additional changes. This should be resolved.

vigneshhari (approved):

PR #1160, opened 2023-01-20
Add facility_id filter to User

Ashesh3:
## Proposed Changes

- Allow facility_id to be used to filter users.

### Associated Issue/PR
  - Issue: https://github.com/coronasafe/care_fe/issues/4566
  - PR: https://github.com/coronasafe/care_fe/pull/4642

@coronasafe/code-reviewers

### DIFF
--- a/care/users/api/viewsets/users.py
+++ b/care/users/api/viewsets/users.py
@@ -70,6 +70,20 @@ class UserFilterSet(filters.FilterSet):
 
     user_type = filters.CharFilter(method="get_user_type", field_name="user_type")
 
+    def get_facilities(
+        self,
+        queryset,
+        field_name,
+        value,
+    ):
+        if value:
+            return queryset.filter(facilities__external_id=value)
+        return queryset
+
+    facility_id = filters.CharFilter(
+        method="get_facilities", field_name="facility_id", lookup_expr="exact"
+    )
+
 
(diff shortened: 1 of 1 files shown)


vigneshhari:
I would make this a separate api view, maybe under the facilities detailed view

Ashesh3:
> I would make this a separate api view, maybe under the facilities detailed view

Hi, I'm trying to understand your comment about making this a separate API view under the facilities detailed view. Are you suggesting that I create a new endpoint specifically for filtering users by facility and that this endpoint should be located under the detailed view of facilities? For example, `/facilities/<facility_id>/users`? Don't we already have a dedicated endpoint for this at: https://github.com/coronasafe/care/blob/master/care/users/api/viewsets/users.py#L218

Just want to make sure I understand your suggestion correctly before implementing it.

vigneshhari:
Filtering users by facility involves the use of "not-so-clean" SQL queries, we already have a facility users list API, and we can use that API to show the users from a given facility if needed. maybe as a separate UI page.

bodhish:
@vigneshhari we will want that filter for the page to be functional.

bodhish:
@Ashesh3 if we already have a page then lets not work on this now; I think I missed that part during the call. 

I was thinking you where creating a new page just to get around this filter issue.

PR #1173, opened 2023-02-01
fixed fetching user skills

### LINKED ISSUE #1172: Skills api is not working as expected
**Description of the bug**
When we update skills of a user, all users skills gets updated with the same skills

**To Reproduce**
1. add a new skill to a user
2. check skills of other users
3. you will find other user skills have also been updated

**Expected behavior**
Skills and users should have a many to many relationship and each user should have a separate set of skills

khavinshankar:
## Proposed Changes
- Fixed fetching user skills
Initially, skills associated with logged in user was being fetched, changed to user from URL param (`users/{user}/skills`)

### Associated Issue
  - Fixes #1172 

cc: @aparnacoronasafe @nihal467 @Pragati1610 
 
@coronasafe/code-reviewers

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete

### DIFF
--- a/care/users/api/viewsets/userskill.py
+++ b/care/users/api/viewsets/userskill.py
@@ -1,34 +1,26 @@
-from rest_framework import filters as drf_filters
+from django.db.utils import IntegrityError
+from rest_framework.exceptions import ValidationError
 from rest_framework.generics import get_object_or_404
 from rest_framework.viewsets import GenericViewSet, ModelViewSet
 
 from care.users.api.serializers.userskill import UserSkillSerializer
-from care.users.models import User, UserSkill
+from care.users.models import UserSkill
 from care.utils.queryset.user import get_users
-
-from django.db.utils import IntegrityError
-from rest_framework.exceptions import ValidationError
 
 
 class UserSkillViewSet(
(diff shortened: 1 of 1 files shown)


nihal467:
@gigincg @vigneshhari can you review it ASAP

care/users/api/viewsets/userskill.py
```diff
@@ -1,34 +1,26 @@
...
-            queryset = queryset.filter(user__state=user.state)
-        elif user.user_type >= User.TYPE_VALUE_MAP["DistrictLabAdmin"]:
-            queryset = queryset.filter(user__district=user.district)
-        else:
-            queryset = queryset.filter(user=user)
-        return queryset
+        username = self.kwargs["users_username"]
+        user = get_object_or_404(get_users(self.request.user).filter(username=username))
+
+        return self.queryset.filter(user=user)
```
vigneshhari, line 23: This will allow writes for all users, we need to allow reads only for unauthorized users.

vigneshhari (changes requested):

PR #1193, opened 2023-02-14
HCX Integration

khavinshankar:


care/hcx/api/viewsets/policy.py
```diff
@@ -0,0 +1,25 @@
...
+
+class PolicyViewSet(
+    CreateModelMixin,
+    DestroyModelMixin,
+    ListModelMixin,
+    RetrieveModelMixin,
+    UpdateModelMixin,
+    GenericViewSet,
+):
+    queryset = Policy.objects.all()
```
vigneshhari, line 22: No Authorisation Present

PR #1194, opened 2023-02-16
Add endpoint to clear user's home facility

Ashesh3:
This PR adds a new endpoint to the user API that allows users to clear their home facility. The new endpoint is implemented in the `clear_home_facility` method of the `UserViewSet` and can be accessed via a DELETE request.

### Associated Issue
  - https://github.com/coronasafe/care_fe/issues/4820

### Associated PR
  - https://github.com/coronasafe/care_fe/pull/4838
 
@coronasafe/code-reviewers

### DIFF
--- a/care/users/api/viewsets/users.py
+++ b/care/users/api/viewsets/users.py
@@ -265,6 +265,28 @@ class UserViewSet(
         },
     )
     @action(detail=True, methods=["DELETE"], permission_classes=[IsAuthenticated])
+    def clear_home_facility(self, request, *args, **kwargs):
+        user = self.get_object()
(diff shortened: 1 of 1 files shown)


vigneshhari:
if requesting user has permission to the selected user, do we need to check anything else?

Ashesh3:
> if requesting user has permission to the selected user, do we need to check anything else?

Correct, we do not need to check for facility access with the user again because the facility to be removed is not specified by the user, but rather it is the facility already associated with the user. I've made the appropriate changes regarding this.

gigincg (approved):

PR #1196, opened 2023-02-18
Allow skills access for any user

Ashesh3:
This PR allows any user to access the skills of any other user.

Fixes https://github.com/coronasafe/care_fe/issues/4830

### Associated Issue
  - https://github.com/coronasafe/care_fe/issues/4830
 
@coronasafe/code-reviewers

### DIFF
--- a/care/users/api/viewsets/userskill.py
+++ b/care/users/api/viewsets/userskill.py
@@ -4,7 +4,7 @@ from rest_framework.generics import get_object_or_404
 from rest_framework.viewsets import GenericViewSet, ModelViewSet
 
 from care.users.api.serializers.userskill import UserSkillSerializer
-from care.users.models import UserSkill
+from care.users.models import User, UserSkill
 from care.utils.queryset.user import get_users
 
 
@@ -18,7 +18,7 @@ class UserSkillViewSet(
 
     def get_queryset(self):
         username = self.kwargs["users_username"]
-        user = get_object_or_404(get_users(self.request.user).filter(username=username))
+        user = get_object_or_404(User, username=username)
 
         return self.queryset.filter(user=user)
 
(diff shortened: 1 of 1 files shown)


care/users/api/viewsets/userskill.py
```diff
@@ -18,7 +18,7 @@ class UserSkillViewSet(
 
     def get_queryset(self):
         username = self.kwargs["users_username"]
-        user = get_object_or_404(get_users(self.request.user).filter(username=username))
+        user = get_object_or_404(User, username=username)
```
vigneshhari, line 21: Making this change also allows anyone to update skills for any users, Update with authorization or allow only for list/retrieve calls.

vigneshhari (changes requested):

care/users/api/viewsets/userskill.py
Ashesh3: Fixed now. Added a permission class with all necessary changes. 

- [x] Only district admin and above users should be able to add or remove skills for any users

- [x] All the users should be able to view any other user's linked skills irrespective of permissions

vigneshhari: The user itself should be able to update their skills

Ashesh3: @vigneshhari Done.

care/users/api/viewsets/userskill.py
```diff
@@ -1,25 +1,49 @@
...
+            if requesting_user == user:
+                return True
+
+            if (
+                requesting_user.user_type < User.TYPE_VALUE_MAP["DistrictAdmin"]
+                or requesting_user.user_type in READ_ONLY_USER_TYPES
+            ):
+                return False
+
+            if not requesting_user.user_type >= user.user_type:
```
vigneshhari, line 29: There is no check that confirms if the user is part of the same dist as the dist admin or the same state as the state admin here, but i will approve this PR for now. Do make these changes later or start a new issue for it. 

This should ideally be a helper function that checks if a user has control over another user. this might be useful in many other occasions.

vigneshhari (approved):

PR #1198, opened 2023-02-21
Only allow >=DistrictAdmin to clear home facility

Ashesh3:
According to issue https://github.com/coronasafe/care_fe/issues/4820 , only user with district admin or above access (not read-only users) should be able to clear the home facility from a user. They are the same set of users who can link facilities and set home facilities. 

This PR changes the permissions required to clear the home facility to allow only `>=DistrictAdmin` excluding the read only user types.

@coronasafe/code-reviewers

### DIFF
--- a/care/users/api/viewsets/users.py
+++ b/care/users/api/viewsets/users.py
@@ -17,6 +17,7 @@ from rest_framework.serializers import ValidationError
 from rest_framework.viewsets import GenericViewSet
 
 from care.facility.api.serializers.facility import FacilityBasicInfoSerializer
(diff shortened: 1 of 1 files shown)


gigincg:
@Ashesh3 But a District Admin should not be able to clear access for a State Admin. Is that edge handled here?

Ashesh3:
> @Ashesh3 But a District Admin should not be able to clear access for a State Admin. Is that edge handled here?

Should be fixed. I didn't infer the edge case from the linked issue which stated that *any user* >= District admin should be able to clear home facility of *any* other user.

vigneshhari (approved):

PR #1199, opened 2023-02-21
do not accept middleware address with scheme

sainak:
## Proposed Changes

- prefix middleware address with scheme in openid auth
- update the facility config validator to only accept the domain name without scheme or trailing slash
 
@coronasafe/code-reviewers

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete

### DIFF
--- a/care/facility/api/serializers/facility.py
+++ b/care/facility/api/serializers/facility.py
@@ -13,6 +13,7 @@ from care.users.api.serializers.lsg import (
 )
 from care.utils.csp import config as cs_provider
 from config.serializers import ChoiceField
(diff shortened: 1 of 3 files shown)


khavinshankar (changes requested):
@sainak can you recheck the regex, even the valid URLs are failing

image

sainak:
> @sainak can you recheck the regex, even the valid URLs are failing
> 
> image

That is the intended behavior because we use the same value for websockets in frontend as well, so saving it with scheme breaks the frontend requirement

PR #1223, opened 2023-03-20
feat: Add patient and bed count to Facility serializers

Ashesh3:
Required for: https://github.com/coronasafe/care_fe/issues/4885
Frontend PR: https://github.com/coronasafe/care_fe/pull/5131

This commit adds two new fields, patient_count and bed_count, to the FacilityBasicInfoSerializer and FacilitySerializer classes in care/facility/api/serializers/facility.py. These fields are calculated using the Bed and PatientRegistration models and return the total number of beds and active patients associated with each facility, respectively.

This information is useful for displaying occupancy rates and other related data in the frontend application.
 
@coronasafe/code-reviewers

### DIFF
--- a/care/facility/api/serializers/facility.py
+++ b/care/facility/api/serializers/facility.py
@@ -4,7 +4,9 @@ from django.contrib.auth import get_user_model
 from rest_framework import serializers
 
 from care.facility.models import FACILITY_TYPES, Facility, FacilityLocalGovtBody
+from care.facility.models.bed import Bed
 from care.facility.models.facility import FEATURE_CHOICES
+from care.facility.models.patient import PatientRegistration
 from care.users.api.serializers.lsg import (
     DistrictSerializer,
     LocalBodySerializer,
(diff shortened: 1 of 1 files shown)


Ashesh3:
@vigneshhari any updates on this?

vigneshhari:
This might add a bit more stress on the database, create a ticket to summarise both counts at the facility level with signals. I will merge this in to keep you folks unblocked.

care/facility/api/serializers/facility.py
```diff
@@ -83,6 +97,8 @@ class FacilitySerializer(FacilityBasicInfoSerializer):
     read_cover_image_url = serializers.URLField(read_only=True)
     # location = PointField(required=False)
     features = serializers.MultipleChoiceField(choices=FEATURE_CHOICES)
+    patient_count = serializers.SerializerMethodField()
```
vigneshhari, line 100: You don't need to redefine this since it's already present in the parent class.

vigneshhari (changes requested):

care/facility/api/serializers/facility.py
Ashesh3: Noted. Changes made.

Ashesh3:
> This might add a bit more stress on the database, create a ticket to summarise both counts at the facility level with signals. I will merge this in to keep you folks unblocked.

Made an issue for that here: https://github.com/coronasafe/care/issues/1259

PR #1231, opened 2023-03-27
Add unique constraint for AssetLocation name and facility

Ashesh3:
Required for https://github.com/coronasafe/care_fe/issues/5152

This PR adds a validation check to the `AssetLocation` serializer to ensure that an asset location with the same name and facility does not already exist. The function `validate` in the `AssetLocationSerializer` compares the passed name with the existing entries in the database and raises a `ValidationError` if a match is found.  In addition, the validation now applies to both create and update operations.

The `unique_together` attribute is added to the `AssetLocation` model in the `models.py` file.

The `get_serializer_context` method is also updated to include the facility information when serializing the data.

@coronasafe/code-reviewers

### DIFF
--- a/care/facility/api/serializers/asset.py
+++ b/care/facility/api/serializers/asset.py
@@ -24,6 +24,23 @@ class AssetLocationSerializer(ModelSerializer):
     facility = FacilityBareMinimumSerializer(read_only=True)
     id = UUIDField(source="external_id", read_only=True)
 
+    def validate(self, data):
+        facility = self.context["facility"]
+        if "name" in data:
+            name = data["name"]
+            asset_location_id = self.instance.id if self.instance else None
+            if (
+                AssetLocation.objects.filter(name=name, facility=facility)
+                .exclude(id=asset_location_id)
+                .exists()
+            ):
+                raise ValidationError(
+                    {
+                        "name": "Asset location with this name and facility already exists."
+                    }
+                )
+        return data
+
     class Meta:
         model = AssetLocation
         exclude = (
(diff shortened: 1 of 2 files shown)


care/facility/api/serializers/asset.py
```diff
@@ -21,6 +22,14 @@
 
 
 class AssetLocationSerializer(ModelSerializer):
+    name = serializers.CharField(
+        validators=[
+            UniqueValidator(
+                queryset=AssetLocation.objects.all(),
```
vigneshhari, line 28: This is applied across all asset locations, not just in a given facility.

care/facility/migrations/0337_auto_20230327_1319.py
```diff
@@ -0,0 +1,17 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('facility', '0336_auto_20230222_1602'),
+    ]
+
+    operations = [
+        migrations.AlterUniqueTogether(
+            name='assetlocation',
+            unique_together={('name', 'facility')},
```
vigneshhari, line 15: We cant merge this without checking if there are any current cases like this in production, in which case the migration would fail, @mathew-alex this needs to be checked manually to ensure we don't break anything

vigneshhari (changes requested):
Changes Requested

care/facility/migrations/0337_auto_20230327_1319.py
gigincg: @vigneshhari Wouldn't a Validation do the trick instead of having a Constraint?

vigneshhari: Just validation should do the trick.

Ashesh3: Switched to just validation.

care/facility/api/serializers/asset.py
Ashesh3: Removed.

Ashesh3:
Requested changes are made. Only validation is done to check for duplicate locations within the same facility. No constraints.

vigneshhari:
Changes look good, Just confirm if the filter is case sensitive, We want it case insensitive.

PR #1235, opened 2023-03-30
Add patient ordering by category

Ashesh3:
Required for: https://github.com/coronasafe/care_fe/issues/5197

This PR adds a new filter backend, `PatientCustomOrderingFilter`, to allow custom ordering of the `Patient` queryset based on the `category_severity` field. This field is annotated using a `Case` statement that maps each category value to its corresponding severity order. The filter backend accepts a new query parameter `ordering` to specify the ordering direction, which can be either ascending or descending.

This change also imports the `Case` and `When` classes from `django.db.models` to be used in the PatientCustomOrderingFilter`
implementation.

Overall, this PR enhances the functionality of the `PatientViewSet` by providing a new custom ordering option based on the severity of the patient's category.

@coronasafe/code-reviewers

### DIFF
--- a/care/facility/api/viewsets/patient.py
+++ b/care/facility/api/viewsets/patient.py
@@ -5,6 +5,8 @@ from json import JSONDecodeError
 from django.conf import settings
 from django.contrib.postgres.search import TrigramSimilarity
 from django.core.validators import validate_email
+from django.db import models
+from django.db.models import Case, When
 from django.db.models.query_utils import Q
 from django.utils.timezone import localtime, now
 from django_filters import rest_framework as filters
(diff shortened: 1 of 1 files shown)


siddnikh:
LGTM

care/facility/api/viewsets/patient.py
```diff
@@ -197,6 +200,30 @@ def filter_list_queryset(self, request, queryset, view):
...
+                When(last_consultation__category=cat, then=order)
+                for cat, order in category_ordering.items()
+            ]
+            queryset = queryset.annotate(
+                category_severity=Case(
+                    *when_statements,
+                    default=(len(category_ordering) + 1),
+                    output_field=models.IntegerField(),
+                )
+            ).order_by(ordering)
```
vigneshhari, line 222: Shouldn't you be ordering based on the new annotated field?

vigneshhari (approved):
If this is a requirement then i would recommend creating a migration and moving the values into something that is queryable. Do create an issue for that and track separately.

care/facility/api/viewsets/patient.py
Ashesh3: @vigneshhari That is being already done, the annotated field is `category_severity` and here `ordering` can be either `category_severity` or `-category_severity`.

Ashesh3:
> If this is a requirement then i would recommend creating a migration and moving the values into something that is queryable. Do create an issue for that and track separately.

For the separate issue, you mean a migration to move the values to numeric? So that we can return the category sorted in any order without needing a custom `filter_backend`? In that case I'd imagine the numeric value to be lowest [0] (Comfort) to highest [n] (Critical).

PR #1246, opened 2023-04-12
Support free flow text for assigned_facility in Shifting form

Ashesh3:
Required for: https://github.com/coronasafe/care_fe/pull/5323

## Proposed Changes

- Add `assigned_facility_external` to store the raw name of facility which are not in CARE. This is used to support free flow input in theShifting form for Assigned facility.
- `preferred_vehicle_choice` is made optional, because it is being hid from the frontend.

PR is split into two commits, `changes` and `formatting`

@coronasafe/code-reviewers


care/facility/api/serializers/shifting.py
```diff
@@ -182,16 +244,25 @@ def create(self, validated_data):
...
         ).id
 
-        if "assigned_facility" in validated_data:
-            assigned_facility_external_id = validated_data.pop("assigned_facility")["external_id"]
+        if validated_data.get("assigned_facility_external"):
+            validated_data["assigned_facility"] = None
+        elif validated_data.get("assigned_facility"):
+            validated_data["assigned_facility_external"] = None
+
+        if validated_data.get("assigned_facility"):
```
vigneshhari, line 262: This condition is just above it as well right?

care/facility/api/serializers/shifting.py
```diff
@@ -182,16 +244,25 @@ def create(self, validated_data):
...
         validated_data["shifting_approving_facility_id"] = Facility.objects.get(
             external_id=shifting_approving_facility_external_id
         ).id
 
-        if "assigned_facility" in validated_data:
-            assigned_facility_external_id = validated_data.pop("assigned_facility")["external_id"]
+        if validated_data.get("assigned_facility_external"):
+            validated_data["assigned_facility"] = None
+        elif validated_data.get("assigned_facility"):
+            validated_data["assigned_facility_external"] = None
```
vigneshhari, line 260: If both dont exist then we need a validation error right?

care/facility/api/serializers/shifting.py
```diff
@@ -182,16 +244,25 @@ def create(self, validated_data):
 
         validated_data["is_kasp"] = False
 
-        orgin_facility_external_id = validated_data.pop("orgin_facility")["external_id"]
+        # orgin_facility_external_id = validated_data.pop("orgin_facility")["external_id"]
```
vigneshhari, line 247: Why is this commented out?

care/facility/api/serializers/shifting.py
```diff
@@ -135,25 +175,44 @@ def update(self, instance, validated_data):
...
+        if validated_data.get("assigned_facility"):
+            assigned_facility_external_id = validated_data.pop("assigned_facility")[
+                "external_id"
+            ]
             if assigned_facility_external_id:
                 validated_data["assigned_facility_id"] = Facility.objects.get(
                     external_id=assigned_facility_external_id
                 ).id
                 assigned = True
+        if validated_data.get("assigned_facility_external"):
```
vigneshhari, line 196: Not sure what is happening in this block

vigneshhari (changes requested):

care/facility/api/serializers/shifting.py
Ashesh3: Unused variable - Linter change. orgin_facility_external_id is not used anywhere. (L248 is the only place and it was already commented).

care/facility/api/serializers/shifting.py
Ashesh3: Since we are assigning an external facility, there's is no data about it in the care platform, hence `assigned_facility_id` and `assigned_facility_object` cannot be set in this case. So, this block removes those. The only information we have about this external facility is its name. No `id` or `facility_object` is available to set. So they are set to `None`

care/facility/api/serializers/shifting.py
Ashesh3: The above one is for the `update` method, this one is in the `create` method.

care/facility/api/serializers/shifting.py
Ashesh3: This is already being done if both don't exist.
https://github.com/coronasafe/care/pull/1246/files#diff-e73c078ebb0722d6c6cbbdf3071b60a55aaa9945316336701e6b4e9c934fcbc8R209

```
       if "status" in validated_data:
            if validated_data["status"] in RECIEVING_REQUIRED_STATUS:
                if (
                    not (
                        instance.assigned_facility
                        or instance.assigned_facility_external
                    )
                ) and (not assigned):
                    raise ValidationError(
                        {
                            "status": [
                                "Destination Facility is required for moving to this stage."
                            ]
                        }
                    )
```

care/facility/api/serializers/shifting.py
vigneshhari: Line 262 and 259 are the same condition right ?

Ashesh3: Yep those two are the same.. I had separated those code blocks as they served different purposes. The purpose of the initial if else block was to clear out the "other" facility source, so that both external and care facilities could not be set together.

The second `if` block was to set the value of `assigned_facility_id` (old code, already present).

I can proceed with merging those two conditions. So, it'll look like: (Basically merge those two `if` blocks)

```
        if validated_data.get("assigned_facility_external"):
            validated_data["assigned_facility"] = None
        elif validated_data.get("assigned_facility"):
            validated_data["assigned_facility_external"] = None
            assigned_facility_external_id = validated_data.pop("assigned_facility")[
                "external_id"
            ]
            if assigned_facility_external_id:

                validated_data["assigned_facility_id"] = Facility.objects.get(
                    external_id=assigned_facility_external_id
                ).id
```

Should I go ahead with this change?

gigincg (approved):

PR #1269, opened 2023-04-24
Allow transfer for patients defaults to `true` if `peacetime`, else `false`

rithviknishad:
## Proposed Changes

- Allow transfer is defaulted to true if PEACETIME, defaulted to false if not PEACETIME.

#### When `PEACETIME=True`

image

### Associated Issue
  - fixes coronasafe/care_fe#5381

### Architecture changes
  -
 
@coronasafe/code-reviewers

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete


gigincg (changes requested):
Can we do this in the serializer instead & disable allow_transfer when warmode is enabled

cc: @vigneshhari

Issue #1274, opened 2023-04-26
Add support for bulk creation of beds


siddnikh:
Required for: https://github.com/coronasafe/care_fe/issues/5402

Naming of the beds can be: Name specified by creator - index

PR #1275, opened 2023-04-27
Modified bedCreate API View to support bulk creating beds

siddnikh:
Required for https://github.com/coronasafe/care_fe/pull/5413
Will solve #1274 

We could have used ListSerializer for this particular purpose, but since we are bulk creating with only one difference in each object (their name is going to be <name given> - <index>), we don't need validation for each object.

Changes made:
 - Modified BedSerializer to have a 'number_of_beds' field which takes an integer value. If it is equal to 1, we proceed how we did previously, otherwise we bulk_create Bed objects.


care/facility/api/viewsets/bed.py
```diff
@@ -57,6 +59,23 @@ class BedViewSet(
...
+        if number_of_beds > 1:
+            data = [serializer.validated_data] * number_of_beds
+            for i, d in enumerate(data):
+                d['name'] = d['name'] + f' - {i}'
+            res = Bed.objects.bulk_create(data)
+            return Response(res, status=status.HTTP_201_CREATED)
+        
+        self.perform_create(serializer)
+        headers = self.get_success_headers(serializer.data)
+        return Response(serializer.data, status=status.HTTP_201_CREATED, headers=headers)
```
sainak, line 77: you can do this in the serializer itself, override the create method

siddnikh: @sainak 
The reason I chose to implement this here, was that serializer's create method also performs a number of checks. Those checks could help in debugging if the need be.

care/facility/api/serializers/bed.py
```diff
@@ -31,6 +31,8 @@ class BedSerializer(ModelSerializer):
     location = UUIDField(write_only=True, required=True)
     facility = UUIDField(write_only=True, required=True)
 
+    number_of_beds = IntegerField(required=False, default=1)
```
sainak, line 34: import statement missing

care/facility/tests/test_bed_create.py
```diff
@@ -0,0 +1,23 @@
...
+from care.facility.tests.mixins import TestClassMixin
+
+class SingleBedTest(TestClassMixin, TestCase):
+    def test_create(self):
+        user = self.users[0]
+        sample_data = {
+            "bed_type": "REGULAR",
+            "description": "Testing creation of beds.",
+            "facility": "657c32be-d584-476c-9ce2-0412f0e7692e",
+            "location": "7f0c2bba-face-47f1-a547-87617a8c9b04",
```
sainak, line 13: avoid hardcoding these, refer the already created sel.facility and create a location in the function

care/facility/tests/test_multiple_bed_create.py
```diff
@@ -0,0 +1,23 @@
...
+            "name": "Test Bed",
+            "number_of_beds": 5,
+        }
+        response = self.new_request(
+            ("/api/v1/bed/", sample_data, "json"),
+            {"post": "create"},
+            BedViewSet,
+            user
+        )
+        self.assertIs(response.status_code, status.HTTP_201_CREATED)
```
sainak, line 23: keep bed tests in a single file

sainak (changes requested):

sainak (changes requested):
Tests are failing
Format the code using [precommit](https://care-be-docs.coronasafe.network/local-setup/configuration.html#setting-up-pre-commit)

care/facility/api/viewsets/bed.py
```diff
@@ -57,6 +59,29 @@ class BedViewSet(
     search_fields = ["name"]
     filterset_class = BedFilter
 
+    def create(self, request, *args, **kwargs):
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        validated_data = serializer.validated_data
+        number_of_beds = validated_data.get("number_of_beds")
+        del validated_data["number_of_beds"]
```
sainak, line 67:
```suggestion
        number_of_beds = validated_data.pop("number_of_beds", None)
```

care/facility/api/viewsets/bed.py
```diff
@@ -57,6 +59,29 @@ class BedViewSet(
...
+        # Bulk creating n number of beds
+        if number_of_beds > 1:
+            objs = []
+            for i in range(1, number_of_beds + 1):
```
sainak, line 71: we should limit number of beds created in bulk
@gigincg

sainak (changes requested):

care/facility/api/viewsets/bed.py
gigincg: @sainak I agree. Let's Limit to 100 at a time?

cc: @siddnikh

care/facility/api/viewsets/bed.py
```diff
@@ -65,7 +65,7 @@ def create(self, request, *args, **kwargs):
         validated_data = serializer.validated_data
         number_of_beds = validated_data.pop("number_of_beds", 1)
         # Bulk creating n number of beds
-        if number_of_beds > 1:
+        if number_of_beds > 1 and number_of_beds <= 100:
```
sainak, line 68: @siddnikh it would be nice to raise a validation error for number_of_beds > 100

siddnikh: Yeah, it's in the serializer file.

sainak (approved):

rithviknishad (approved):

care/facility/api/viewsets/bed.py
```diff
@@ -57,6 +59,28 @@ class BedViewSet(
...
+        number_of_beds = validated_data.pop("number_of_beds", 1)
+        # Bulk creating n number of beds
+        if number_of_beds > 1 and number_of_beds <= 100:
+            objs = []
+            for i in range(1, number_of_beds + 1):
+                temp_data = dict(validated_data.copy())
+                temp_data["name"] = temp_data["name"] + f" - {i}"
+                objs.append(Bed(**temp_data))
+
+            res = Bed.objects.bulk_create(objs=objs, batch_size=number_of_beds)
```
vigneshhari, line 75: why do you need batch size when the batch size is always going to be the number of objects to be created?

care/facility/api/viewsets/bed.py
```diff
@@ -57,6 +59,28 @@ class BedViewSet(
...
+            return Response(res, status=status.HTTP_201_CREATED)
```
vigneshhari, line 76: Why would you need the response in this case? The bed names are predictable right?

care/facility/tests/test_bed_create.py
```diff
@@ -0,0 +1,61 @@
...
+            "description": "Testing creation of beds.",
+            "facility": self.facility.external_id,
+            "location": self.asset_location.external_id,
+            "name": "Test Bed",
+            "number_of_beds": 5,
+        }
+        response = self.new_request(
+            ("/api/v1/bed/", sample_data, "json"), {"post": "create"}, BedViewSet, user
+        )
+        self.assertIs(response.status_code, status.HTTP_201_CREATED)
```
vigneshhari, line 61: The test needs to check if the beds are created in the DB, at least check the count

care/facility/api/viewsets/bed.py
```diff
@@ -57,6 +59,28 @@ class BedViewSet(
     search_fields = ["name"]
     filterset_class = BedFilter
 
+    def create(self, request, *args, **kwargs):
```
vigneshhari, line 62: All of our core logic is usually placed in the serializer, can we move the current logic there?

care/facility/tests/test_bed_create.py
```diff
@@ -0,0 +1,61 @@
...
+            "description": "Testing creation of beds.",
+            "facility": self.facility.external_id,
+            "location": self.asset_location.external_id,
+            "name": "Test Bed",
+            "number_of_beds": 1,
+        }
+        response = self.new_request(
+            ("/api/v1/bed/", sample_data, "json"), {"post": "create"}, BedViewSet, user
+        )
+        self.assertIs(response.status_code, status.HTTP_201_CREATED)
```
vigneshhari, line 34: Need to check the existence of the bed.

vigneshhari (changes requested):

care/facility/api/viewsets/bed.py
sainak: serializer.create should return single object so we would need to override and check the count and return appropriate response from viewset.create

care/facility/tests/test_bed_create.py
```diff
@@ -0,0 +1,79 @@
...
+        self.assertEqual(
+            Bed.objects.filter(facility=self.facility).count(),
+            sample_data["number_of_beds"],
+        )
+
+
+class MultipleBedTest(TestBase, TestClassMixin, APITestCase):
```
Ashesh3, line 51: LGTM

Suggestion: Add a test that goes over the maximum limit of 100 beds and make sure appropriate HTTP status and validation error are returned.

Ashesh3 (approved):

care/facility/tests/test_bed_create.py
sainak: yep added

rithviknishad (approved):

vigneshhari (approved):

PR #1277, opened 2023-04-27
feat: Add service request history for asset

Ashesh3:
Required for: https://github.com/coronasafe/care_fe/issues/5363
This pull request adds a new `AssetService` model to the project, which will allow us to store all records of service requests for each asset. This pull request also introduces an `AssetService` serializer in the `care/facility/api/serializers/asset.py` file. The `AssetService` serializer is used to retrieve the last service of an Asset. In the `AssetSerializer` serializer, `last_service` is added as a read-only field, and `last_serviced_on` and `note` are added as write-only fields. The `last_service` field retrieves the last service of the Asset using the `AssetService` serializer. 

This pull request also includes changes to the `create` and `update` methods of the `AssetSerializer` serializer to save the `last_service` data when an Asset is created or updated. The `AssetServiceFilter` filter is also added to the `care/facility/api/viewsets/asset.py` file to filter Assets by external id.

Migration has been created to move existing service request to a new AssetService object linked to that asset.
 
@coronasafe/code-reviewers


care/facility/migrations/0351_assetservice.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def move_last_serviced_on_and_notes(apps, schema_editor):
+    Asset = apps.get_model('facility', 'Asset')
+    AssetService = apps.get_model('facility', 'AssetService')
+
+    for asset in Asset.objects.all():
+        if asset.last_serviced_on or asset.notes:
+            asset_service = AssetService(asset=asset, serviced_on=asset.last_serviced_on, note=asset.notes)
+            asset_service.save()
```
sainak, line 15: use bulk create

care/facility/api/serializers/asset.py
```diff
@@ -119,10 +140,39 @@ def validate(self, attrs):
...
+        last_serviced_on = validated_data.pop("last_serviced_on", None)
+        note = validated_data.pop("note", None)
+        asset_instance = super().create(validated_data)
+        if last_serviced_on or note:
+            asset_service = AssetService(
+                asset=asset_instance, serviced_on=last_serviced_on, note=note
+            )
+            asset_service.save()
+            asset_instance.last_service = asset_service
+            asset_instance.save()
```
sainak, line 153: wrap this in a transaction

sainak (changes requested):
run this branch through precommit

sainak (approved):
LGTM, squash the migrations after abdm is merged

Ashesh3:
Lint won't run on the auto generated squashed migration file.

care/facility/api/viewsets/asset.py
```diff
@@ -308,3 +310,24 @@ def get_queryset(self):
...
         return queryset
+
+
+class AssetServiceFilter(filters.FilterSet):
+    qr_code_id = filters.CharFilter(field_name="asset__qr_code_id")
+    external_id = filters.CharFilter(field_name="asset__external_id")
+
+
+class AssetServiceViewSet(ListModelMixin, RetrieveModelMixin, GenericViewSet):
+    queryset = (
```
vigneshhari, line 321: Are we okay showing service history without any authZ

vigneshhari:
Squash all migrations to a single one.

vigneshhari:
@rithviknishad add your review.

care/facility/api/viewsets/asset.py
Ashesh3: @vigneshhari If I remember correctly, the asset details are publicly available and are shown on the state dashboard, in that case it should be fine to make their service history public too, right? 

cc: @gigincg

Ashesh3: @vigneshhari Aparna confirmed to have it publicly visible.

care/facility/models/asset.py
```diff
@@ -177,3 +182,10 @@ class AssetTransaction(BaseModel):
     performed_by = models.ForeignKey(
         User, on_delete=models.PROTECT, null=False, blank=False
     )
+
+
+class AssetService(BaseModel):
+    asset = models.ForeignKey(Asset, on_delete=models.PROTECT, null=False, blank=False)
+
+    serviced_on = models.DateField(default=None, null=True, blank=True)
```
rithviknishad, line 190: Do we need to allow asset service records without a serviced_on datetime for the record?

rithviknishad (commented):
rebase and squash the migrations btw

care/facility/models/asset.py
Ashesh3: Currently the system allows users to enter a blank date with a note or vice versa, so I carried over that behavior. It doesn't make sense to have a service record without a date, but if it's being used already then, we have to keep the behavior, otherwise we can make the date mandatory.

rithviknishad: so the timeline table would be sorted by? modified/created date?

rithviknishad: @gigincg is it fine to allow serviced_on to be blank?

we could just allow null for migration purposes but have `blank=False` to make it a required field in forms.

Ashesh3: okay, let's do that!

Ashesh3, line 190:
```suggestion
    serviced_on = models.DateField(default=None, null=True, blank=False)
```

care/facility/api/serializers/asset.py
```diff
@@ -125,10 +146,40 @@ def validate(self, attrs):
...
+        note = validated_data.pop("note", None)
+        with transaction.atomic():
+            asset_instance = super().create(validated_data)
+            if last_serviced_on or note:
+                asset_service = AssetService(
+                    asset=asset_instance, serviced_on=last_serviced_on, note=note
+                )
+                asset_service.save()
+                asset_instance.last_service = asset_service
+                asset_instance.save()
```
rithviknishad, line 160:
```suggestion
                asset_instance.save(update_fields=["last_service"])
```

rithviknishad (changes requested):
Tests are failing; you'll need to makemigrations again 😅

rithviknishad (approved):

sainak (approved):

care/facility/api/serializers/asset.py
```diff
@@ -61,12 +64,67 @@ class Meta:
...
+    edited_by = UserBaseMinimumSerializer(read_only=True)
+
+    class Meta:
+        model = AssetServiceEdit
+        exclude = ("asset_service",)
+
+
+class AssetServiceSerializer(ModelSerializer):
+    id = UUIDField(source="external_id", read_only=True)
+    asset = AssetBareMinimumSerializer(read_only=True)
```
sainak, line 86:
```suggestion
```
this is not required anymore

care/facility/api/serializers/asset.py
```diff
@@ -61,12 +64,67 @@ class Meta:
...
+        if serviced_on == instance.serviced_on and note == instance.note:
+            return instance
+
+        with transaction.atomic():
+            edit = AssetServiceEdit(
+                asset_service=instance,
+                edited_on=now(),
+                edited_by=user,
+                serviced_on=validated_data.get("serviced_on", instance.serviced_on),
+                note=validated_data.get("note", instance.note),
```
sainak, line 110:
```suggestion
                serviced_on=serviced_on,
                note=note,
```

sainak (changes requested):

sainak (approved):

care/facility/models/asset.py
```diff
@@ -177,3 +182,32 @@ class AssetTransaction(BaseModel):
...
     )
+
+
+class AssetService(BaseModel):
+    asset = models.ForeignKey(Asset, on_delete=models.PROTECT, null=False, blank=False)
+
+    serviced_on = models.DateField(default=None, null=True, blank=False)
+    note = models.TextField(default="", null=True, blank=True)
+
+    edits = models.ManyToManyField("AssetServiceEdit", related_name="edits")
```
rithviknishad, line 193: Why is this many to many?

rithviknishad (approved):

care/facility/models/asset.py
sainak, line 193:
```suggestion
```

care/facility/models/asset.py
```diff
@@ -177,3 +182,32 @@ class AssetTransaction(BaseModel):
...
+
+    @property
+    def edit_history(self):
+        return self.edits.order_by("-edited_on")
+
+
+class AssetServiceEdit(models.Model):
+    asset_service = models.ForeignKey(
+        AssetService, on_delete=models.CASCADE, null=False, blank=False
```
sainak, line 202:
```suggestion
        AssetService, on_delete=models.CASCADE, null=False, blank=False, related_name="edits"
```

care/facility/api/serializers/asset.py
```diff
@@ -61,12 +64,66 @@ class Meta:
...
+        exclude = ("asset_service",)
+
+
+class AssetServiceSerializer(ModelSerializer):
+    id = UUIDField(source="external_id", read_only=True)
+    edits = serializers.SerializerMethodField()
+
+    def get_edits(self, obj):
+        edits = AssetServiceEdit.objects.filter(asset_service=obj)
+        return AssetServiceEditSerializer(edits, many=True).data
```
sainak, line 90:
```suggestion
    edits = AssetServiceEditSerializer(many=True)
```

care/facility/api/viewsets/asset.py
```diff
@@ -338,3 +340,53 @@ def get_queryset(self):
...
+    RetrieveModelMixin,
+    UpdateModelMixin,
+    GenericViewSet,
+):
+    queryset = (
+        AssetService.objects.all()
+        .select_related(
+            "asset",
+        )
+        .order_by("-created_date")
```
sainak, line 361:
```suggestion
        )
        .prefetch_related("edits")
        .order_by("-created_date")
```

sainak (changes requested):

care/facility/models/asset.py
Ashesh3: image

Resolved.

sainak:
@Ashesh3 migrations needs to updated

sainak (approved):

rithviknishad (approved):

sainak (approved):

rithviknishad (approved):

vigneshhari (approved):

PR #1291, opened 2023-05-08
Lambda deployment for pull requests

mathew-alex:


vigneshhari:
@mathew-alex we can move the migrations out of the docker image and plan to release this.

PR #1310, opened 2023-05-23
optimized the query performed to fetch skills in UserAssignedSerializer

### LINKED ISSUE #1189: Optimize the query to fetch the skills associated to a User in `UserAssignedSerializer`
Optimize the query to fetch the skills associated with a User in `UserAssignedSerializer`

- As a follow-up of #1188 
- Ref. comment: https://github.com/coronasafe/care/pull/1188#issuecomment-1427023488

yaswanthsaivendra:
## Proposed Changes

- deleted add_skills method in UserAssignedSerializer
-  overided to_representation method .
- Made use of prefetch_related method to fetch skills from a many to many relationship to skills. So that skills are prefetched using a single query . This way we can access the prefetched skills from memory rather than making multiple db requests.
- prefetch_related (internally makes use of JOIN) to fetch all the related objects in a single query.

### Associated Issue
  - Fixes #1189 

### Architecture changes
  -
 
@coronasafe/code-reviewers

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete

### DIFF
--- a/config/auth_views.py
+++ b/config/auth_views.py
@@ -101,7 +101,9 @@ class TokenRefreshSerializer(serializers.Serializer):
             data["refresh"] = str(refresh)
 
         # Updating users active status
-        User.objects.filter(id=refresh["user_id"]).update(last_login=localtime(now()))
+        User.objects.filter(external_id=refresh["user_id"]).update(
+            last_login=localtime(now())
+        )
 
         return data
 


care/users/api/serializers/user.py
```diff
@@ -382,6 +377,14 @@ class Meta:
...
         )
 
+    def to_representation(self, instance):
+        representation = super().to_representation(instance)
+        instance = User.objects.prefetch_related("skills").get(pk=instance.pk)
+        skills = instance.skills.filter(userskill__deleted=False)
+        skills_data = SkillSerializer(skills, many=True).data
+        representation["skills"] = skills_data
+        return representation
+
```
sainak, line 387: This will still make N+1 queries for a list of users, the most efficient way would be to prefetch these objects where the queryset is being called.

yaswanthsaivendra: Ohh yes , I got it know. I will make the changes

yaswanthsaivendra:
@sainak 
UserAssingedSerializer is used at 2 places.
1. in FacilityUserViewSet, where I have made the changes to prefetch the skills.
2. in [PatientConsulationSerializer](https://github.com/coronasafe/care/blob/7d14e41251787fe1bbb952ac22278f998b86b24a/care/facility/api/serializers/patient_consultation.py#L68)  here i am not sure how it can be done.
Can u suggest , what can be done here?

sainak:
@yaswanthsaivendra in patient consultation view you can prefetch `assigned_to__skills`

care/facility/api/viewsets/facility_users.py
```diff
@@ -21,13 +25,17 @@ class Meta:
...
+    queryset = User.objects.all().prefetch_related(
+        Prefetch("skills", queryset=Skill.objects.filter(userskill__deleted=False))
+    )
     permission_classes = [IsAuthenticated]
     filter_backends = [filters.DjangoFilterBackend]
 
     def get_queryset(self):
         try:
-            facility = Facility.objects.get(external_id=self.kwargs.get("facility_external_id"))
+            facility = Facility.objects.get(
```
sainak, line 36: This will override the queryset property, so the optimization that you make won't work
I would recommend using reverse fk lookup to filter users of a facility

yaswanthsaivendra: oops 😬 Got it!

yaswanthsaivendra:
Hey @sainak ,
I made the necessary changes , can u review it please!

sainak:
@yaswanthsaivendra there are no changes in this pr
you should create a new branch for every pull request

yaswanthsaivendra:
> @yaswanthsaivendra there are no changes in this pr you should create a new branch for every pull request

sorry, I messed a bit. Pushed the changes Now.

The worst thing, I did at the start is making changes in my master branch which leads to all this mess in the commits. Will avoid it in my further prs.

care/users/api/serializers/user.py
```diff
@@ -358,11 +359,20 @@ class UserAssignedSerializer(serializers.ModelSerializer):
...
+    #     qs = obj.skills.filter(userskill__deleted=False)
+    #     return SkillSerializer(qs, many=True).data
+
+    @staticmethod
+    def skills_eager_loading(queryset):
+        """Perform necessary eager loading of data."""
+        queryset = queryset.prefetch_related(
+            Prefetch("skills", queryset=Skill.objects.filter(userskill__deleted=False))
+        )
+        return queryset
```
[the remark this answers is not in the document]
yaswanthsaivendra: Yup did that!!

sainak (approved):

PR #1318, opened 2023-05-27
Added a way to handle filtering admitted but no bed assigned 

### LINKED ISSUE #1184: Patient filter: Add `None` option for Last admitted to Bed filter to filter patients that are admitted but has no bed assigned
In Advanced filter on the patient list, https://care.coronasafe.in/patients
The last admitted to bed type filter is not in sync with the updated bed management.

The problem is, how do we search for a patient that is admitted, but for which no bed is allocated yet.

yaswanthsaivendra:
## Proposed Changes

- updated the PatientFilterSet to handle the None option
- made the necessary change in frontend and created an [PR](https://github.com/coronasafe/care_fe/pull/5550) for the same.

### Associated Issue
  - Fixes #1184 

### Architecture changes
  -
 
@coronasafe/code-reviewers

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete


care/facility/api/viewsets/patient.py
```diff
@@ -147,8 +147,25 @@ def filter_by_category(self, queryset, name, value):
...
+        field_name="last_consultation__current_bed__bed__bed_type",
+        method="filter_by_bed_type",
     )
+
+    def filter_by_bed_type(self, queryset, name, value):
+        values = value.split(",")
+        union_queryset = PatientRegistration.objects.none()
+        for val in values:
+            val.strip()
+            if val == "8":
```
rithviknishad, line 159: Should we use `8` for filtering for no bed assigned?

What happens when we have to add a new bed type, but that new bed type's value would be 8 right? And this would cause conflict here, right?

```suggestion
            if val == "None":
```

rithviknishad (changes requested):

rithviknishad (approved):

care/facility/api/viewsets/patient.py
yaswanthsaivendra: Got it!

care/facility/api/viewsets/patient.py
```diff
@@ -147,8 +147,25 @@ def filter_by_category(self, queryset, name, value):
...
+            val = val.strip()
+            if val == "None":
+                union_queryset |= queryset.filter(
+                    last_consultation__current_bed__isnull=True
+                )
+            else:
+                union_queryset |= queryset.filter(
+                    last_consultation__current_bed__bed__bed_type=val
+                )
+        return union_queryset
```
vigneshhari, line 167: Use a Q function to write this filter, using an | with the base queryset can make the query really complicated.

vigneshhari (changes requested):

gigincg:
cc: @yaswanthsaivendra

care/facility/api/viewsets/patient.py
yaswanthsaivendra: okay!

yaswanthsaivendra: @vigneshhari , made changes please review the update code!

care/facility/api/viewsets/patient.py
```diff
@@ -147,8 +147,23 @@ def filter_by_category(self, queryset, name, value):
...
+        filter_q = Q()
+
+        for val in values:
+            val = val.strip()
+            if val == "None":
+                filter_q |= Q(last_consultation__current_bed__isnull=True)
+            else:
+                filter_q |= Q(last_consultation__current_bed__bed__bed_type=val)
+
+        return queryset.filter(filter_q)
```
sainak, line 165: you can just query `last_consultation__current_bed__bed__bed_type__in=values` instead of a for loop

sainak (changes requested):

care/facility/api/viewsets/patient.py
yaswanthsaivendra: But we don't have the "None" as value in the choices of `last_consultation__current_bed__bed_type` rgt?? So how does the above query work??

sainak: https://docs.djangoproject.com/en/4.2/ref/models/querysets/#in
and IN operator is a shorthand for multiple OR operators in SQl so None would be ignored

But you won't get `None` in a list, it'll be just `None` or list of bed types

yaswanthsaivendra: Ohh okay.
I considered the case where "None"  would be coming in a list.(as it's simply one of the IDs of the const object of bed types they have in frontend)
Coz the frontend functionality is a multi select filter. So there might be a case where they will querying for "all the patients who are having an ICU bed alloted or no bed alloted".
If you're sure that we won't be getting the "None" option in a list of options from the frontend then do lemme know. I will make the changes accordingly.

sainak: No you won't get None within the csv list, but instead the `value` can be an **empty string** or **None Type** in python

care/facility/api/viewsets/patient.py
```diff
@@ -147,8 +147,23 @@ def filter_by_category(self, queryset, name, value):
...
+    def filter_by_bed_type(self, queryset, name, value):
+        values = value.split(",")
+        filter_q = Q()
+
+        for val in values:
+            val = val.strip()
+            if val == "None":
+                filter_q |= Q(last_consultation__current_bed__isnull=True)
+            else:
+                filter_q |= Q(last_consultation__current_bed__bed__bed_type=val)
```
sainak, line 163:
```suggestion
        try:
            values.remove("None")
            filter_q |= Q(last_consultation__current_bed__isnull=True)
        except KeyError:
            pass
        filter_q |= Q(ast_consultation__current_bed__bed__bed_type__in=values)
```

sainak (changes requested):

sainak (approved):

yaswanthsaivendra:
@vigneshhari @sainak @rithviknishad Added tests , Please Review!!

rithviknishad (approved):

sainak (approved):

nihal467:
LGTM, @vigneshhari can you review it, don't merge it this week

nihal467:
@vigneshhari can you merge it, we can planning to close it this week

vigneshhari (approved):

PR #1328, opened 2023-05-30
Delete notifications older than 90 days

### LINKED ISSUE #1320: Periodically hard delete Notification records that are older than 90 days.
**Is your feature request related to a problem? Please describe.**
The notification table is huge. We don't want notification records older than 90 days.

**Describe the solution you'd like**
Periodically hard delete records from Notifications table that are older than 90 days.

yaswanthsaivendra:
## Proposed Changes

- created a periodic task to delete notifications that are older than 90 days.

### Associated Issue
  - Fixes #1320 

### Architecture changes
  -
 
@coronasafe/code-reviewers

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete


sainak (approved):

vigneshhari:
Can you create a test for this as well? Once you have the tests added we can merge.

vigneshhari (approved):

vigneshhari:
You could have gotten away without mocking in these tests by using the update method of the django queryset ( they do not invoke the save methods and the created timestamps can be added manually "to my knowledge" ), even in this case, we are not mocking the database interactions, we are just mocking the time method. 
In almost all cases we would not want to mock the database, the database is what we are usually testing against. mocking is useful when there is an external service involved, like an SMS API or an integration with an API, in our unit tests we mock all the responses they can give ( success, failure, exception ) and test if the function we wrote handles every case correctly.

yaswanthsaivendra:
> You could have gotten away without mocking in these tests by using the update method of the django queryset ( they do not invoke the save methods and the created timestamps can be added manually "to my knowledge" ), even in this case, we are not mocking the database interactions, we are just mocking the time method. In almost all cases we would not want to mock the database, the database is what we are usually testing against. mocking is useful when there is an external service involved, like an SMS API or an integration with an API, in our unit tests we mock all the responses they can give ( success, failure, exception ) and test if the function we wrote handles every case correctly.

Yaa ik that i haven't mocked any database interactions in the pushed code . I just asked it coz I first wrote the test by mocking database interactions, later i realised that isn't a gud idea. So i changed it. Thanks for sharing ur opinion. :smile:

PR #1329, opened 2023-06-01
Delete incomplete file uploads

### LINKED ISSUE #1326: Peroidic tasks to remove incomplete file uploads
**Is your feature request related to a problem? Please describe.**
Currently, I believe we don't have a logic for deleting the incomplete s3 file uploads.

**Describe the solution you'd like**
Create a `periodic_task` which handles the deletion of incomplete s3 file  uploads. 

soln I'm thinking of - the function would be deleting the files which are created till yesterday and are having `upload_completed=False`

yaswanthsaivendra:
## Proposed Changes

- delete incomplete file uploads on s3 as well as their entries on db.

### Associated Issue
  - Fixes #1326 

@coronasafe/code-reviewers

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete


care/facility/tasks/fileupload/delete_incomplete_file_uploads.py
```diff
@@ -0,0 +1,20 @@
...
+    run_every=crontab(minute="0", hour="0")
+)  # Run the task daily at midnight
+def delete_incomplete_file_uploads():
+    yesterday = timezone.now() - timedelta(days=1)
+    incomplete_uploads = FileUpload.objects.filter(
+        created__date__lte=yesterday, upload_completed=False
+    )
+    for upload in incomplete_uploads:
+        upload.delete_object()
+        upload.delete()
```
rithviknishad, line 20: Why a loop to delete? We can just do `incomplete_uploads.update(deleted=True)` right?

rithviknishad (approved):

sainak (approved):

yaswanthsaivendra:
Updated the code - to make bulk delete from s3, thereby reducing the number of api requests.

care/facility/tasks/fileupload/delete_incomplete_file_uploads.py
```diff
@@ -0,0 +1,25 @@
...
+        created__date__lte=yesterday, upload_completed=False
+    )
+
+    s3_keys = [
+        f"{upload.FileType(upload.file_type).name/upload.internal_name}"
+        for upload in incomplete_uploads
+    ]
+
+    FileUpload.bulk_delete_objects(s3_keys)
+    incomplete_uploads.delete()
```
rithviknishad, line 25: This is a hard delete of the incomplete `FileUpload` records.

@sainak @vigneshhari is this okay?

vigneshhari: In this case i think it should be fine.

vigneshhari:
@khavinshankar @sainak can you guys check if this affects discharge summaries or ABDM uploads.

vigneshhari:
@sainak I know that it affects all uploaded files, i want to know if it affects any of our existing workflows, since we use files in HCX and discharge summaries and other stuff like that.

khavinshankar:
@vigneshhari this shouldn't be a problem  for HCX or discharge summary.

care/facility/tasks/fileupload/delete_incomplete_file_uploads.py
```diff
@@ -0,0 +1,25 @@
...
+from care.facility.models.file_upload import FileUpload
+
+
+@periodic_task(
+    run_every=crontab(minute="0", hour="0")
+)  # Run the task daily at midnight
+def delete_incomplete_file_uploads():
+    yesterday = timezone.now() - timedelta(days=1)
+    incomplete_uploads = FileUpload.objects.filter(
+        created__date__lte=yesterday, upload_completed=False
```
vigneshhari, line 16: `created__date` should be `created_date` right ?

vigneshhari (changes requested):

vigneshhari:
Please add unit tests, unit tests are required for every PR's.

vigneshhari:
> Updated the code - to make bulk delete from s3, thereby reducing the number of api requests.

Can you check if the bulk API has an upper limit, This might create an issue when the first time it's run.

yaswanthsaivendra:
> > Updated the code - to make bulk delete from s3, thereby reducing the number of api requests.
> 
> Can you check if the bulk API has an upper limit, This might create an issue when the first time it's run.

Yes, the upper limit is 1000 keys per a request. I updated code to handle the keys in batches. Btw just to confirm , our bucket is not object versioned rgt?

yaswanthsaivendra:
> Please add unit tests, unit tests are required for every PR's.

Actually the reason I haven't added the tests is that I haven't tests for most of the existing code expect for some APIs in `/care/facility/tests/` and for 1 or 2 models in `care/facility/models/tests/` . Am I missing something?
Can u suggest me a good place where can I include tests for the celery_task and bulk_delete_objects ?

vigneshhari:
> > Please add unit tests, unit tests are required for every PR's.
> 
> Actually the reason I haven't added the tests is that I haven't tests for most of the existing code expect for some APIs in `/care/facility/tests/` and for 1 or 2 models in `care/facility/models/tests/` . Am I missing something? Can u suggest me a good place where can I include tests for the celery_task and bulk_delete_objects ?

We don't have unit tests for most ( >90% ) of the code base, but recently we started enforcing unit testing, you can create unit tests like existing ones and call the celery methods directly to test them. You have the option to test the S3 operations via localstack, if you don't want to go that route, you can mock the S3 operations.

yaswanthsaivendra:
> > > Please add unit tests, unit tests are required for every PR's.
> > 
> > 
> > Actually the reason I haven't added the tests is that I haven't tests for most of the existing code expect for some APIs in `/care/facility/tests/` and for 1 or 2 models in `care/facility/models/tests/` . Am I missing something? Can u suggest me a good place where can I include tests for the celery_task and bulk_delete_objects ?
> 
> We don't have unit tests for most ( >90% ) of the code base, but recently we started enforcing unit testing, you can create unit tests like existing ones and call the celery methods directly to test them. You have the option to test the S3 operations via localstack, if you don't want to go that route, you can mock the S3 operations.

Okay

yaswanthsaivendra:
> @mathew-alex are S3 objects versioned in any deployment ?

Hey I have gone through the code base quite a few times, wherever s3 configs or s3 methods made use of. No where Versioning details are mentioned. if versioning exists, it has to dealt while creation and retrieving of objects. So, ig none of our existing deployments are using versioned s3 buckets.

mathew-alex:
> > @mathew-alex are S3 objects versioned in any deployment ?
> 
> Hey I have gone through the code base quite a few times, wherever s3 configs or s3 methods made use of. No where Versioning details are mentioned. if versioning exists, it has to dealt while creation and retrieving of objects. So, ig none of our existing deployments are using versioned s3 buckets.

@yaswanthsaivendra - bucket versioning is enabled in cloud bucket settings.

vigneshhari:
@yaswanthsaivendra please reply to @mathew-alex 's comment

Issue #1405, opened 2023-06-24
Add Validation for landline numbers in Phone Number field in Patient Registration


siddnikh:
The Phone Number validator rejects landline numbers. Even though it accepts the "0" part of the number, it expects one of 4,5,6,7,8 or 9 to be the next digit. STD codes don't always start with these digits, hence we need to change the regex for this.

Follow up for: https://github.com/coronasafe/care_fe/issues/5776

PR #1406, opened 2023-06-24
Changed Regex Validator for Phone Number Field for Patient Registration Form

siddnikh:
Solves #1405 
The landline numbers weren't getting accepted because the regex expected one of 4,5,6,7,8 or 9 after 0 in the starting of the phone number.
This won't work because most of the std codes start with 1 or 2 or 3.


sainak (approved):

vigneshhari:
@sainak I am merging this MR, could you create an issue to add some tests for this change?  ( or you can do it yourself! )

PR #1440, opened 2023-07-06
Add discharge_reason filter for patient

Ashesh3:
## Proposed Changes

- Allow patient list to be filtered with `last_consultation__discharge_reason`

### Associated Issue

- https://github.com/coronasafe/care_fe/issues/5474
- https://github.com/coronasafe/care_fe/pull/5843


sainak (approved):

sainak:
> This filter would work only when `is_active=False` right?

we have a `is_active` filter for it

Ashesh3:
> This filter would work only when `is_active=False` right?

Yep, active=False is being automatically applied on the frontend and sent along with this filter when this filter is selected.

vigneshhari (approved):

PR #1443, opened 2023-07-07
Medibase: Drop littletable; instead store as list from ORM's `values_list`

rithviknishad:
## Proposed Changes

- MedibaseMedicineTable no longer uses `littletable` and is now simply a `list` fetched ORM's `values_list`

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete

### DIFF
--- a/care/facility/api/viewsets/prescription.py
+++ b/care/facility/api/viewsets/prescription.py
@@ -140,7 +140,7 @@ class MedibaseViewSet(ViewSet):
     def serailize_data(self, objects):
         return [
             {
-                "external_id": x[0],
+                "id": x[0],
                 "name": x[1],
                 "type": x[2],
                 "generic": x[3],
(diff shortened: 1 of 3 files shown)


nihal467:
I tested in  tunnel provided by @rithviknishad's and added multiple medicines to a patients looks good to me in a general view

rithviknishad:
Worker timeout occurs as loading to in-memory on wsgi.py takes too much time for (0.25cpu)
No memory issues though.

rithviknishad:
But this happens for me even with lazy load (not on wsgi.py, load on first search request) on my CPU.
Probably because my `0.25` cpu is not same as staging instances `0.25` cpu

care/facility/api/viewsets/prescription.py
```diff
@@ -180,8 +172,6 @@ def list(self, request):
 
         if request.GET.get("query", False):
             query = request.GET.get("query").strip().lower()
```
sainak, line 174:
```suggestion
        if query := request.query_params.get("query"):
            query = query.strip().lower()
```

sainak (approved):

vigneshhari (approved):

PR #1450, opened 2023-07-13
add table to track user facility allocation

sainak:
## Proposed Changes

- create a new table to track user's start date and end date at home_facility
- create signal to update the table on changes to the user model
- create migration to back-fill previous record (we don't have the exact allocation date at current home_facility of users so using the users created date)

### Associated Issue

- Closes coronasafe/care_fe#5862

## Merge Checklist

- [x] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete


care/users/models.py
```diff
@@ -363,3 +364,16 @@ def save(self, *args, **kwargs) -> None:
...
+
+class UserFacilityAllocation(models.Model):
+    """
+    This model tracks the allocation of a user to a facility for metabase analytics.
+    """
+
+    user = models.ForeignKey(User, on_delete=models.CASCADE, related_name="+")
+    facility = models.ForeignKey(
+        "facility.Facility", on_delete=models.CASCADE, related_name="+"
+    )
```
rithviknishad, line 377:
```suggestion
    user = models.ForeignKey(User, on_delete=models.CASCADE)
    facility = models.ForeignKey(
        "facility.Facility", on_delete=models.CASCADE)
```

@gigincg should we allow backward relation so that we could do things like show a user allocation timeline graph in the care fe or list users in a facility for a specific time range?

rithviknishad (approved):

care/users/models.py
vigneshhari: @gigincg bump!

sainak: This is not planned currently

vigneshhari (approved):

PR #1454, opened 2023-07-14
Asset Tracking and Monitoring Enhancements

Ashesh3:
This PR introduces several changes aimed at improving asset tracking and monitoring.

- AssetAvailabilityRecord Model: Introduced a new model AssetAvailabilityRecord in models/asset.py and the corresponding migration file. This model will keep track of the availability status of each asset, which includes not monitored, operational, down, and under maintenance states.

- Asset Monitoring Task: Added a new Celery task in tasks/asset_monitor.py for checking the status of each asset periodically. The task fetches the status from the middleware and updates the AssetAvailabilityRecord accordingly.

- Asset Status Enum: Created a new Enum class AvailabilityStatus in utils/assetintegration/asset_statuses.py to represent the availability status of assets.

- Task Scheduling: Added the new asset monitoring task to the period tasks list in tasks/init.py. This task will be executed every 30 minutes.

- Asset Model Update: Updated the Asset model in models/asset.py to include a reference to the new AssetAvailabilityRecord model.

## Celery Task Pseudocode:
The Celery task `check_asset_status` is set to run every 30 minutes and is responsible for checking and updating the status of all assets.

1. **Fetch All Assets:** The task begins by fetching all assets from the database.

2. **Iterate Over Assets:** For each asset, it checks if the asset class is defined and if the local IP address is available in the asset's metadata. If either of these conditions is not met, it continues to the next asset.

3. **Get Asset Class:** The task then determines the asset class using the `AssetClasses` enumeration and the asset's metadata, which includes the middleware hostname.

4. **Check Cached Results:** Before making a request to the middleware API, the task checks if the status of assets for the given hostname is already cached. If it is, the task uses the cached results instead of making a new API call.

5. **API Call:** If the results are not cached, the task makes a GET request to the middleware API to fetch the status of all devices.

6. **Determine New Status:** The task then determines the new status of the asset based on the response from the API. If the asset's local IP address is not in the status records returned by the API, the asset's status is set to "not monitored".

7. **Get Last Record:** The task fetches the most recent status record for the asset.

8. **Create New Record:** If there is no previous record or if the asset's status has changed, the task creates a new `AssetAvailabilityRecord` with the new status.

9. **Error Handling:** If there is any exception during the process, it is caught and printed to the console. The task then proceeds to the next asset.

The purpose of this task is to ensure that the asset status is updated regularly, allowing for efficient management and monitoring of assets.


sainak (changes requested):

sainak (approved):

sainak:
@Ashesh3 add tests for api endpoints

care/facility/api/viewsets/asset.py
```diff
@@ -128,6 +130,17 @@ def retrieve(self, request, *args, **kwargs):
         return Response(hit)
 
 
+class AssetAvailabilityFilter(filters.FilterSet):
+    external_id = filters.CharFilter(field_name="asset__external_id")
+
+
+class AssetAvailabilityViewSet(ListModelMixin, RetrieveModelMixin, GenericViewSet):
+    queryset = AssetAvailabilityRecord.objects.all().select_related("asset")
```
vigneshhari, line 138: Are we sure we want to open this endpoint to everyone? cc @gigincg

care/facility/models/asset.py
```diff
@@ -17,6 +17,13 @@ def get_random_asset_id():
     return str(uuid.uuid4())
 
 
+class AvailabilityStatus(models.IntegerChoices):
```
vigneshhari, line 20: Since this is going to be used in metabase it might make sense to switch to string based values to make it easier for the metabase folks

care/facility/tasks/asset_monitor.py
```diff
@@ -0,0 +1,91 @@
...
+                    ].value(
+                        {
+                            **asset.meta,
+                            "middleware_hostname": hostname,
+                        }
+                    )
+                    result = asset_class.api_get(asset_class.get_url("devices/status"))
+                    middleware_status_cache[hostname] = result
+                except Exception as e:
+                    print("Error in Asset Status Check", e)
```
vigneshhari, line 46: Dont Print, use logging

care/facility/api/viewsets/asset.py
Ashesh3: @vigneshhari Gigin said it's okay for this endpoint to be public for now.

Ashesh3:
> @Ashesh3 add tests for api endpoints

Tests added.

sainak (approved):

vigneshhari (approved):

PR #1466, opened 2023-07-19
Mark asset as down when middleware is down

Ashesh3:
## Proposed Changes

- When the middleware server is down, all assets addressing that middleware are marked as down.
- When the middleware does not report status of a particular asset, that asset is marked down.
- If the middleware is down, we show a warning instead of an exception.

### DIFF
--- a/care/facility/tasks/asset_monitor.py
+++ b/care/facility/tasks/asset_monitor.py
@@ -55,7 +55,7 @@ def check_asset_status():
                     logger.warn(f"Middleware {hostname} is down", exc_info=True)
 
             # If no status is returned, setting default status as down
-            if not result or len(result) == 0:
+            if not result:
                 result = [{"time": timezone.now().isoformat(), "status": []}]
 
             middleware_status_cache[hostname] = result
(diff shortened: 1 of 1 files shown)


care/facility/tasks/asset_monitor.py
```diff
@@ -45,42 +49,43 @@ def check_asset_status():
...
                 except Exception:
-                    logger.exception("Error in Asset Status Check - Fetching Status")
-                    middleware_status_cache[hostname] = None
-                    continue
+                    logger.warn(f"Middleware {hostname} is down", exc_info=True)
 
-            if not result:
-                continue
+            # If no status is returned, setting default status as down
+            if not result or len(result) == 0:
```
sainak, line 58:
```suggestion
            if not result:
```
`not {}` returns true so no need to do length check

sainak (approved):

care/facility/tasks/asset_monitor.py
sainak: yes `not []` also returns true

Ashesh3: Oh yea, How did I miss that. 🤔

sainak (approved):

rithviknishad (approved):

PR #1475, opened 2023-07-25
Ignore admission date during discharge process if not provided

Ashesh3:
## Proposed Changes

This pull request introduces a change that allows the discharge process to ignore the admission date if it's not provided. This avoids potential validation errors and exceptions during the discharge process when the admission date is absent.

- Modified the validation logic in patient_consultation.py to check if the admission date exists before comparing it with the death date.

Fixes https://github.com/coronasafe/care_fe/issues/5916

### DIFF
--- a/care/facility/api/serializers/patient_consultation.py
+++ b/care/facility/api/serializers/patient_consultation.py
@@ -437,7 +437,10 @@ class PatientConsultationDischargeSerializer(serializers.ModelSerializer):
                 raise ValidationError(
                     {"death_datetime": "This field value cannot be in the future."}
                 )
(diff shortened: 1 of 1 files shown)


rithviknishad (approved):

sainak (approved):

sainak:
Also fixes https://coronasafe-network.sentry.io/issues/4306016273/

rithviknishad (changes requested):
See #1480

Ashesh3:
> See #1480

Fixed.

rithviknishad (approved):

PR #1484, opened 2023-07-28
Added `op_no` and `ip_OR_op` Filters for Patients

Ashesh3:
## Proposed Changes

This pull request introduces two new filters for the `Patient` model:

1. Filter based on `op_no`: This filter allows users to search patients based on the `op_no` field of the `last_consultation` related model. The filter uses the `icontains` lookup expression which is case-insensitive.

2. Filter based on `ip_OR_op_no`: This filter allows users to search patients based on either `ip_no` or `op_no` fields of the `last_consultation` related model. This filter uses a custom method that applies a `Q` object to perform an OR operation between `ip_no` and `op_no` filters.

### Associated Issue

- https://github.com/coronasafe/care_fe/issues/5918

## Files Updated

1. `care/facility/api/viewsets/patient.py`: Added the new filters and their implementations.

### DIFF
--- a/care/facility/api/viewsets/patient.py
+++ b/care/facility/api/viewsets/patient.py
@@ -81,6 +81,10 @@ class PatientFilterSet(filters.FilterSet):
     ip_no = filters.CharFilter(
         field_name="last_consultation__ip_no", lookup_expr="icontains"
     )
+    op_no = filters.CharFilter(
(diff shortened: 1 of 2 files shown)


rithviknishad (approved):
Tests :)

Ashesh3:
> Tests :)

Tests added, thanks for the nudge. 😅

sainak (approved):
LGTM

PR #1502, opened 2023-08-02
Skip validation if asset warranty date is not changed

### LINKED ISSUE #1500: Warranty expiry in past is not allowing asset detail page to update
**Describe the bug**

When updating the asset details, if the warranty expiry is in the past, it becomes difficult to update other fields in the existing form.

**To Reproduce**

1. Go to https://care.coronasafe.in/facility/1193b6cb-0b31-47b8-8c5f-a5c819114359/assets/ff99d042-4e40-4908-81a7-e1e1aa3a1e3b/update
2. Update the form
3. See error

**Expected behavior**

To be discussed on EOD Call

**Screenshots**

image

rithviknishad:
## Proposed Changes

- Skip validation if asset warranty date is not changed

### Associated Issue

- Fixes #1500

## Merge Checklist

- [x] Tests added/fixed
- [ ] Update docs in `/docs`
- [x] Linting Complete


rithviknishad:
@rithviknishad to add test cases

Ashesh3 (approved):
This works fine. Though, would it be better quality wise if we merge this `if` with the latter one and only raise the Validation error in the below case? Then we'd not have to pop out data being sent by the user, since the warranty is same anyways and we just don't validate it, but it does make it a long if condition. What are your thoughts?

```
if self.instance
    and self.instance.warranty_amc_end_of_validity
    and self.instance.warranty_amc_end_of_validity != attrs["warranty_amc_end_of_validity"]
    and attrs.get("warranty_amc_end_of_validity")
    and attrs["warranty_amc_end_of_validity"] < datetime.now().date()
    :
        raise ValidationError(
```

Ashesh3 (approved):

sainak (approved):

vigneshhari (approved):

PR #1176, opened 2023-02-04
Added migrations to add skills to db

khavinshankar:
## Proposed Changes

- Added migration to add specialization skills to the database

cc: @mathew-alex 

### Associated Issue
  - None

### Architecture changes
  - None
 
@coronasafe/code-reviewers

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete

PR #1525, opened 2023-08-17
Migration to Rename Skill and Add New Skill

Ashesh3:
Fixes https://github.com/coronasafe/care_fe/issues/6091
Fixes https://github.com/coronasafe/care_fe/issues/6090

This PR introduces a new migration for the 'users' app to address two tasks:

1. **Rename Skill:**  We had a typo in one of the skill names ("Genreal Surgeon"). This migration corrects it to "General Surgeon". The migration is designed to handle the case where the Skill object doesn't exist, preventing any potential exceptions.

2. **Add New Skill:** This migration also adds a new skill named "General Medicine" to the Skill model. It uses Django's `get_or_create()` method to ensure that the new skill is only added if it doesn't already exist, preventing any duplicate entries.


rithviknishad (commented):
How did we populate skills in the first place? 

This could end up creating this skill on a fresh DB right during migrations right?

And later when we populate all skills on new instance we would end up having a duplicate right? (Since skill with the typo one is added after the corrected on was added)

Ashesh3:
> How did we populate skills in the first place?
> 
> This could end up creating this skill on a fresh DB right during migrations right?
> 
> And later when we populate all skills on a new instance we would end up having a duplicate right? (Since skill with the typo one is added during population after the corrected one was added during migration)

The skills for existing deployments were loaded in a migration: https://github.com/coronasafe/care/pull/1176/files#diff-abf84bc2d0fa38116d391f2eb563939f4bb958dbc2c19d72d774e1bff6390070R58-R59

Maybe we should separately load the Skills to the database from a JSON files like we do for the states for new deployments. This PR will correct the typo in existing deployments.

rithviknishad:
@Ashesh3 The migrations file that you've referred to seems to be part of old migrations and not present in the squashed/new migrations.

0 Skills present on a fresh instance :/

Ashesh3:
> @Ashesh3 The migrations file that you've referred to seems to be part of old migrations and not present in the squashed/new migrations.
> 
> 0 Skills present on a fresh instance :/

I am sure this isn't the expected behavior. We need the same set of skills on every instance. How about we ship this migration (since this is a P1 issue) to fix the typo in existing instances then integrate loading skills into the `load_data` management command?

Ashesh3:
We have decided to merge this for now. An additional PR will be created later to move loading skills into the `load_data` management comment.

Ashesh3:

care/users/migrations/0008_rename_skill_and_add_new_20230817_1937.py
```diff
@@ -0,0 +1,23 @@
...
+
+from django.db import migrations
+
+
+def rename_skill_and_add_new(apps, schema_editor):
+    Skill = apps.get_model("users", "Skill")
+    skill = Skill.objects.filter(name="Genreal Surgeon").first()
+    if skill:
+        skill.name = "General Surgeon"
+        skill.save()
```
sainak, line 11:
```suggestion
    Skill.objects.filter(name="Genreal Surgeon").update(name="General Surgeon")
```

rithviknishad (approved):

vigneshhari (approved):

PR #1530, opened 2023-08-19
Check bed name before making new multiple beds with same name

Pranshu1902:
## Proposed Changes

- Instead of straight away adding the number `1` behind each bed name, we first check the number from which the bed names are available and then start adding the beds. For example if the user is adding 2 beds with name `Bed` and there is already a bed with name `Bed 1`, then the new beds created will be `Bed 2` & `Bed 3` instead of `Bed 1` & `Bed 2`.

### Associated Issue

Fixes [#1529](https://github.com/coronasafe/care/issues/1529)

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete

### DIFF
--- a/care/facility/api/viewsets/bed.py
+++ b/care/facility/api/viewsets/bed.py
@@ -66,11 +66,21 @@ class BedViewSet(
         # Bulk creating n number of beds
         if number_of_beds > 1:
             data = serializer.validated_data.copy()
+            start = 1
+            while True:
+                if not Bed.objects.filter(
+                    name=f"{serializer.validated_data['name']} {start}",
+                    facility=serializer.validated_data['facility'],
+                    location=serializer.validated_data["location"]
+                ).exists():
+                    break
+                start += 1
+
             data.pop("name")
             beds = [
                 Bed(
                     **data,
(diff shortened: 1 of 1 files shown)


care/facility/api/viewsets/bed.py
```diff
@@ -66,11 +66,21 @@ def create(self, request, *args, **kwargs):
...
             data = serializer.validated_data.copy()
+            start = 1
+            while True:
+                if not Bed.objects.filter(
+                    name=f"{serializer.validated_data['name']} {start}",
+                    facility=serializer.validated_data['facility'],
+                    location=serializer.validated_data["location"]
+                ).exists():
+                    break
+                start += 1
```
sainak, line 77: you can achieve this with a single query, search and reverse sort by name

Pranshu1902:
@sainak any update regarding whether check is needed at all?

sainak:
The solution would be to index the bed names and before creating beds do a lookup of any existing beds with names 

but creating unique index can cause conflict, so this is a blocker

@vigneshhari what are your thoughts?

Pranshu1902:
Resolved in #1783

PR #1550, opened 2023-08-23
Added support to map care facilities with abdm health facility

khavinshankar:
## Proposed Changes

- Added support to map care facilities with abdm health facility

### Associated Issue

- https://github.com/coronasafe/care_fe/issues/6117

### Architecture changes

- Add HealthFacility model in abdm

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete


Ashesh3 (approved):

rithviknishad (approved):

care/abdm/api/serializers/health_facility.py
```diff
@@ -0,0 +1,9 @@
+from rest_framework import serializers
+
+from care.abdm.models import HealthFacility
+
+
+class HealthFacilitySerializer(serializers.ModelSerializer):
```
vigneshhari, line 6: Switch to use external id instead of integer ID

care/abdm/api/viewsets/health_facility.py
```diff
@@ -0,0 +1,67 @@
...
+class HealthFacilityViewSet(
+    GenericViewSet,
+    CreateModelMixin,
+    ListModelMixin,
+    RetrieveModelMixin,
+    UpdateModelMixin,
+):
+    serializer_class = HealthFacilitySerializer
+    model = HealthFacility
+    queryset = HealthFacility.objects.all()
```
vigneshhari, line 27: List and Retrieve have no authZ, was that intended?

vigneshhari (changes requested):

care/abdm/api/viewsets/health_facility.py
khavinshankar: no, intended to add it but forgot later

added it

vigneshhari (approved):

PR #1560, opened 2023-08-25
Add Hard Limit to Medicines Search

rithviknishad:
## Proposed Changes

- Add a Hard Limit of 100 to Medicines Search

## Merge Checklist

- [x] Tests added/fixed
- [x] Update docs in `/docs`
- [x] Linting Complete

### DIFF
--- a/care/facility/api/viewsets/prescription.py
+++ b/care/facility/api/viewsets/prescription.py
@@ -176,8 +176,7 @@ class MedibaseViewSet(ViewSet):
         from care.facility.static_data.medibase import MedibaseMedicineTable
 
         queryset = MedibaseMedicineTable
-
-        limit = request.query_params.get("limit", 30)
+        limit = min(int(request.query_params.get("limit", 30)), 100)
 
         if query := request.query_params.get("query"):
             query = query.strip().lower()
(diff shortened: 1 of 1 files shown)


care/facility/api/viewsets/prescription.py
```diff
@@ -176,8 +176,7 @@ def list(self, request):
         from care.facility.static_data.medibase import MedibaseMedicineTable
 
         queryset = MedibaseMedicineTable
-
-        limit = request.query_params.get("limit", 30)
+        limit = min(int(request.query_params.get("limit", 30)), 100)
```
Ashesh3, line 179: Won't it raise a `ValueError` if the "limit" query parameter is not a valid integer?

sainak: yes it will

rithviknishad: isn't that expected?

Ashesh3: I mean, it will raise an Exception at the backend rather than catching an invalid Integer and showing a message to the user, right?

rithviknishad: Got it!

Ashesh3 (approved):

sainak (approved):

vigneshhari (approved):

PR #1572, opened 2023-09-04
add field readmission + custom migration

rithviknishad:
## Proposed Changes

- Added read-only field: `is_readmission` to the Consultation model.
- Added logic to set the `is_readmission` to True during the `on_create` of a consultation.
- Custom migration to fill existing records.

### Associated Issue

- https://github.com/coronasafe/care_fe/issues/6143
- Front-end: https://github.com/coronasafe/care_fe/pull/6239

### Tested with staging db copy

image

image

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [x] Linting Complete


Ashesh3 (approved):

care/facility/migrations/0385_patientconsultation_is_readmission.py
```diff
@@ -0,0 +1,48 @@
...
+    dependencies = [
+        ("facility", "0384_patientconsultation_verified_by"),
+    ]
+
+    def fill_is_readmission(apps, schema_editor):
+        PatientConsultation = apps.get_model("facility", "PatientConsultation")
+
+        bulk_update_list = []
+        qs = PatientConsultation.objects.filter(
+            suggestion="A", admission_date__isnull=False
```
sainak, line 18:
```suggestion
            suggestion="A"
```
we should include ongoing consultations as well

care/facility/migrations/0385_patientconsultation_is_readmission.py
```diff
@@ -0,0 +1,48 @@
...
+        )
+
+        def is_readmission(consultation):
+            # readmission=True if last_consultation.discharge_date is within 30 days of admission_date
+            return qs.filter(
+                patient=consultation.patient,
+                created_date__lt=consultation.created_date,
+                discharge_date__isnull=False,
```
sainak, line 26:
```suggestion
                created_date__lt=consultation.created_date,
                discharge_date__isnull=False,
                suggestion="A"
```
we only need to consider admissions, right?

care/facility/migrations/0385_patientconsultation_is_readmission.py
rithviknishad: ongoing consultations too will have `admission_date`. What ongoing admissions won't have is a `discharge_date`. So isn't the case already handled?

Also, check the screenshot in the front-end PR. The latest ongoing consultation already has readmission=True set.

image

care/facility/migrations/0385_patientconsultation_is_readmission.py
rithviknishad: Yes. However, the filter was already applied in `qs` previously right.

https://github.com/coronasafe/care/pull/1572/files#diff-7f1e6658d9fb78cc639f1256f43aaefbc027080ecf1ceaded4499e3d360813a3R18

sainak (approved):

khavinshankar (approved):

vigneshhari (approved):

PR #1635, opened 2023-09-25
Added warranty filter to Assets

GokulramGHV:
## Proposed Changes

Added `DateFromToRangeFilter` to the AssetFilter to filter assets by their warranty expiry date.

### Associated Issue

- Related to Coronasafe/care_fe#6241

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete


care/facility/tests/test_asset_api.py
```diff
@@ -131,3 +130,29 @@ def test_asset_filter_in_use_by_consultation(self):
         response = self.client.get("/api/v1/asset/?in_use_by_consultation=false")
         self.assertEqual(response.status_code, status.HTTP_200_OK)
         self.assertEqual(len(response.data["results"]), 1)
+
+    def test_asset_filter_warranty_amc_end_of_validity(self):
+        asset1 = Asset.objects.create(
+            name="asset1",
+            current_location=self.asset_location,
+            warranty_amc_end_of_validity=datetime.now().date(),
```
Ashesh3, line 138: - Use the linter to correct the failing lint check.
- Do not directly use naive datetime to create objects that expect a timezone aware datetime object, wrap this around Django's `make_aware` method. Example: `make_aware(datetime.now().date())`
- Or instead, here you can directly use `django.utils.timezone.now` to get the present aware datetime.

Ashesh3 (changes requested):
- Use the linter to correct the failing lint check.
-

Ashesh3 (approved):

care/facility/api/viewsets/asset.py
```diff
@@ -122,6 +122,9 @@ class AssetFilter(filters.FilterSet):
         method="filter_in_use_by_consultation"
     )
     is_permanent = filters.BooleanFilter(method="filter_is_permanent")
+    warranty_amc_end_of_validity = filters.DateFromToRangeFilter(
+        field_name="warranty_amc_end_of_validity"
+    )
```
rithviknishad, line 127: is the `field_name` required in this case?

```suggestion
    warranty_amc_end_of_validity = filters.DateFromToRangeFilter()
```

Ashesh3: Shouldn't be required, since the `field_name` and attribute name both are same. 🤔

rithviknishad (approved):

sainak (approved):

vigneshhari (approved):

PR #1636, opened 2023-09-25
Refactor ICD11 Table; Cleanup in-memory search

rithviknishad:
## Proposed Changes

- All ICD11 diagnoses are loaded into the DB from now on. None is skipped.
- ID for entries with icd ID of format (`.../<possible_duplicate_id>/mms/...`) is now mapped in the following manner:

```
http://id.who.int/icd/entity/594985340                    -> 594985340  👈 Did not multiply by 1e3 as then it would break existing relations.
http://id.who.int/icd/entity/594985340/mms                -> 5949853400001
http://id.who.int/icd/entity/594985340/mms/unspecified    -> 5949853400003
```

- New ICD Entries are a superset of previous entries.
- The new table can be used to build relations (for production use, not just metabase purposes anymore)
- The old `MetaICD11Diagnosis` table can be deleted once metabase queries are updated to the new table.
- Only necessary fields: `id` and `label` are loaded onto the in-memory table.
- Lazy load the in-memory table (to prevent read from db before table is created)

### Associated Issue

- Fixes https://github.com/coronasafe/care_fe/issues/6299
- Fixes https://github.com/coronasafe/care_fe/issues/6249

### Architecture changes

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [x] Linting Complete


care/facility/static_data/icd11.py
```diff
@@ -1,43 +1,28 @@
...
 
+from django.db import connection
 from littletable import Table
 
+from care.facility.models.icd11_diagnosis import ICD11Diagnosis
+
 
 def fetch_data():
     with open("data/icd11.json", "r") as json_file:
         return json.load(json_file)
```
sainak, line 12: this is only being used in the management command so it would be better to move it there

sainak, line 12: this is only being used in the management command so it would be better to move it there

sainak (approved):

Ashesh3 (approved):

rithviknishad:
marking as hold for https://github.com/coronasafe/care_fe/issues/5293

nihal467:
LGTM, Tested in local

vigneshhari (approved):

PR #1665, opened 2023-10-09
Ensure Deterministic Ordering of Local Self Government and Ward Data

Ashesh3:
Fixes https://github.com/coronasafe/care_fe/issues/6735

In this PR, we have made changes to ensure that the Local Self Government (LSG) and Ward data are always processed in a deterministic order. This is achieved by sorting the data based on the 'name' and 'localbody_code' before it's inserted into the database. This ensures that regardless of the order in which the data is read from the JSON files, it will always be inserted into the database in the same order.

Additionally, we have also sorted the list of JSON files before processing them. This ensures that the files are always processed in the same order.

These changes help in maintaining consistency and predictability when dealing with LSG and Ward data and can be particularly useful in scenarios where the order of data matters.


rithviknishad (approved):

rithviknishad:
@sainak need review

sainak (approved):

Ashesh3:

vigneshhari (approved):

PR #1553, opened 2023-08-24
added webpush on notes creation

Bhavik-ag:
### Changes
- Create webpush notification on note creation
- Send `patient_id` and `facility_id` in the message 
- Add new notification type `PATIENT_NOTE_ADDED`
- fixes #1552

PR #1688, opened 2023-10-27
Add notifications for doctor notes

### LINKED ISSUE #1552: make patient notes realtime
### Changes
- Create webpush notification on note creation
- Send `patient_id` and `facility_id` in the message

Ashesh3:
Continuation of https://github.com/coronasafe/care/pull/1553

## Changes

- Create webpush notification on note creation
- Send `patient_id` and `facility_id` in the message 
- Add new notification type `PATIENT_NOTE_ADDED`
- fixes #1552 


nihal467:
LGTM

care/facility/api/viewsets/patient.py
```diff
@@ -649,8 +651,34 @@ def perform_create(self, serializer):
...
             created_by=self.request.user,
         )
+
+        message = {
+            "facility_id": str(patient.facility.external_id),
+            "patient_id": str(patient.external_id),
+            "from": "patient/doctor_notes/create",
+        }
+
+        NotificationGenerator(
```
vigneshhari, line 667: Do we really need 2 notifications for the same thing?

care/facility/api/viewsets/patient.py
```diff
@@ -649,8 +651,34 @@ def perform_create(self, serializer):
...
+            event=Notification.Event.MESSAGE,
```
vigneshhari, line 668: Message was meant for admin to facility comms, lets not reuse that.

care/utils/notification_handler.py
```diff
@@ -390,7 +403,10 @@ def generate(self):
                             json.dumps(
                                 {
                                     "external_id": str(notification_obj.external_id),
-                                    "title": self.message,
```
vigneshhari, line 393: Are you sure this wont break existing notifications?

vigneshhari (changes requested):

care/utils/notification_handler.py
Ashesh3: Yep, the service worker code has been updated to meet this change. This change was required because it was not possible before to send custom message without directly using the `send_webpush_user` method since there was no way to differentitate between a notification or a raw webpush message. This change allows us to do that using NotificationGenerator's `.generate`.

Service worker change: https://github.com/coronasafe/care_fe/pull/6224/files#diff-0aba05b7e68a2bb0e8cebbd6297fafc218eaeff08345a12bd379181c4a88e454L115

Service worker will auto update itself for every existing sessions on a page reload.

care/facility/api/viewsets/patient.py
Ashesh3: I believe so, we have a clear distinction between a Notification and a standalone webpush message in the frontend, the webpush hook we have does not respond if it's just a notification. We could either make it capture all Notifications but I feel that might cause some side effects since till now it only captures webpush "messages" and notification types are not sent to it.
So that's why two NotificationGenerator, one is for the actual Browser notification that the patient note was added and one is just a one time webpush message captured by webpush hook we have to refetch the patient notes. (This message is one off)

care/facility/api/viewsets/patient.py
Ashesh3: I see, I've moved it to a new type `PUSH_MESSAGE`.

care/facility/models/notification.py
```diff
@@ -23,6 +23,7 @@ class Medium(enum.Enum):
 
     class Event(enum.Enum):
         MESSAGE = 0
+        PUSH_MESSAGE = 220
```
sainak, line 26: we can use a value between 0 and 20, or move this below PATIENT_NOTE_ADDED

Ashesh3: I felt the same way, to use a value between 0 and 20, but hesitated thinking that if we did have any previous event types which occupied that range and were later removed, new notifications under the same range would have a conflicting event type with those old notification data. I did not spend much time to go through and verify that, instead i found using 220 to be simpler, I've moved this event type to the bottom.

sainak (approved):

rithviknishad (approved):

vigneshhari (approved):

PR #1689, opened 2023-10-27
fix(asset_location): added duty_staff endpoint

aeswibon:
# Feature Request

## Proposed Changes

- Adds `duty_staff` field to `asset location` model and serializer

## Associated PR

- Required by coronasafe/care_fe#6512

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [x] Linting Complete

@coronasafe/code-reviewers


nihal467:
@cp-Coder remove the maximum limit of 3 users

cp-Coder:
Okay, I will remove the validation

nihal467:
LGTM

rithviknishad (commented):
When removing a duty staff user from asset location, those records are not set to `deleted=True` instead it's permanently deleted right?

cc: @sainak

sainak:
> When removing a duty staff user from asset location, those records are not set to `deleted=True` instead it's permanently deleted right?
> 
> cc: @sainak
yeah we can use a custom through model


cc: @vigneshhari

rithviknishad (changes requested):
@cp-Coder I don't see a custom through model!

care/facility/models/asset.py
```diff
@@ -46,11 +46,45 @@ class RoomType(enum.Enum):
...
+        through_fields=("asset_location", "user"),
+        blank=True,
+    )
+
     middleware_address = models.CharField(
         null=True, blank=True, default=None, max_length=200
     )
 
 
+class AssetLocationDutyStaff(models.Model):
```
rithviknishad, line 62: Shouldn't this be inherited from BaseModel instead?

care/facility/models/asset.py
```diff
@@ -46,11 +46,45 @@ class RoomType(enum.Enum):
...
+    user = models.ForeignKey(User, on_delete=models.CASCADE, null=False, blank=False)
+
+    deleted = models.BooleanField(default=False)
+
+    created_by = models.ForeignKey(
+        User,
+        on_delete=models.PROTECT,
+        null=False,
+        blank=False,
+        related_name="assigned_duty_staff",
```
rithviknishad, line 75: Do we need this relation?

```suggestion
        related_name="+",
```

care/facility/models/asset.py
aeswibon: I don't think we need other BaseModel fields for this model. These fields are sufficient.

rithviknishad: How would you track the datetime the user was assigned to a particular location? You'd need the created_date from the BaseModel right?

rithviknishad: @sainak thoughts?

rithviknishad (changes requested):
> When removing a duty staff user from asset location, those records are not set to deleted=True instead it's permanently deleted right?

There still seems to be no API endpoint/way to unlink a staff from a location.

rithviknishad (changes requested):
But this just deletes and recreates the same objects everytime a new user is assigned to a location. Say there are three users, A, B, C. And you want to remove user C and add user D.

In the current approach, you are removing all existing ones for a particular location (A, B, C) and you create new M2M records with (A, B, D).
This performs unnecessary soft deletes and create operations for user A and B. And if we inspect the created_date of those records, it'd end up being the new date, although A and B user were linked before this datetime.

This seems unnecessary.

What's required is, two API endpoints:

`POST /location/{location__external_id}/duty_staff` which will link a user to a location.

`DELETE /location{id}/duty_staff/{external_id}` which will unlink a user from a location (soft-delete).

This would avoid unnecessary soft deletes and creates too, and the `created_date` of the m2m record also would be correct.

care/facility/api/viewsets/asset.py
```diff
@@ -118,6 +120,61 @@ def get_facility(self):
...
+                    )
+                )
+
+            AssetLocationDutyStaff.objects.bulk_create(duty_staff_objects)
+
+        return Response(status=status.HTTP_201_CREATED)
+
+    @extend_schema(tags=["asset_location"])
+    @action(methods=["POST"], detail=True)
+    def remove_duty_staff(self, request, facility_external_id, external_id):
```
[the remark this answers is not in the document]
aeswibon: I was thinking of deleting multiple users. So, I thought "POST" method would be more appropriate

rithviknishad (approved):

sainak (approved):

rithviknishad (approved):

Ashesh3 (approved):

khavinshankar (approved):

care/facility/api/viewsets/asset.py
```diff
@@ -118,6 +120,82 @@ def get_facility(self):
...
+        query = AssetLocationDutyStaff.objects.filter(
+            asset_location=asset, user__id=duty_staff, deleted=False
+        )
+
+        if query.exists():
+            raise ValidationError(
+                {"duty_staff": "Staff already assigned to the location"}
+            )
+
+        user = User.objects.filter(id=duty_staff, home_facility=asset.facility)
```
vigneshhari, line 145: This needs to be performed first, we need to check if a use exists before doing anything else, we also need to ensure that the id is an integer or perform some validation on it.

vigneshhari, line 145: use first() here itself. no need for 2 queries.

vigneshhari (changes requested):

rithviknishad:
Marking as hold as the requirements have changed and a more refined feature "User Availability" will be built that'd replace this "Duty Staff for Locations" feature.

Issue: https://github.com/coronasafe/care_fe/issues/6968

nihal467:
reopening the PR,  we can fix and merge it for now, as the larger PR of generic scheduling is under the planning phase

@rithviknishad can you clear conflict and make it ready for testing

vigneshhari:
Closing because of inactivity.

PR #1735, opened 2023-12-01
District Filter to display State's name along with the district's name

GokulramGHV:
## Proposed Changes

- Added state name to DistrictSerializer's to_representation method

### Associated Issue

- coronasafe/care_fe#6634

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete


sainak:
@nihal467 @GokulramGHV can you verify if this is not affecting other district fields?

GokulramGHV:
@sainak I've checked the frontend and I've not found any instance where district name is used directly instead of district name.
Afaik, i don't think it's affecting the other fields

sainak (approved):

Ashesh3 (approved):

rithviknishad (approved):

sainak:
@GokulramGHV can you check why the test is failing?

vigneshhari:
This might create a ton of N+1 queries if the query does not select the state as well.

vigneshhari:
This MR is going to be closed, if this feature is required, the state has to be serialized as a separate field.

PR #1737, opened 2023-12-03
Enhanced Error Handling in AssetViewSet

Ashesh3:
## Proposed Changes

In this commit, two key changes were made to the `operate_assets` function within `viewsets/asset.py`. The modifications enhance error reporting and provide clearer error messages for the user.

1. Detailed Validation Error Message: 
    In case of a `ValidationError`, the key 'message' in the response dictionary was renamed to 'detail', ensuring consistent terminology across all responses.

2. Improved APIException Handling: 
    The previous implemention simply returned the exception's `detail` and `status_code`. Now, it's refactored to a more informative error message when communication with the middleware fails. A new status (HTTP_424_FAILED_DEPENDENCY) is returned to signify that the failure is due to an unmet dependency, providing better context on the nature of the error.

image

```
  if (error.response.status > 400 && error.response.status < 500) {
        .....
              Notification.Error({
                msg: error.response.data.detail,
              });
```
The frontend expects the error to be sent in `detail` key for `4xx` errors. 
https://github.com/coronasafe/care_fe/blob/develop/src/Redux/fireRequest.tsx#L152-L164


care/facility/api/viewsets/asset.py
```diff
@@ -347,7 +347,12 @@ def operate_assets(self, request, *args, **kwargs):
             )
 
         except APIException as e:
-            return Response(e.detail, e.status_code)
+            return Response(
+                {
+                    "detail": f"Communication with the middleware failed.\nReceived status code: {e.status_code}"
+                },
+                status=status.HTTP_424_FAILED_DEPENDENCY,
```
sainak, line 354: 502 is a more appropriate for this error as 4xx are usually reserved for client errors
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/502

rithviknishad: Just a thought, maybe we could test the connection to the asset when updating the configurations, and throw a 4xx error in the update configuration API, that way we are clear it's a misconfiguration.

rithviknishad: > That'd just not show any error notification, right. Currently we do not handle showing error messages for 5xx status code, afaik.

We could show the error messages for 5xx too right? Not sure why we were not doing that. Was there a reason?

cc: @gigincg

nihal467: related PR : https://github.com/coronasafe/care_fe/pull/6873

sainak (approved):

Ashesh3:
@rithviknishad @vigneshhari could you review this?

rithviknishad (approved):

vigneshhari (approved):

PR #1748, opened 2023-12-06
Add More Facilities and patients to dummy data

Ashesh3:
## Proposed Changes

- Add More Facilities and patients to dummy data

Fixes https://github.com/coronasafe/care_fe/issues/6732


sainak (changes requested):
Deleting this will throw an error in https://github.com/coronasafe/care/blob/033d0855cdb4a602327e1f4f19808a8d706db4ed/care/facility/management/commands/load_dummy_data.py#L29-L31

as the first file is missing, the other two will fail due to constraint error.

These files are useful for setting up the backend for initial testing, so we should keep these instead and only run
https://github.com/coronasafe/care/blob/033d0855cdb4a602327e1f4f19808a8d706db4ed/care/facility/management/commands/load_dummy_data.py#L25-L28
for cypress

sainak:
> Hey, I am not deleting any files 😅 Just added more facilities and patients to the existing dummy data. I think the change size `+1 −3,128` might have been confusing since the new data is all in a single line. This is to discourage someone manually working with the json file directly.

Oh so that's why the title 😅,
formatting the json gave more clarity,

can you regenerate the fixtures by excluding, `facility.historicalpatientregistration` and `facility.historicalfacilitycapacity` they are not required and half of the file size
and the `facility_users.json` seems redundant, so both the command and file can be removed

sainak (approved):

vigneshhari (approved):

PR #1759, opened 2023-12-08
Fix ssl issues on middleware urls

sainak:
## Proposed Changes

- revert to urllib3 1.x

### Associated Issue

- urllib3 changed the verification procedure for ssl certs
- this causes domains like https://dev_middleware.coronasafe.live/ to fail ssl verification
- Weirdly [this url](https://gpg.singhaakash.dev/) works fine even though both have same CA, just the CN and sans have changed
- so until we identify the issue with our certs we are downgrading urllib3 to 1.x

image
image

### Architecture changes

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -45,6 +45,7 @@ redis = {extras = ["hiredis"], version = "==5.0.0"}
 requests = "==2.31.0"
 sentry-sdk = "==1.30.0"
 whitenoise = "==6.5.0"
+urllib3 = "==1.26.18"
 
 [dev-packages]
 black = "==23.9.1"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "d3c7b62f66862a3d2da451f26539502d9a0c5e73afa164ca789d8c4aea7085bd"
+            "sha256": "f15b2c12a4d47e7395d48029747a1fc2eb9461ecaecfc534eaac615dd06d9e8b"
         },
         "pipfile-spec": 6,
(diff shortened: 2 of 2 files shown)


rithviknishad (approved):

Ashesh3 (approved):

sainak:
As we checked more middleware instances, we discovered that this only happening with https://dev_middleware.coronasafe.live/

Ashesh3:
It appears that the root of our SSL verification issue is indeed related to the use of underscores in domain names. 

I conducted some tests and found the following:
- `test_abc.domain.com` - fails SSL verification
- `abc.domain.com` - passes SSL verification

These tests were carried out on my Cloudflare domain around 5:45. 

Further, I referred to the DigiCert knowledge base, which confirms that underscores are not allowed in Fully Qualified Domain Names (FQDNs). Here's the link for reference: [DigiCert Knowledge Base](https://www.digicert.com/kb/ssl-support/underscores-not-allowed-in-fqdns.htm).

Additionally, I checked the CA/Browser Forum's Ballot SC12 regarding the sunset of underscores in DNS names. The ballot has passed, clearly indicating a shift in standards concerning domain naming practices. You can find more details here: [CAB Forum - Ballot SC12](https://cabforum.org/2018/11/12/ballot-sc-12-sunset-of-underscores-in-dnsnames/).

Based on these findings, it's evident that the issue lies in the use of underscores in the domain name, which is not compliant with SSL hostname matching standards. Therefore, this pull request will be closed as the problem does not stem from our code.

PR #1783, opened 2023-12-26
Make bed names unique for location

### LINKED ISSUE #1782: Same bed name validation is not working while multiple bed creation
**Describe the bug**

Same bed name validation is not working while multiple bed creation, for example "test bed" name multiple beds is created from 1 to 5 , the bed created will be test bed 1 but if we already have a test bed 1, then the validation are not been checked while creating the bed

**To Reproduce**

**Expected behavior**

- [ ] 

**Screenshots**

image

sainak:
This pull request adds a unique constraint to the `Bed` model to ensure that bed names are unique within a specific location. It also includes a migration to fix any existing duplicate bed names by appending a number to the duplicate names.

fixes #1782


Ashesh3 (approved):

sainak:
the frontend displays the name in title case irrespective of the value, but the backend is validating the values as case-sensitive,
so to check for case-insensitive names should we 
change the existing values to lowercase or 
just run the validator as case-insensitive?

cc @gigincg

care/facility/migrations/0405_bed_name_constraint.py
```diff
@@ -0,0 +1,49 @@
...
+    Bed.objects.filter(old_name__isnull=False).update(name=models.F("old_name"))
+
+
+def fix_duplicate_bed_names(apps, schema_editor):
+    Bed = apps.get_model("facility", "Bed")
+    duplicates = (
+        Bed.objects.values("location", "name")
+        .annotate(count=models.Count("pk"))
+        .filter(count__gt=1)
+    )
```
rithviknishad, line 22: This does not account for case-insensitive duplicates right?

sainak: No, currently not

Ashesh3 (approved):

rithviknishad (approved):

vigneshhari (approved):

vigneshhari (approved):

PR #1784, opened 2023-12-26
add events module

sainak:
## Proposed Changes

- Brief of changes made.

### Associated Issue

### Architecture changes

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete


care/facility/models/events.py
```diff
@@ -0,0 +1,80 @@
...
+        if self.description is not None and not self.description.strip():
+            self.description = None
+        return super().save(*args, **kwargs)
+
+    def __str__(self) -> str:
+        return f"{self.model} - {self.name}"
+
+
+class PatientConsultationEvent(models.Model):
+    # external_id = use ulid
```
vigneshhari, line 48: We need some sort of non integer id for each event, i recommend making that the primary key

care/facility/events/handler.py
```diff
@@ -0,0 +1,129 @@
...
+    caused_by: int,
+    created_date: datetime = None,
+    old: Model = None,
+):
+    if created_date is None:
+        created_date = now()
+
+    if isinstance(objects, (QuerySet, list, tuple)):
+        if old is not None:
+            raise ValueError("diff is not available when objects is a list or queryset")
```
vigneshhari, line 114: Should we just log or raise a sentry issue for this for now ?

care/facility/events/handler.py
```diff
@@ -0,0 +1,129 @@
...
+        for obj in objects:
+            object_model = obj.__class__.__name__
+            create_consultation_event_task.apply_async(
```
vigneshhari, line 117: Events should always be generated live inside a transaction, No background worker should be involved.

vigneshhari: We want to be 100% certain that a change has an event associated with it. let's build it that way from the beginning.

care/facility/events/handler.py
```diff
@@ -0,0 +1,129 @@
...
+    consultation_id: int,
+    object_model: str,
+    object_id: int,
+    caused_by: int,
+    created_date: datetime,
+    diff: str = None,
+):
+    Model = apps.get_model("facility", object_model)
+
+    print(f"Creating Event for {object_model} {object_id}")
```
vigneshhari, line 46: Remove all instances of print

vigneshhari: Log if you really need to. ( chances are you don't  )

vigneshhari (changes requested):

care/facility/events/handler.py
sainak: this is a parameter validation for the dev

vigneshhari (approved):

PR #1789, opened 2023-12-28
Add consultation_filed filter in PatientFilterSet

Ashesh3:
Associated Issue: https://github.com/coronasafe/care_fe/issues/6921
Frontend PR: https://github.com/coronasafe/care_fe/pull/6933

This pull request adds a new filter called `consultation_filed` to the `PatientFilterSet` class. The filter allows users to filter patients based on whether they have a filed consultation or not. The filter is implemented using the `filter_consultation_filed` method, which applies the necessary conditions to the queryset.


rithviknishad (approved):

Ashesh3:
https://github.com/coronasafe/care_fe/issues/6921#issuecomment-1874195235

PR #1794, opened 2023-12-30
Send asset downtime status along with asset details

Pranshu1902:
## Proposed Changes

Modify the `AssetAvailabilityViewSet` query set to send the asset uptime status of all assets related to the user
image

### Associated Issue
Partially Fixes [#6819](https://github.com/coronasafe/care_fe/issues/6819)

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete

### DIFF
--- a/care/facility/api/viewsets/asset.py
+++ b/care/facility/api/viewsets/asset.py
@@ -215,17 +215,6 @@ class AssetAvailabilityViewSet(ListModelMixin, RetrieveModelMixin, GenericViewSe
     filter_backends = (filters.DjangoFilterBackend,)
     filterset_class = AssetAvailabilityFilter
 
-    # filter queryset to get all assets accessible by a user
-    def get_queryset(self):
-        queryset = AssetAvailabilityRecord.objects.all().select_related("asset")
-
-        # Check if the requesting user belongs to the related Facility
-        queryset = queryset.filter(
-            Q(asset__current_location__facility__users=self.request.user)
-            | Q(asset__current_location__facility__users__isnull=True)
-        )
-        return queryset
-
 
 class AssetViewSet(
     ListModelMixin,
--- a/care/facility/api/viewsets/asset.py
+++ b/care/facility/api/viewsets/asset.py
@@ -1,6 +1,17 @@
 from django.conf import settings
 from django.core.cache import cache
-from django.db.models import Exists, OuterRef, Q, Subquery
+from django.db.models import (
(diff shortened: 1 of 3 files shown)


rithviknishad (changes requested):
Changes required as per https://github.com/coronasafe/care_fe/pull/6952#pullrequestreview-1799610356

Pranshu1902:
@sainak could you review these changes

care/facility/api/viewsets/asset.py
```diff
@@ -252,6 +263,27 @@ def get_queryset(self):
...
+                        then=F("assetavailabilityrecord__timestamp"),
+                    ),
+                    default=Value(None),
+                    output_field=BooleanField(),
+                )
+            )
+        )
+
+        for asset in queryset:
+            setattr(asset, "down", bool(asset.latest_status))
```
sainak, line 285:
```suggestion
        last_status=Subquery(
		        AssetAvailabilityRecord.objects.filter(asset=OuterRef("pk"))
		        .order_by("-created_date")
		        .values("status")[:1]
    			)
        )
```

care/facility/api/viewsets/asset.py
```diff
@@ -252,6 +263,27 @@ def get_queryset(self):
             queryset = queryset.filter(
                 current_location__facility__id__in=allowed_facilities
             )
+
+        return queryset
+
+    def filter_queryset(self, queryset: QuerySet) -> QuerySet:
```
sainak, line 269: you could do this in the get_queryset, that way you will be able to use this to filter assets

sainak (changes requested):

sainak:
add test cases

rithviknishad (commented):
This would mean that the "AssetSerializer" can now only be used by "AssetViewset" since the field is annotated in the `get_queryset`.

If we were to use the "AssetSerializer" elsewhere, wouldn't this cause issues since we would not/need not be annotating it there.

@sainak is this correct?

rithviknishad (approved):

sainak (approved):

nihal467:
LGTM

Pranshu1902:
@vigneshhari could you review this PR

vigneshhari (approved):

vigneshhari:
It might be better to cache this data in the main record to make the query simpler.

PR #1795, opened 2023-12-30
Copy consciousness_level when cloning last daily round

Ashesh3:
Related issue: https://github.com/coronasafe/care_fe/issues/6839

We're recently added consciousness_level to normal round type, hence it should be copied when cloning the last daily round record.


rithviknishad (approved):

sainak (approved):

nihal467:
@vigneshhari can you review it :)

PR #1812, opened 2024-01-08
Support for deleting a location

Pranshu1902:
## Proposed Changes

Add support to delete an Asset Location

### Associated Issue

Partially Fixes [#6976](https://github.com/coronasafe/care_fe/issues/6976)

image

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete


sainak:
@Pranshu1902 can you add test cases where beds, assets and patients are linked to a location

Pranshu1902:
@sainak done, could you review it?

care/facility/tests/test_asset_location_api.py
```diff
@@ -78,3 +82,9 @@ def test_create_asset_location_invalid_middleware(self):
         self.assertEqual(
             response.data["middleware_address"][0].code, "invalid_domain_name"
         )
+
+    def test_delete_asset_location(self):
+        response = self.client.delete(
+            f"/api/v1/facility/{self.facility.external_id}/asset_location/{self.asset_location.external_id}/",
+        )
+        self.assertEqual(response.status_code, status.HTTP_204_NO_CONTENT)
```
rithviknishad, line 90: @Pranshu1902  Kindly read the issue correctly. This test should assert that the API rejects deletion of such locations.

image

rithviknishad (changes requested):
I don't see any implementation of where the API rejects for the reasons specified in the issue

image

care/facility/api/viewsets/asset.py
```diff
@@ -118,6 +119,15 @@ def get_facility(self):
     def perform_create(self, serializer):
         serializer.save(facility=self.get_facility())
 
+    def destroy(self, request, *args, **kwargs):
+        instance = self.get_object()
+        if instance.bed_set.count():
+            raise ValidationError("Cannot delete a Location with associated Beds")
+        if instance.asset_set.count():
+            raise ValidationError("Cannot delete a Location with associated Assets")
```
rithviknishad, line 127: Since we have soft deletes, do reverse relations queryset account for it? Shouldn't we filter by `deleted=False`?

cc: @sainak 

```suggestion
        if instance.bed_set.filter(deleted=False).count():
            raise ValidationError("Cannot delete a Location with associated Beds")
        if instance.asset_set.filter(deleted=False).count():
            raise ValidationError("Cannot delete a Location with associated Assets")
```

rithviknishad (approved):

Ashesh3 (approved):

sainak (approved):

nihal467:
Functionality LGTM

Pranshu1902:
@vigneshhari could you review this PR

vigneshhari (approved):

PR #1818, opened 2024-01-10
Fix patient discharge_reason filter and add it to export

Ashesh3:
This pull request fixes the patient discharge_reason filter in the patient API viewset and adds it to the export functionality.


rithviknishad (approved):

khavinshankar (approved):

sainak:
We can keep the old names

Ashesh3:
> We can keep the old names

That's alright, We'll switch everything back to `discharge_reason` after a week of production, in the next stage as planned. Let's just keep it distinct for now to follow along the changes made 😅

sainak (approved):

vigneshhari (approved):

PR #1819, opened 2024-01-10
Add nurse user type

sainak:
This pull request adds a new user type called "Nurse" and migrates the existing staff users to the nurse role. It also includes updates to permissions checks.


rithviknishad (approved):

Ashesh3 (approved):

care/facility/api/viewsets/file_upload.py
```diff
@@ -54,6 +55,16 @@ def get_serializer_class(self):
     def get_queryset(self):
         if "file_type" not in self.request.GET:
             raise ValidationError({"file_type": "file_type missing in request params"})
+
+        if self.request.user.user_type in (
```
vigneshhari, line 59: Maybe switch this to a permission check, kinda misplaced in the query set. A new Permission Check would be great.

vigneshhari (changes requested):

sainak (commented):
move permission checks out of geet_queryset

vigneshhari (approved):
Approved without review.

PR #1823, opened 2024-01-16
Allows diagnoses starting with disease code to be available in ICD11 Search

rithviknishad:
## Proposed Changes

- Allows diagnoses starting with disease code to be available in ICD11 searched regardless of whether it's a chapter/category/leaf entity or not

### Associated Issue

- Fixes https://github.com/coronasafe/care_fe/issues/7042

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete

### DIFF
--- a/care/facility/api/viewsets/icd.py
+++ b/care/facility/api/viewsets/icd.py
@@ -22,11 +22,10 @@ class ICDViewSet(ViewSet):
     def list(self, request):
         from care.facility.static_data.icd11 import ICDDiseases
 
-        queryset = ICDDiseases
+        queryset = ICDDiseases.where(has_code=True)
         if request.GET.get("query", False):
             query = request.GET.get("query")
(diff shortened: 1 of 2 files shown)


Ashesh3 (approved):

sainak (approved):

vigneshhari (approved):

vigneshhari:
Fix tests and we should be good to merge

rithviknishad:
@vigneshhari the issue was when running the tests, the in-memory table was being built before the DB was populated with the ICD11 data. Hence 0 records were loaded onto the in-memory table.

It was not being lazy loaded accidentally due to other functions in the same file being imported from other places. I've now ensured that ICD11 is always lazy-loaded.

vigneshhari (approved):

PR #1824, opened 2024-01-16
Added whitelist of file types in file upload

khavinshankar:
## Proposed Changes

- Added whitelist of file types in file upload (currently allowing images, videos, audios, pdf, and docs)

### Associated Issue

- wrt Beagal security audit report

## Merge Checklist

- [x] Tests added/fixed
- [ ] Update docs in `/docs`
- [x] Linting Complete


Ashesh3 (approved):
LGTM
The whitelist seems really narrow. We can add more as the requirements roll in.
Also, should this be moved to the [`FileUploadCreateSerializer`](https://github.com/coronasafe/care/blob/fileupload-whitelist/care/facility/api/serializers/file_upload.py#L126) serializer? @sainak

rithviknishad (commented):
This would prevent users from uploading DICOM images right?

rithviknishad (approved):

sainak (approved):

care/facility/api/serializers/file_upload.py
```diff
@@ -120,19 +122,28 @@ class Meta:
             "signed_url",
             "internal_name",
             "original_name",
+            "mime_type",
         )
-        write_only_fields = ("associating_id",)
+        write_only_fields = ("associating_id", "mime_type")
```
vigneshhari, line 127: We can remove this.

care/facility/api/serializers/file_upload.py
```diff
@@ -120,19 +122,28 @@ class Meta:
...
 
     def create(self, validated_data):
         user = self.context["request"].user
+        mime_type = validated_data["mime_type"]
```
vigneshhari, line 131: use pop to remove and fetch at the same time.

care/facility/api/serializers/file_upload.py
```diff
@@ -120,19 +122,28 @@ class Meta:
...
+
+        if mime_type not in settings.ALLOWED_MIME_TYPES:
+            raise ValidationError("Invalid File Type")
```
vigneshhari, line 134: Validation Error should be a json maybe something like ```{"detail" : "Invalid FIle Type"}```

vigneshhari, line 134: Use the Validation error from the DRF package.

config/settings/base.py
```diff
@@ -509,6 +509,45 @@
     default=f"https://{FILE_UPLOAD_BUCKET}.s3.amazonaws.com",
 )
 
+ALLOWED_MIME_TYPES = env.json(
+    "ALLOWED_MIME_TYPES",
+    default=[
```
vigneshhari, line 514: Either switch this to a `list` or a `tuple` method from `env` or make the default value a json object.

care/facility/models/file_upload.py
```diff
@@ -83,14 +83,17 @@ def save(self, *args, **kwargs):
...
 
-    def signed_url(self, duration=60 * 60):
+    def signed_url(self, duration=60 * 60, mime_type=None):
         s3Client = boto3.client("s3", **cs_provider.get_client_config())
+        params = {
+            "Bucket": settings.FILE_UPLOAD_BUCKET,
+            "Key": f"{self.FileType(self.file_type).name}/{self.internal_name}",
+        }
+        if mime_type:
+            params["ContentType"] = mime_type
```
vigneshhari, line 93: Have you checked with a live S3 instance and tried to send an invalid payload? If so what was the response?

vigneshhari (changes requested):
Changes Required

care/facility/models/file_upload.py
khavinshankar: When we set a content type and upload a different file (with a different mime type), the file is getting uploaded to the bucket, but it's not rendered in the iframe when fetched.

khavinshankar: @vigneshhari 

I tried using generate_presigned_post too, still, we can bypass the ContentType value set while creating the pre-signed URL, any suggestions to make this work? As a last option, we could proxy the file content through the backend!

care/facility/api/serializers/file_upload.py
```diff
@@ -120,19 +122,27 @@ class Meta:
...
+            "mime_type",
         )
         write_only_fields = ("associating_id",)
 
     def create(self, validated_data):
         user = self.context["request"].user
+        mime_type = validated_data.pop("mime_type")
+
+        if mime_type not in settings.ALLOWED_MIME_TYPES:
+            raise ValidationError({"detail": "Invalid FIle Type"})
```
vigneshhari, line 134:
```suggestion
            raise ValidationError({"detail": "Invalid File Type"})
```

vigneshhari (approved):

PR #1866, opened 2024-02-01
Adds view for listing discharged patients of a facility

rithviknishad:
## Proposed Changes

- Adds view for listing discharged patients of a facility
- Support for filtering the view's queryset by patient name

### Associated Issue

- https://github.com/coronasafe/care_fe/issues/4678

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete
- [ ] Any other necessary step

_*Only PR's with test cases included and passing lint and test pipelines will be reviewed*_

@coronasafe/care-backend-maintainers @coronasafe/care-backend-admins

PR #1868, opened 2024-02-01
added user type check before creating new asset

kshitijv256:
## Proposed Changes

- Added user type check before creating new asset

### Associated Issue

- https://github.com/coronasafe/care_fe/issues/7081


sainak (changes requested):
use dry permissions for this validation
https://github.com/dbkaplan/dry-rest-permissions

rithviknishad (approved):

rithviknishad (approved):

care/facility/api/serializers/asset.py
```diff
@@ -213,6 +214,7 @@ def validate(self, attrs):
         return super().validate(attrs)
 
     def create(self, validated_data):
+        user = self.context["request"].user
```
sainak, line 217: where is this being used?

sainak:
add test cases

kshitijv256:
> add test cases

@sainak 
I am unable to create a test case as different user type is needed, can you give some idea how

rithviknishad:
@kshitijv256 example: https://github.com/coronasafe/care/pull/1866/files#diff-6e69c1bec5735c00ad8c3637de1e270b87399c4a6958c4496a9d145e3025d333R90

PR #1869, opened 2024-02-02
Track occupation of patient

konavivekramakrishna:
## Proposed Changes

- change occupation choices
-  change head_of_household can be null

### Associated Issue
- https://github.com/coronasafe/care_fe/issues/6945


konavivekramakrishna:
@rithviknishad 

 can you please review this? I am facing an issue: if a patient first selects an occupation, later, if the patient unselects it, the system should either delete the record or simply keep the occupation field as null

rithviknishad:
@sainak, how about we keep a char choices field `occupation` in the patient registration table to keep it simple instead?

care/facility/models/patient.py
```diff
@@ -571,17 +571,16 @@ def format_diagnoses(diagnosis_ids):
 class PatientMetaInfo(models.Model):
     class OccupationEnum(enum.Enum):
         STUDENT = 1
-        MEDICAL_WORKER = 2
-        GOVT_EMPLOYEE = 3
-        PRIVATE_EMPLOYEE = 4
-        HOME_MAKER = 5
-        WORKING_ABROAD = 6
-        OTHERS = 7
```
sainak, line 579: we should keep the existing choices

care/facility/api/serializers/patient.py
```diff
@@ -360,10 +360,18 @@ def update(self, instance, validated_data):
...
+                    patient.meta_info = meta_info_obj
+                else:
+                    for key, value in meta_info.items():
+                        setattr(patient.meta_info, key, value)
                 patient.meta_info.save()
 
+            else:
+                #patient.save()
+                if patient.meta_info:
+                    patient.meta_info.delete()
```
sainak, line 374: Instead of deleting the meta object just set the valuse to none

https://www.django-rest-framework.org/api-guide/serializers/#writable-nested-representations

sainak (changes requested):

sainak:
> @sainak, how about we keep a char choices field `occupation` in the patient registration table to keep it simple instead?

meta table is a good to have, it helps in keeping the main table less cluttered

care/facility/models/patient.py
konavivekramakrishna: @sainak but In the issue, only a few options for occupation were specified https://github.com/coronasafe/care_fe/issues/6945#issue-2059884492

konavivekramakrishna: https://github.com/coronasafe/care_fe/issues/6945#issue-2059884492

@sainak In the issue, only a few options for occupation were specified

rithviknishad: @konavivekramakrishna that'd cause old data to be mixed with new data. A medical worker patient before would end up becoming a businessman after this.

@sainak but we were not recording these data so far (only 22 records are there in staging which was captured years ago). So if we just clean the staging, we could have a cleaner migration right?

sainak: yes we can after confirmation from prod

konavivekramakrishna: @rithviknishad @sainak 

can you please conform regarding the occupation choices

konavivekramakrishna: should I add new choices to existing choices or remove them

konavivekramakrishna:
@rithviknishad 

can you please review the changes

care/facility/models/patient.py
rithviknishad: @sainak we are safe to delete.

@Ashesh3 had checked the prod. instances

cc: @konavivekramakrishna

care/facility/models/patient.py
```diff
@@ -571,17 +571,15 @@ def format_diagnoses(diagnosis_ids):
...
+        HEALTH_CARE_WORKER = 3
+        HEALTH_CARE_LAB_WORKER =4
+        ANIMAL_HANDLER = 5
+        OTHERS = 6
     OccupationChoices = [(item.value, item.name) for item in OccupationEnum]
 
-    occupation = models.IntegerField(choices=OccupationChoices)
-    head_of_household = models.BooleanField()
+    occupation = models.IntegerField(choices=OccupationChoices,blank=True, null=True)
+    head_of_household = models.BooleanField( blank=True,null=True)
```
sainak, line 582:
```suggestion
    head_of_household = models.BooleanField(blank=True, null=True)
```

lint

care/facility/models/patient.py
```diff
@@ -571,17 +571,15 @@ def format_diagnoses(diagnosis_ids):
...
+        BUSINESSMAN = 2
+        HEALTH_CARE_WORKER = 3
+        HEALTH_CARE_LAB_WORKER =4
+        ANIMAL_HANDLER = 5
+        OTHERS = 6
     OccupationChoices = [(item.value, item.name) for item in OccupationEnum]
 
-    occupation = models.IntegerField(choices=OccupationChoices)
-    head_of_household = models.BooleanField()
+    occupation = models.IntegerField(choices=OccupationChoices,blank=True, null=True)
```
sainak, line 581:
```suggestion
    occupation = models.IntegerField(choices=OccupationChoices, blank=True, null=True)
```
lint

sainak:
@konavivekramakrishna run precommit on the changed files

https://github.com/coronasafe/care/blob/master/CONTRIBUTING.md#getting-started

PR #1870, opened 2024-02-02
draft add insurer name in csv export

konavivekramakrishna:
## Proposed Changes

- add insurer name in patients csv export
### Associated Issue
- https://github.com/coronasafe/care_fe/issues/6517

image


care/facility/models/patient.py
```diff
@@ -530,6 +530,9 @@ def annotate_diagnosis_ids(*args, **kwargs):
         "last_consultation__new_discharge_reason": "Reason for discharge",
         "last_consultation__discharge_date": "Date of discharge",
         "last_consultation__discharge_date__time": "Time of discharge",
+        
+        # insurance details
+         "policy__insurer_name": "Insurer Name",
```
rithviknishad, line 535: What happens if there are multiple insurance policies?

konavivekramakrishna: @rithviknishad 

when trying to import policy, getting error as circular dependency how to solve this.

rithviknishad: In the function passed as the value in CSV_MAKE_PRETTY for this, you can import it the model inside the function.

@sainak is there a better solution?

sainak: we can move this to patient viewset

PR #1876, opened 2024-02-07
Add consultation serializer for Investigation

Ashesh3:
This pull request adds a consultation serializer for the Investigation model. It includes changes to the `patient_investigation.py` file in the `care/facility/api/serializers` directory.


rithviknishad (changes requested):
Related viewset's queryset is not having `selecte_related("consultation")` to prevent N+1 queries

rithviknishad (approved):

sainak (approved):

vigneshhari (approved):

PR #1900, opened 2024-02-20
Add assigned_clinicians field to PatientConsultation model

Ashesh3:
This pull request adds the `assigned_clinicians` field to the `PatientConsultation` model, allowing multiple clinicians to be assigned to a patient consultation.

Phases for the change
1. [Schema Changes](https://github.com/coronasafe/care/pull/1900)
2. [Migrating Data and Deprecating assigned_to](https://github.com/coronasafe/care/pull/1903)

Related Issue: https://github.com/coronasafe/care_fe/issues/6276


sainak (approved):

care/facility/models/patient_consultation.py
```diff
@@ -168,6 +168,10 @@ class PatientConsultation(PatientBaseModel, ConsultationRelatedPermissionMixin):
         related_name="patient_assigned_to",
     )
 
+    assigned_clinicians = models.ManyToManyField(
+        User, related_name="patient_assigned_clinician"
```
sainak, line 172:
```suggestion
        User, related_name="assigned_patients"
```

rithviknishad (approved):

Ashesh3:

care/facility/models/patient_consultation.py
```diff
@@ -338,3 +344,14 @@ def has_object_email_discharge_summary_permission(self, request):
 
     def has_object_generate_discharge_summary_permission(self, request):
         return self.has_object_read_permission(request)
+
+
+class ConsultationClinician(models.Model):
```
rithviknishad, line 349: BaseModel? @sainak

Ashesh3: Question would be, if we need to store the extra info that the BaseModel provides, `external_id, created_date, modified_date` and also soft delete for this table which is just storing the m2m relations. 🤔

sainak: Its okay without it, we can track the changes in events

sainak (approved):

Ashesh3:

vigneshhari (approved):

PR #1903, opened 2024-02-21
Migrate consultation's assigned_to to assigned_clinicians

Ashesh3:
Requires: https://github.com/coronasafe/care/pull/1900
This pull request migrates the `assigned_to` field in the `consultation` model to `assigned_clinicians`. This change is made to improve clarity and consistency in the codebase.

Phases for the change
1. [Schema Changes](https://github.com/coronasafe/care/pull/1900)
2. [Migrating Data and Deprecating assigned_to](https://github.com/coronasafe/care/pull/1903)

Related Issue: https://github.com/coronasafe/care_fe/issues/6276


Ashesh3:
@sainak @rithviknishad @vigneshhari could you guys review this once?

care/facility/migrations/0422_auto_20240220_1233.py
```diff
@@ -0,0 +1,38 @@
...
+
+def transfer_assigned_to_to_clinicians(apps, schema_editor):
+    PatientConsultation = apps.get_model("facility", "PatientConsultation")
+
+    for patient_consultation in PatientConsultation.objects.all():
+        if patient_consultation.assigned_to:
+            # Direct assignment to ManyToManyField not possible; need to use .add()
+            patient_consultation.assigned_clinicians.add(
+                patient_consultation.assigned_to
+            )
```
rithviknishad: We could filter by `assigned_to__isnull=False` and do a bulk create instead right?

care/facility/api/viewsets/patient.py
```diff
@@ -275,7 +275,7 @@ def filter_queryset(self, request, queryset, view):
                 q_filters = Q(facility__id__in=allowed_facilities)
                 if view.action == "retrieve":
                     q_filters |= Q(consultations__facility__id__in=allowed_facilities)
-                q_filters |= Q(last_consultation__assigned_to=request.user)
+                q_filters |= Q(last_consultation__assigned_clinicians=request.user)
```
sainak, line 278: Im not sure if this filter will work, assigned_clinicians is a m2m field now

care/facility/api/viewsets/patient_investigation.py
```diff
@@ -139,7 +139,7 @@ def get_queryset(self):
             )
         allowed_facilities = get_accessible_facilities(self.request.user)
         filters = Q(consultation__patient__facility_id__in=allowed_facilities)
-        filters |= Q(consultation__assigned_to=self.request.user)
+        filters |= Q(consultation__assigned_clinicians=self.request.user)
```
sainak, line 142: verify this filter
another occurrence on L187

care/users/api/viewsets/users.py
```diff
@@ -63,8 +63,13 @@ def get_user_type(
         value,
     ):
         if value:
```
sainak, line 66:
```suggestion
        # TODO: use https://django-filter.readthedocs.io/en/stable/ref/filters.html#multiplechoicefilter
        if value:
```

care/utils/notification_handler.py
```diff
@@ -144,23 +144,23 @@ def deserialize_extra_data(self, extra_data):
 
     def generate_extra_users(self):
         if isinstance(self.caused_object, PatientConsultation):
-            if self.caused_object.assigned_to:
-                self.extra_users.append(self.caused_object.assigned_to.id)
+            for clinician in self.caused_object.assigned_clinicians.all():
```
sainak:
```suggestion
            for clinician in self.caused_object.assigned_clinicians.all().only("id"):
```

we dont require all fields here so lets improve the query
Similar occurrences below V

sainak (commented):
One more thing to note: m2m lookups can result in duplicates

care/utils/notification_handler.py
Ashesh3: @sainak Interesting suggestion!

I was reading through https://docs.djangoproject.com/en/5.0/topics/db/optimization/#use-queryset-defer-and-only and was curious to profile the two queries, but while doing so, I am unable to use `.only` on `User` model.

(Trying these out via django shell)

```
PatientConsultation.objects.get(id=2268).assigned_clinicians.all().only("id")

FieldError: Field User.local_body cannot be both deferred and traversed using select_related at the same time.
Stacktrace: https://rentry.co/xourynxu/raw
```

The same error can also be replicated via `User.objects.all().only("id")`.

It appears that the `care.users.models.CustomUserManager` class is using local_body in `selected_related` which might be causing this issue.
```
class CustomUserManager(UserManager):
    def get_queryset(self):
        qs = super().get_queryset()
        return qs.filter(deleted=False, is_active=True).select_related(
            "local_body", "district", "state"
        )
```

Is there any way to work around this limitation of not being able to couple `.only` and `select_related`?

care/facility/migrations/0422_auto_20240220_1233.py
Ashesh3: Done!

Ashesh3:
> m2m lookups can result in duplicates

Hey, could you elaborate a bit more on this? Lookups can result in duplications? do you mean that a select query might return the same record twice in some cases? if so, which cases?

care/utils/notification_handler.py
Ashesh3: Resolved after discussing in DMs.

Ashesh3:
@rithviknishad @sainak @khavinshankar @vigneshhari could you guys review this?

Issue #1338, opened 2023-06-02
Fix N+1 queries in `/api/v1/bed/`


sainak:
Sentry Issue: [CARE-KA](https://coronasafe-network.sentry.io/issues/3980443680/?referrer=github_integration)
|  |  |
| ------------- | --------------- |
| **Transaction Name** | /api/v1/bed/ |
| **Parent Span** | django.view - bed-list |
| **Repeating Spans (0)** | db - SELECT facility_facility.id, facility_facility.external_id, facility_facility.created_date, facility_facility.modified_date, facility_facility.deleted, facility_facility.name, facility_facility.is_active, facility_facility.verified, facility_facility.facility_type, facility_facility.kasp_empanelled, facility_facility.features, facility_facility.longitude, facility_facility.latitude, facility_facility.pincode, facility_facility.address, "facility... |
||db - SELECT (1) AS a FROM facility_consultationbed WHERE (facility_consultationbed.deleted = %s AND facility_consultationbed.bed_id = %s AND facility_consultationbed.end_date IS NULL) LIMIT 1|

Yash12Bansal:
@sainak please assign this issue to me

jasonxiexy:
Hi @sainak,

I would like to contribute to the project by addressing the Fix N+1 queries issue (#1338) in the /api/v1/bed/ endpoint. Could I please be assigned to work on this task? I believe I can help optimize the queries and improve performance in this area.

Thank you,
Jason Xie

himanshu-sharmav:
Hi @sainak, I would like to work on this issue.

Harsh-D-2004:
@sainak Can u assign me this??

PR #1908, opened 2024-02-23
Fixed N+1 queries issue in /api/v1/bed/

dhruv-goyal-10:
## Proposed Changes

- **is_occupied** is the property defined in the **Bed Model**, which was contributing to N + 1 queries issue, and I have annotated the attribute **_is_occupied** in the queryset, which basically corresponds to the same value as the **is_occupied property** in Bed Model. I have handled the various attribute cases, in **BedSerializer**, by defining **is_occupied** field as **SerializerMethodField**, hence there will be no change in API response.

- Have added **location__facility** in the select_related field, as this was also contributing to the N + 1 queries issue.

### Associated Issue

- Solves #1338

## Merge Checklist

- [x] Tests added/fixed
- [ ] Update docs in `/docs`
- [x] Linting Complete

### DIFF
--- a/care/facility/api/serializers/bed.py
+++ b/care/facility/api/serializers/bed.py
@@ -83,8 +83,6 @@ class BedSerializer(ModelSerializer):
         return super().validate(attrs)
 
     def get_is_occupied(self, instance):
-        if hasattr(instance, "_is_occupied"):
-            return instance._is_occupied
         return instance.is_occupied
 
 
--- a/care/facility/api/serializers/bed.py
+++ b/care/facility/api/serializers/bed.py
@@ -4,6 +4,7 @@ from django.shortcuts import get_object_or_404
 from django.utils import timezone
 from rest_framework.exceptions import ValidationError
 from rest_framework.serializers import (
+    BooleanField,
     CharField,
     DateTimeField,
     IntegerField,
@@ -39,7 +40,7 @@ class BedSerializer(ModelSerializer):
     bed_type = ChoiceField(choices=BedTypeChoices)
 
     location_object = AssetLocationSerializer(source="location", read_only=True)
-    is_occupied = SerializerMethodField()
+    is_occupied = BooleanField(default=False, read_only=True)
 
     location = UUIDField(write_only=True, required=True)
     facility = UUIDField(write_only=True, required=True)
@@ -81,9 +82,6 @@ class BedSerializer(ModelSerializer):
                 {"location": "Field is Required", "facility": "Field is Required"}
             )
         return super().validate(attrs)
-
-    def get_is_occupied(self, instance):
-        return instance.is_occupied
 
 
 class AssetBedSerializer(ModelSerializer):
--- a/care/facility/api/serializers/bed.py
+++ b/care/facility/api/serializers/bed.py
@@ -4,7 +4,6 @@ from django.shortcuts import get_object_or_404
 from django.utils import timezone
 from rest_framework.exceptions import ValidationError
 from rest_framework.serializers import (
-    BooleanField,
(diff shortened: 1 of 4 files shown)


dhruv-goyal-10:
@sainak @rithviknishad @vigneshhari Can you please review this PR?

sainak:
hey @dhruv-goyal-10 thanks for the pr,
after looking at the bed model i can see that the is_occupied property on the model is not being used
https://github.com/coronasafe/care/blob/6cb4b3f26c03bece09f818299f4dccfb321ab86b/care/facility/models/bed.py#L43-L45

so we can just annotate the queryset with is_occupied directly if `self.action == list`

dhruv-goyal-10:
@sainak @rithviknishad @vigneshhari I have completed the changes as suggested by @sainak. Can you review my PR now?

care/facility/api/serializers/bed.py
sainak, line 1: is_occupied = BooleanField(default=False, read_only=True)
is correct, you can revert the changes in this file

dhruv-goyal-10:
Hey @sainak! I have done changes suggested by you. Can you please have a look at my PR now?

sainak (approved):

Ashesh3 (approved):

sainak:
> @sainak I later realised this, commenting out the is_occupied property will although not affect the results in case of list method, but it will affect the result in case of retrieve method (fetching a particular object using id), coz we aren't annotating is_occupied property in that case and it will always corresponds to False (the default value).
> Isn't it?

Currently we are not using bed retrieve on Fe but you are correct that will give inconsistent results, It won't affect the performance so lets remove the if condition

care/facility/api/viewsets/bed.py
```diff
@@ -63,6 +63,16 @@ class BedViewSet(
...
         queryset = self.queryset
+
+        if self.action == "list":
+            queryset = queryset.annotate(
+                is_occupied=Exists(
+                    ConsultationBed.objects.filter(
+                        bed__id=OuterRef("id"), end_date__isnull=True
+                    )
+                )
+            )
```
sainak, line 74:
```suggestion
        queryset = queryset.annotate(
            is_occupied=Exists(
                ConsultationBed.objects.filter(
                    bed__id=OuterRef("id"), end_date__isnull=True
                )
            )
        )
```

dhruv-goyal-10:
@sainak Made the changes as suggested by you.
Also, thanks for an insightful discussion.

rithviknishad (approved):

dhruv-goyal-10:
@sainak Do I need to write test cases for this PR to get merged?

sainak:
Yes!
For this case create a bunch of records and checked the number of queries executed

dhruv-goyal-10:
@sainak I have added the test case for this as suggested by you, Please have a look now!

care/facility/tests/test_bed_api.py
```diff
@@ -0,0 +1,37 @@
...
+        )
+        self.bed2 = Bed.objects.create(
+            name="bed2", location=self.asset_location, facility=self.facility
+        )
+        self.bed3 = Bed.objects.create(
+            name="bed3", location=self.asset_location, facility=self.facility
+        )
+
+    def test_list_beds(self):
+        with self.assertNumQueries(5):
```
sainak, line 35:
```suggestion
        # includes 3 queries for auth and 1 for pagination count
        with self.assertNumQueries(5):
```

sainak (approved):

Ashesh3 (approved):

vigneshhari (approved):

PR #1909, opened 2024-02-23
Fixed N+1 queries issue in /api/v1/consultationbed/

### LINKED ISSUE #1339: Fix N+1 queries in `/api/v1/consultationbed/`
Sentry Issue: [CARE-KD](https://coronasafe-network.sentry.io/issues/4000487637/?referrer=github_integration)
|  |  |
| ------------- | --------------- |
| **Transaction Name** | /api/v1/consultationbed/ |
| **Parent Span** | django.view - consultationbed-list |
| **Repeating Spans (0)** | db - SELECT facility_assetlocation.id, facility_assetlocation.external_id, facility_assetlocation.created_date, facility_assetlocation.modified_date, facility_assetlocation.deleted, facility_assetlocation.name, facility_assetlocation.description, facility_assetlocation.location_type, facility_assetlocation.facility_id FROM facility_assetlocation WHERE facility_assetlocation.id = %s |
||db - SELECT facility_facility.id, facility_facility.external_id, facility_facility.created_date, facility_facility.modified_date, facility_facility.deleted, facility_facility.name, facility_facility.is_active, facility_facility.verified, facility_facility.facility_type, facility_facility.kasp_empanelled, facility_facility.features, facility_facility.longitude, facility_facility.latitude, facility_facility.pincode, facility_facility.address, "facility... |
||db - SELECT (1) AS a FROM facility_consultationbed WHERE (facility_consultationbed.deleted = %s AND facility_consultationbed.bed_id = %s AND facility_consultationbed.end_date IS NULL) LIMIT 1|

dhruv-goyal-10:
## Proposed Changes

- The related fields of **Bed** like **"bed__facility", "bed__location", "bed__location__facility"** were causing the N + 1 queries issue, hence added them in select_related.

- The related fields of prefetched_model **asset** was also causing the N + 1 queries issue, hence added the related fields of it in select_related, while using **Prefetch() object**.

- To solve the N + 1 query issue related to **is_occupied** property of **Bed Model**, I have made necessary changes in 
**PatientConsultationSerializer**, to set the **_is_occupied** property of Bed instance, whose value is equal to the **is_occupied** property of **Bed Model**.

### Associated Issue

- Fixes #1339 

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete

### DIFF
--- a/care/facility/api/serializers/bed.py
+++ b/care/facility/api/serializers/bed.py
@@ -4,7 +4,6 @@ from django.shortcuts import get_object_or_404
 from django.utils import timezone
 from rest_framework.exceptions import ValidationError
 from rest_framework.serializers import (
-    BooleanField,
     CharField,
     DateTimeField,
     IntegerField,
@@ -40,7 +39,7 @@ class BedSerializer(ModelSerializer):
     bed_type = ChoiceField(choices=BedTypeChoices)
 
     location_object = AssetLocationSerializer(source="location", read_only=True)
-    is_occupied = BooleanField(default=False, read_only=True)
+    is_occupied = SerializerMethodField()
 
     location = UUIDField(write_only=True, required=True)
     facility = UUIDField(write_only=True, required=True)
@@ -54,6 +53,11 @@ class BedSerializer(ModelSerializer):
         if value > 100:
             raise ValidationError("Cannot create more than 100 beds at once.")
         return value
+
+    def get_is_occupied(self, instance):
+        if hasattr(instance, "_is_occupied"):
+            return instance._is_occupied
+        return instance.is_occupied
 
     class Meta:
         model = Bed
@@ -348,3 +352,11 @@ class ConsultationBedSerializer(ModelSerializer):
         validated_data.pop("assets", None)
 
         return super().update(instance, validated_data)
+
+    def to_representation(self, instance: ConsultationBed):
+        if hasattr(instance, "bed"):
+            if instance.end_date is None:
+                instance.bed._is_occupied = True
+            else:
+                instance.bed._is_occupied = False
+        return super().to_representation(instance)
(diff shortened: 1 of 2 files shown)


dhruv-goyal-10:
@sainak @rithviknishad @vigneshhari Can you please review this PR?

sainak:
https://github.com/coronasafe/care/pull/1908#issuecomment-1963925609

sainak:
> @sainak In this particular case, although commenting out is_property field of Bed Model will not affect the results of **ConsultationBedSerializer**, as we are going to handle the is_occupied property of Bed in this serializer.
> 
> But in other serializers like **AssetBedSerializer, PatientAssetBedSerializer**, which are also using **BedSerializer**, since we are not handling the is_occupied property of Bed in these serializers, so it will always correspond to the False (the default value).
> 
> What are your thoughts?

you can use the double underscore syntax `__` to annotate the is_occupied property on the bed model it should be picked up by the serializer field

dhruv-goyal-10:
@sainak But since **Bed** is the related object, how can we annotate the property of Bed while fetching the objects of **ConsultationBed**.
Because, as per my knowledge, we can only annotate the property of the objects present in the queryset and not of the related object, present in the queryset.

sainak:
> @sainak But since **Bed** is the related object, how can we annotate the property of Bed while fetching the objects of **ConsultationBed**. Because, as per my knowledge, we can only annotate the property of the objects present in the queryset and not of the related object, present in the queryset.

ive seen some examples, and i think it should work, give it a try

sainak:
@dhruv-goyal-10 
```python
ConsultationBed.objects.all().annotate(
    bed__is_occupied=Exists(
        ConsultationBed.objects.filter(bed__id=OuterRef("id"), end_date__isnull=True)
    )
).values_list("bed__is_occupied", flat=True)
```

Issue #1541, opened 2023-08-22
Move the Investigations and Investigation Groups into JSON


mathew-alex:
Move the Investigations and Investigation Groups into a proper JSON like python dict structure to allow easy updates and additions.
The current way is loading data from a string format 
https://github.com/coronasafe/care/blob/master/care/users/management/commands/populate_investigations.py#L9

The investigation list should contain all items from this sheet.
https://docs.google.com/spreadsheets/d/1hN13p-BDbb1Pkl3a8tcVdWFsFxncju2QPfeOEkKh51E/edit#gid=169054315

Both of the below constraints should be valid.
1. An investigation can be associated with multiple investigation groups.
2. Multiple investigations can be under same investigation group.

DraKen0009:
@sainak  I would like to work on this issue.

PR #1912, opened 2024-02-24
converted investigation and investigation group data into JSON

DraKen0009:
## Proposed Changes

- converted the data present populate_investigation.py from string to json

### Associated Issue

- #1541 

## Merge Checklist

- [x] tested using python manage.py populate_investigations


rithviknishad (changes requested):
Keep the data outside of the python code in a JSON file. (Like how we have for ICD11 and Medibase)

DraKen0009:
@rithviknishad  made the required changes , please have a look

DraKen0009:
@sainak  do we need to update the data according to comment before ?
image

Updated to - 
image

DraKen0009:
Also the data at the bottom of the table is not accurate (min, max values)
image

I can update them accordingly and update the json file with it.

sainak:
@DraKen0009 the rows marked yellow needs to be updated, and the ones marked red needs to be deleted

>Also the data at the bottom of the table is not accurate (min, max values)
which rows are you referring to?

DraKen0009:
135-145 ( what i have currently noticed)

- like in line 135 
Min should be 11 and max should be 18
But given 0 and 25

sainak:
> 135-145 ( what i have currently noticed)
> 
>     * like in line 135
>       Min should be 11 and max should be 18
>       But given 0 and 25

I think what you are referring to is ideal value, that is its mentioned in the col just before min, which needs to be added in the model as well, min and max look okay to me

sainak (approved):

Ashesh3 (approved):

vigneshhari (approved):

sainak (changes requested):
TODO:
- make investigation names unique, and add suffix to duplicate values in db
- find and rename duplicates in data/investigations.json
- use upsert to create or update data

sainak (approved):

sainak:
@rithviknishad review required

rithviknishad (approved):

vigneshhari (approved):

PR #1925, opened 2024-02-28
Add validation checks in Resource Request serializer

rash-27:
## Proposed Changes

- Add validation checks in Resource Request serializer

### Associated Issue

- #1918 

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete

### DIFF
--- a/care/facility/api/serializers/resources.py
+++ b/care/facility/api/serializers/resources.py
@@ -144,19 +144,25 @@ class ResourceRequestSerializer(serializers.ModelSerializer):
         if "status" in validated_data:
             validated_data.pop("status")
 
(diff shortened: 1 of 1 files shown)


care/facility/api/serializers/resources.py
sainak, line 1: @rash-27 we have a custon serializer field that takes care of fetching the object
so you can do something like
origin_facility = ExternalIdSerializerField(queryset=Facility.objects.all(),required=True)
and that should eliminate the need for fetching the facility multiple times

rash-27: Ok sorry, I haven't noticed it, I'll change it as soon as possible, Thanks for the review

PR #1926, opened 2024-02-28
added validate on ResetPasswordUserSerializer for username 

itxsoumya:
# Before Update
While we were triggering `http://127.0.0.1:9000/api/v1/password_reset/` for password reset it was returning a status `OK` without verifying if the `username` exists in `User` table or not.
Screenshot from 2024-02-28 21-50-46

# FIX
* I updated the `ResetPasswordUserSerializer` where it validates whether the username exists or not.

# After fix

Screenshot from 2024-02-28 21-57-34

### DIFF
--- a/care/users/reset_password_views.py
+++ b/care/users/reset_password_views.py
@@ -41,6 +41,16 @@ HTTP_IP_ADDRESS_HEADER = getattr(
 class ResetPasswordUserSerializer(serializers.Serializer):
     username = serializers.CharField()
 
+    def validate(self, attrs):
+        username = attrs.get('username')
+
+        try:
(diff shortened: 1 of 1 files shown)


sainak:
hey @itxsoumya we were intentionally not showing the info that the username is incorrect
and the issue is to fix the schema on https://careapi.ohc.network/swagger/

itxsoumya:
@sainak got it. I would be fixing that. Though it is not the fix for that issue But still don't you think this can be also merged for the validation issue.

sainak:
>Though it is not the fix for that issue But still don't you think this can be also merged for the validation issue.

We intentionally removed that validation.

Issue #1817, opened 2024-01-10
Load medibase and icd11 from static json


sainak:
medibase and icd 11 are static data which could be loaded to directly to cache without fetching them from the db

- update the medibase model to use atc_classification key as unique id
- migrate all related models to use this id as fk
- make id(atc_classification) as pk 
- remove external_id
- update the load redis index functions to directly use the json to populate the index

- ~make loading static data run parallel to the migration in the beat container~

cc: @rithviknishad

DraKen0009:
@sainak I would like to work on this issue , please assign it to me

PR #1927, opened 2024-02-28
Loading medibase and icd11 static json

DraKen0009:
## Proposed Changes

1.Updated medibase.json
- removed _id field
- added id field

2.Updated `MedibaseMedicine` model , removed external_id

3.Updated  `python manage.py load_redis_index` command to directly use the json to populate the redis

4.Migrate all related models to use id as fk instead of external_id [ working on it ]

### Associated Issue

- #1817 

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete


DraKen0009:
please let me know if i missed any use of external_id in reference to `MedibaseMedicine` model

care/facility/static_data/icd11.py
```diff
@@ -34,18 +34,15 @@ def get_representation(self) -> ICD11Object:
...
-                chapter=diagnosis[2] or "",
-                has_code=1 if re.match(DISEASE_CODE_PATTERN, diagnosis[1]) else 0,
-            ).save()
+    with open("data/icd11.json", "r") as f:
+        icd_data = json.load(f)
+    for diagnosis in icd_data:
+        ICD11(
+            id=icd11_id_to_int(diagnosis.get("ID")),
+            label=diagnosis.get("label"),
+            chapter=diagnosis.get("meta_chapter_short", ""),
```
sainak, line 43: this is a generated field,
https://github.com/DraKen0009/care/blob/09813dfaf1e05ab76dc833bd135734ca01c0fa45/care/facility/management/commands/load_icd11_diagnoses_data.py#L142-L143

DraKen0009:
- made some changes to load_icd11_diagnose_data file.
- updated migrations field for removing external_id before populating data

sainak:
@DraKen0009 extract the logic from the management command to a common utils file

the migrations will be complex
- clear the migration where we are loading the medicines
- create the unique id fields but dont make them primary key
- add the load_medicine_data migration, make sure the external_id remains consistent
- create new nullable foreign key relations to the id field wherever we are referencing the external_id
- populate the new foreign key fields
- remove the old fk fields 
- rename the new fk fields to the old name so the other code does not break
- make id primary key
- remove external_id

DraKen0009:
contd. #1941

PR #1935, opened 2024-03-01
Replace age field with DOB in users model

rash-27:
## Proposed Changes

- Replace the age field with the date of birth in Users model 

### Associated Issue

-  #7281 in the frontend requires this change 

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete


rithviknishad (changes requested):
Missing custom migration to migrate existing users' age to DOB

rash-27:
Ok I'll update it

care/users/api/serializers/user.py
```diff
@@ -323,6 +332,12 @@ class Meta:
 
     extra_kwargs = {"url": {"lookup_field": "username"}}
 
+    def validate_date_of_birth(self, value):
+        if value and datetime.today().year - value.year <= 0:
+            raise serializers.ValidationError("Enter a valid date of birth.")
```
sainak, line 337:
```suggestion
        if value and now().year - value.year < 16:
            raise serializers.ValidationError("Age must be greater than 15 years")
```

sainak (changes requested):
update dummy data

care/users/models.py
```diff
@@ -235,7 +234,7 @@ class User(AbstractUser):
     video_connect_link = models.URLField(blank=True, null=True)
 
     gender = models.IntegerField(choices=GENDER_CHOICES, blank=False)
-    age = models.IntegerField(validators=[MinValueValidator(1), MaxValueValidator(100)])
+    date_of_birth = models.DateField(null=True, blank=True)
```
sainak, line 237: make this non nullable after migration

care/users/api/serializers/user.py
```diff
@@ -112,6 +111,14 @@ class Meta:
             "user_permissions",
             "created_by",
         )
+    date_of_birth = serializers.DateField(required=False)
```
sainak, line 114:
```suggestion
    date_of_birth = serializers.DateField(required=True)
```

care/users/api/serializers/user.py
```diff
@@ -275,6 +282,8 @@ class UserSerializer(SignUpSerializer):
 
     home_facility = ExternalIdSerializerField(queryset=Facility.objects.all())
 
+    date_of_birth = serializers.DateField(required=False)
```
sainak, line 285:
```suggestion
    date_of_birth = serializers.DateField(required=True)
```

care/users/migrations/0015_age_to_dateofbirth.py
```diff
@@ -0,0 +1,35 @@
...
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="user",
+            name="date_of_birth",
+            field=models.DateField(blank=True, null=True),
+        ),
+
+        migrations.RunPython(age_to_date_of_birth, migrations.RunPython.noop),
```
rithviknishad, line 29: Needs reverse_code to reverse DOB to age

rithviknishad (changes requested):

rithviknishad (approved):

sainak (approved):

vigneshhari:
Does it make sense to use Jan 1 in the migrations, a random DOB seems wrong to me.

rash-27:
@vigneshhari here in the database we store only the age previously in Users model, 
1. Choosing a random value of date of birth is  to save the age of the user as we eliminate the age field after migration 
2. choosing only jan 1 as random dob is because , we can accurately calculate the age if needed , if we choose another date , our age calculated will get inconsistent if the actual date of birth is before the chosen date of birth
Please correct me , if I am wrong

vigneshhari:
okays, we can merge after the lint and tests pass

rash-27:
@vigneshhari I have modified the tests and they are passing and lint errors are also fixed ,once please review it 
image

vigneshhari (approved):

PR #1945, opened 2024-03-05
Fix bed assignment validation for admission

Ashesh3:
Fixes https://github.com/coronasafe/care_fe/issues/7340

This pull request fixes the bed assignment validation for admission. Previously, if a patient did not have a bed assigned and the suggestion was 'A', an error would occur. This PR ensures that the error is raised only when the patient does not have a bed assigned and the suggestion is 'A'.

### DIFF
--- a/care/facility/tests/test_patient_daily_rounds_api.py
+++ b/care/facility/tests/test_patient_daily_rounds_api.py
@@ -23,12 +23,12 @@ class TestDailyRoundApi(TestUtils, APITestCase):
         cls.admission_consultation_no_bed = cls.create_consultation(
             facility=cls.facility,
             patient=cls.patient,
-            suggestion=PatientConsultation.SUGGESTION_CHOICES[1][0],
+            suggestion=PatientConsultation.SUGGESTION_CHOICES[1][0],  # ADMISSION
         )
         cls.domiciliary_consultation_no_bed = cls.create_consultation(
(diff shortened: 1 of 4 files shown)


sainak (approved):

rithviknishad (approved):

khavinshankar (approved):

care/facility/tests/test_patient_daily_rounds_api.py
```diff
@@ -19,8 +20,15 @@ def setUpTestData(cls) -> None:
         cls.patient = cls.create_patient(district=cls.district, facility=cls.facility)
         cls.asset_location = cls.create_asset_location(cls.facility)
         cls.bed = cls.create_bed(facility=cls.facility, location=cls.asset_location)
-        cls.consultation_without_bed = cls.create_consultation(
-            facility=cls.facility, patient=cls.patient
+        cls.admission_consultation_no_bed = cls.create_consultation(
+            facility=cls.facility,
+            patient=cls.patient,
+            suggestion=PatientConsultation.SUGGESTION_CHOICES[1][0],
```
gigincg, line 26: Can you add a comment on what `SUGGESTION_CHOICES[1][0]` is

gigincg (changes requested):

gigincg (approved):

PR #1964, opened 2024-03-11
Adds bulk import support for assets

### LINKED ISSUE #1956: No API Endpoint for Bulk Import of Assets
**Describe the bug**
There is currently an endpoint available for the bulk import of external results at `/api/v1/external_result/bulk_upsert/`. It would be beneficial to have a similar endpoint for asset imports. This enhancement would help avoid the need for multiple requests to the backend.

image

**To Reproduce**
1. Go to assets tab
2. Click on import assets

UdaySagar-Git:
## Proposed Changes

- This PR adds support for the bulk import of assets through a new endpoint. Previously, there was no dedicated endpoint for this purpose.

### Associated Issue

- Fixes #1956 

image

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete

### DIFF
--- a/care/facility/api/viewsets/asset.py
+++ b/care/facility/api/viewsets/asset.py
@@ -412,6 +412,44 @@ class AssetViewSet(
                 status=status.HTTP_500_INTERNAL_SERVER_ERROR,
             )
 
+    @extend_schema(tags=["asset"])
+    @action(methods=["POST"], detail=False)
+    def bulk_upsert(self, request, *args, **kwargs):
+        if "assets" not in request.data:
+            raise ValidationError({"assets": "No Data was provided"})
+        if not isinstance(request.data["assets"], list):
+            raise ValidationError({"assets": "Invalid Data"})
+        if "location" not in request.data:
+            raise ValidationError({"location": "Location is required"})
+
+        assets = request.data.get("assets", [])
+        location = request.data.get("location")
+        errors = []
+        counter = 0
+        ser_objects = []
+        invalid = False
+        for asset in assets:
+            counter += 1
+            asset["location"] = location
+            serialiser_obj = AssetSerializer(data=asset, context={"request": request})
+            valid = serialiser_obj.is_valid()
+            current_error = serialiser_obj.errors
+            if current_error and (not valid):
+                errors.append({"index": counter, "error": current_error})
+                invalid = True
+            ser_objects.append(serialiser_obj)
+
+        if invalid:
+            return Response(errors, status=status.HTTP_400_BAD_REQUEST)
+
+        for ser in ser_objects:
+            ser.save()
+
+        return Response(
+            {"message": f"Successfully created {counter} assets"},
+            status=status.HTTP_201_CREATED,
+        )
+
 
 class AssetTransactionFilter(filters.FilterSet):
     qr_code_id = filters.CharFilter(field_name="asset__qr_code_id")
--- a/care/facility/api/viewsets/asset.py
+++ b/care/facility/api/viewsets/asset.py
@@ -422,9 +422,8 @@ class AssetViewSet(
         if "location" not in request.data:
             raise ValidationError({"location": "Location is required"})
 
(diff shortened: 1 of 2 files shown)


UdaySagar-Git:
@rithviknishad  @sainak 

can you please review this , 
we can set limit from front end , if no limit is provided ,it sets to infinity

care/facility/api/viewsets/asset.py
```diff
@@ -412,6 +412,52 @@ def operate_assets(self, request, *args, **kwargs):
...
+        if "location" not in request.data:
+            raise ValidationError({"location": "Location is required"})
+
+        limit = request.data.get("limit", None)
+        if limit is None:
+            limit = float("inf")
+        if len(request.data["assets"]) > limit:
+            raise ValidationError(
+                {"assets": f"Maximum of {limit} assets can be created at once"}
+            )
```
rithviknishad: I don't see any use case where FE/client would need to specify a limit, in order to limit number of rows added 😅 

Limit needs to be limited in backend. Why would client specify a limit if it can anyways be infinite?

cc: @sainak

vigneshhari:
I don't really see a use case for this endpoint, @gigincg close if this does not make sense.

gigincg:
@vigneshhari It's simply an optimisation for Bulk Import. Currently the FE spams the BE when Bulk Imports happen

care/facility/api/viewsets/asset.py
UdaySagar-Git: @rithviknishad @sainak ? should i hard code the limit for now to 2000? or completely remove the limit as we have in external result bulk upsert

care/facility/api/viewsets/asset.py
```diff
@@ -412,6 +412,51 @@ def operate_assets(self, request, *args, **kwargs):
...
+
+        limit = 200
+
+        if len(request.data["assets"]) > limit:
+            raise ValidationError(
+                {"assets": f"Maximum of {limit} assets can be created at once"}
+            )
+
+        assets = request.data.get("assets", [])
+        location = request.data.get("location")
```
sainak: it would be better to convert these checks to a serializer instead, that will also provide a correct api schema

care/facility/api/viewsets/asset.py
```diff
@@ -412,6 +412,51 @@ def operate_assets(self, request, *args, **kwargs):
...
+            asset["location"] = location
+            serialiser_obj = AssetSerializer(data=asset, context={"request": request})
+            valid = serialiser_obj.is_valid()
+            current_error = serialiser_obj.errors
+            if current_error and (not valid):
+                errors.append({"index": counter, "error": current_error})
+                invalid = True
+            ser_objects.append(serialiser_obj)
+
+        if invalid:
```
sainak:
```suggestion
        if errors:
```

no need for invalid flag

care/facility/api/viewsets/asset.py
```diff
@@ -412,6 +412,51 @@ def operate_assets(self, request, *args, **kwargs):
...
+            return Response(errors, status=status.HTTP_400_BAD_REQUEST)
+
+        for ser in ser_objects:
+            ser.save()
```
sainak, line 445: use bulk create

sainak (changes requested):

care/facility/api/viewsets/asset.py
UdaySagar-Git: getting some issues when using bulk_create

rithviknishad: what issues exactly?

UdaySagar-Git: image
image

UdaySagar-Git: The problem is with the fields 'note'and 'last_serviced_on' fields , 
image
throws an error :
image

how ever this worked 
image
i m not sure which would be a better approch

PR #1991, opened 2024-03-19
Enhance MEWS calculation by fetching data from last 5 log updates

r-nikhilkumar:
Frontend related issue that required changes from backend✅
### Associated Issue

- [Link to issue here, explain how the proposed solution will solve the reported issue/ feature request.](https://github.com/coronasafe/care_fe/issues/7373)

### Architecture changes

- Added a function get_mews_field() at serializer level to filter the latest 5 relevant data field

/swagger docs:
Screenshot (141)


care/facility/api/serializers/patient_consultation.py
```diff
@@ -192,6 +196,34 @@ def validate_bed_number(self, bed_number):
             bed_number = None
         return bed_number
 
+    def get_mews_field(self, consultation):
+        current_time = localtime(now())
+        past_30_minutes = current_time - timedelta(minutes=30)
+        recentDailyArray = DailyRound.objects.filter(rounds_type = DailyRound.RoundsType.NORMAL.value ,consultation = consultation ,modified_date__gte=past_30_minutes, modified_date__lte=current_time).order_by("-modified_date")
```
sainak:
```suggestion
        recentDailyArray = DailyRound.objects.filter(rounds_type = DailyRound.RoundsType.NORMAL.value ,consultation = consultation ,modified_date__gte=past_30_minutes, modified_date__lte=current_time).order_by("-modified_date").only(mews_field_data.keys())
```
reduce the number of fields returned by the qs


```suggestion
        recentDailyArray = DailyRound.objects.filter(rounds_type = DailyRound.RoundsType.NORMAL.value ,consultation = consultation ,modified_date__gte=past_30_minutes, modified_date__lte=current_time).order_by("-modified_date").only(mews_field_data.keys())[:5]
```
and you can directly limit the qs to 5 results, no need to manually keep track of the count

vigneshhari:
Add tests for your change.

care/facility/api/serializers/patient_consultation.py
rithviknishad: The whole issue is about supporting all types of daily rounds and not just normal round type. But you are filtering by Normal rounds type, why?

rithviknishad: Why are we checking against the `modified_date` instead of `created_date`? A daily round logged few days back could be edited for correction and this would cause the MEWS score to be affected which is not intended right?

care/facility/api/serializers/patient_consultation.py
```diff
@@ -192,6 +196,34 @@ def validate_bed_number(self, bed_number):
             bed_number = None
         return bed_number
 
+    def get_mews_field(self, consultation):
```
rithviknishad, line 233: Can we compute the mews score from the backend itself instead of sending the data to FE and making FE to compute it?

rithviknishad (changes requested):

rithviknishad:
@r-nikhilkumar needs tests + lint checks failing

care/facility/api/serializers/patient_consultation.py
r-nikhilkumar: for the filtering five data, these data will not be available in non-normal round types in the time range thats why I'm filtering only normal types because there are only two types of round types

r-nikhilkumar: okay yeah that could create issue, fine I'll make changes for this

care/facility/api/serializers/patient_consultation.py
r-nikhilkumar: yeah I can, but in frontend there is some checks in mews.jsx to flash a hovering dialog that basically which data is missing in that time range, so for that it's looking more informative that mews score is not getting calculated due to absence which parameter, that's why I'm not calculating in backend

r-nikhilkumar:
@rithviknishad , @vigneshhari can you please help me with how to add tests for this

care/facility/api/serializers/patient_consultation.py
r-nikhilkumar: if there are 9 logs created 4 normal, and 5 critical, just hypothetical case recently in 30 min then if i fetch all logs, then I need to traverse in all the 9 files, that would increase time and it couldn't be more optimal, I think...?

rithviknishad: Are you sure this PR solves this issue? Virtual Nursing Assistant creates Automated rounds and not normal rounds

image

rithviknishad:
In the test, you can file a log update with various time intervals using the `measured_at` date time field and verify if API is responding with expected MEWS Scores

care/facility/api/serializers/patient_consultation.py
rithviknishad: Understood. Then in that case can you update the schema to show the proper response type for this field?

care/facility/api/serializers/patient_consultation.py
r-nikhilkumar: Yeah it's that

care/facility/api/serializers/patient_consultation.py
r-nikhilkumar: Which scheme, schema of backend consultation or frontend consultation

care/facility/api/serializers/patient_consultation.py
r-nikhilkumar: Okay I'll not filter I'll erase that filtering parameters of round type

sainak (changes requested):

sainak:
@rithviknishad do we need the mews score for list request?

rithviknishad:
> @rithviknishad do we need the mews score for list request?

Nope. Needed only in ConsultationDetailSerializer

cc: @r-nikhilkumar

care/facility/api/serializers/patient_consultation.py
r-nikhilkumar: we need all logs in last 30 mins if there would be more than 5 logs and one parameter if updated in the 6th one, we wouldn't get it? I'll update the only(mews_field_data.keys())

r-nikhilkumar:
@sainak do I need to make any other changes other than your suggestions?

care/facility/api/serializers/patient_consultation.py
r-nikhilkumar: @rithviknishad please reply

care/facility/api/serializers/patient_consultation.py
r-nikhilkumar: I added this  only(*(list(mews_field_data.keys()))), because only(mews_field_data.keys()) was throwing error,

r-nikhilkumar: done✅

r-nikhilkumar: done✅

r-nikhilkumar: image
I'm using here count variable to check if at any point the all required data is got updated the loop will break without traversing further and will return the value

r-nikhilkumar: please verify, do we need data from last 30 min or only last 5 logs?, I'll change accordingly

r-nikhilkumar:
@rithviknishad 
image
Do I have to add test file for log update here in tests folder with name test_logUpdate_api.py?

rithviknishad:
yes, or you could simply update test daily rounds file

care/facility/api/serializers/patient_consultation.py
rithviknishad: Backend. What is front-end schema?

care/facility/api/serializers/patient_consultation.py
rithviknishad: @sainak the issue states about 5 fields and not the last 5 rows.

care/facility/api/serializers/patient_consultation.py
r-nikhilkumar: image
image
Screenshot 2024-03-23 161423

@rithviknishad, I annotated it, First I thought to make a serialised class for validation and error checking but by default it was getting validated by dailyround, so I just simply annotated like this, is it okay??

r-nikhilkumar:
@rithviknishad , 
Screenshot (119)

I added a test case, here I submitted 3 data files, one with automated because main goal is for automated round type too and rest are non-automated, and its working fine

here is the result
image

care/facility/api/serializers/patient_consultation.py
sainak: okay

care/facility/api/serializers/patient_consultation.py
```diff
@@ -94,6 +96,8 @@ class PatientConsultationSerializer(serializers.ModelSerializer):
         required=False, allow_null=True, allow_blank=True
     )
 
+    mews_field: MewsType = serializers.SerializerMethodField()
```
rithviknishad: Are you sure this is correct? Could you attach the screenshot of the swagger documentation showing the schema of this field in the PR body?

r-nikhilkumar: yes this is correct, I attached in body

rithviknishad: I meant the schema, not the response output.

r-nikhilkumar: how to check, please tell me the steps

r-nikhilkumar: at endpoint /redoc ??

r-nikhilkumar: no this is not showing there how to edit there, I'm not getting where that class SpectacularAPIView, SpectacularRedocView, SpectacularSwaggerView is, I can't find drf_spectacular.views

rithviknishad: Refer example: https://github.com/coronasafe/care/blob/develop/care/facility/api/serializers/asset.py#L129-L140

r-nikhilkumar: @extend_schema_field(
    {
        "type": "object",
        "properties": {
            "hostname": {"type": "string"},
            "source": {"type": "string", "enum": ["asset", "location", "facility"]},
        },
        "nullable": True,
    }
)

so I just need to add annotations like this and no need to create other class like I did earlier in line mewsfield:mewstype

r-nikhilkumar: Screenshot (141)

done

r-nikhilkumar:
@rithviknishad can you approve this also

care/facility/api/serializers/patient_consultation.py
```diff
@@ -196,6 +201,88 @@ def validate_bed_number(self, bed_number):
...
+            "pulse": None,
+            "temperature": None,
+            "consciousness_level": DailyRound.ConsciousnessType.UNKNOWN.name,
+            "modified_date": None,
+        }
+        recentDailyArray = (
+            DailyRound.objects.filter(
+                consultation=consultation,
+                created_date__gte=past_30_minutes,
+                created_date__lte=current_time,
```
rithviknishad: is this filter necessary? we won't have records created in the future anyways...
```suggestion
```

care/facility/api/serializers/patient_consultation.py
```diff
@@ -196,6 +201,88 @@ def validate_bed_number(self, bed_number):
...
+                    mews_field_data[key] = (
+                        None
+                        if newValue == None
+                        else DailyRound.ConsciousnessType(newValue).name
+                        if key == "consciousness_level"
+                        else newValue
+                    )
+                    count += 1
+            if count == 5:
+                break
```
rithviknishad: Code seems to be not very readable / more complicated than necessary... We simply need to find the latest non-null value for specific fields from the filtered queryset right.

You could also maybe simply use subquery and coalesce.

Refer: 

https://docs.djangoproject.com/en/5.0/ref/models/expressions/#subquery-expressions
https://docs.djangoproject.com/en/5.0/ref/models/database-functions/#django.db.models.functions.Coalesce
https://docs.djangoproject.com/en/5.0/ref/models/querysets/#django.db.models.query.QuerySet.annotate

rithviknishad (commented):
@sainak should we consider keeping this as a separate API?

That way we could compute the MEWS score from the backend itself and return that directly if available, and as for missing fields, we could simply return field validation errors instead.

The current approach doesn't seem to be a clean approach IMO. Thoughts?

care/facility/api/serializers/patient_consultation.py
r-nikhilkumar: I made changes regarding its readibility, please verify

r-nikhilkumar:
@rithviknishad, @Ashesh3 so what to do now?

PR #1996, opened 2024-03-20
Soft delete assets from deleted facilities/hospitals

### LINKED ISSUE #1982: Asset list- shows assets from deleted hospitals
**Describe the bug**
Asset list- shows assets from deleted hospitals

### Action Items

- [ ] Cascade soft delete related models of a facility
- [ ] Custom migration to soft delete existing records of deleted facilities

hrit2773:
Fixes #1982 issue, enabled soft delete of assets by overriding delete method in facility model


rithviknishad (changes requested):
No changes to the fields in the model were made. But why are there migrations for it?

hrit2773:
@rithviknishad I was initially using models.cascade but later I realised I need to perform soft delete so I changed it back to original ones so that's why you can see those redundant migrations.But the functionality is working. Should I remove those redundant migrations?

hrit2773:
@rithviknishad I made a second commit i removed those redundant migration history so effectively i just added the delete method in facility model

rithviknishad (changes requested):
Custom migration needed to delete existing such assets of deleted facility.

https://docs.djangoproject.com/en/5.0/ref/migration-operations/#runpython

rithviknishad:
Also test cases needs to be added. It'd be great if you could follow the PR template 😄

hrit2773:
@rithviknishad ya sure I'll make the changes by today

hrit2773:
@rithviknishad I added custom migrations and also added api test case but the tests are not running in my machine because im using windows and im probably getting some connection errors with the test db like some dbus connection errors and no such bucket found. It is giving a failed test for medibase api maybe some db connection issues in windows

care/facility/models/facility.py
```diff
@@ -174,6 +174,12 @@ class Facility(FacilityBaseModel, FacilityPermissionMixin):
     class Meta:
         verbose_name_plural = "Facilities"
 
+    def delete(self, *args, **kwargs):
+        from care.facility.models.asset import Asset, AssetLocation
+
+        Asset.objects.filter(current_location__facility=self).update(deleted=True)
+        super().delete(self, *args, **kwargs)
+
```
sainak: this will skip
https://github.com/coronasafe/care/blob/ce7073e4d789dd19272d257cc79c2500d3d53e75/care/facility/models/asset.py#L175-L179

a better approach would be to make a cron job that regularly checks for deleted objects and recursively soft deletes related objects

sainak (changes requested):

care/facility/models/facility.py
hrit2773: @sainak but if you check recursively it'll slow down and degrade the performance

hrit2773: @sainak and that cron job would run in Linux but not in windows I guess so I can't test it instead what I can do is to delete it in an atomic way maybe

sainak: by cron job, I mean celerybeat cron task

hrit2773: @sainak ok how frequent the check should be every minute or every hour or every 24 hrs?

sainak: let's run it every 24 hour

hrit2773: @sainak okay I will do it

hrit2773:
@sainak plz review the commit...

hrit2773:
@sainak @rithviknishad i added the cron job and should i remove the tests also for soft delete that i created because now that wont work right and maybe the custom migration is also not needed because every midnight the assets will be soft deleted and should i add logging infos and warnings in the task function? plz review it and tell me if any changes required

rithviknishad:
@hrit2773 You can remove the custom migration since it'd be taken care by the celery task itself right.

care/facility/tasks/soft_delete_assets.py
```diff
@@ -0,0 +1,17 @@
...
+from care.facility.models.facility import Facility
+
+broker_url = os.environ.get("CELERY_BROKER_URL")
+app = Celery("soft_delete_assets", broker=broker_url)
+
+
+@app.task
+def soft_delete_assets_schedule():
+    facilities = Facility.objects.all().values_list("name", flat=True)
+    Asset.objects.exclude(current_location__facility__name__in=facilities).update(
```
sainak: you can directly use the id instead of name

hrit2773:
> @hrit2773 You can remove the custom migration since it'd be taken care by the celery task itself right.

@rithviknishad yes I'll remove it and I'll remove the api test case also because I feel that is not needed

care/facility/tasks/soft_delete_assets.py
hrit2773: @sainak yes yes we can I'll do it

hrit2773:
@rithviknishad @sainak made the required changes plz review these changes

care/facility/tasks/soft_delete_assets.py
```diff
@@ -0,0 +1,17 @@
...
+broker_url = os.environ.get("CELERY_BROKER_URL")
+app = Celery("soft_delete_assets", broker=broker_url)
+
+
+@app.task
+def soft_delete_assets_schedule():
+    facilities = Facility.objects.all().values_list("external_id", flat=True)
+    Asset.objects.exclude(
+        current_location__facility__external_id__in=facilities
+    ).update(deleted=True)
```
rithviknishad: We could simply rely on the PK instead of the UUID right? @sainak 

```suggestion
def soft_delete_assets_schedule():
    facilities = Facility.objects.all().values_list("id", flat=True)
    Asset.objects.exclude(
        current_location__facility__id__in=facilities
    ).update(deleted=True)
```

rithviknishad (approved):

care/facility/tasks/soft_delete_assets.py
hrit2773: @rithviknishad we can do that also but as the filterset class also takes the external_id I considered external id just to avoid any existing wrong entry in the database to spoil the functionality otherwise we can do that

sainak: use id as we try to stick with id for internal queries

celerybeat_schedule.py
sainak, line 1: we register tasks in `care/facility/tasks/__init__.py`

hrit2773: @sainak so should i delete soft_delete_assets.py file and write the same function in the __init__.py?

sainak: no delete the celerybeat_schecule.py and revert changes in celery_app.py

hrit2773: @sainak ok i will do that

hrit2773: @sainak should i change the settings config also because how will it identify that the task should run every 24 hours. I have gone through the __init__.py file

hrit2773: I should make changes in the __init__.py also right and in the soft_delete_assets.py i should mark the function using shared task decorator because it should identify that this task should run every 24 hours right?

hrit2773: I will make the changes by today

hrit2773:
@sainak view the changes.. and also replaced external_id with id

rithviknishad (approved):

hrit2773:
@sainak can you review the changes

hrit2773:
@rithviknishad @sainak lets close this issue today

hrit2773:
@sainak @vigneshhari can you review this PR

sainak (approved):

vigneshhari:
Why is this a background job? cant we just soft delete assets if the facility is archived? we cant add assets to a deleted facility right? @sainak

hrit2773:
> Why is this a background job? cant we just soft delete assets if the facility is archived? we cant add assets to a deleted facility right? @sainak

@vigneshhari I did that initially but @sainak asked me to run an cron job

vigneshhari (approved):

PR #2020, opened 2024-03-27
Implement Voice-Enabled Form Filling Feature in Care with Care Scribe

Ashesh3:
## Overview
This PR introduces the Care Scribe feature, a significant enhancement allowing users to fill out forms in Care using voice commands. This innovation aims to streamline data entry processes, enhance accessibility, and improve user experience within the Care platform.

## Related Pull Requests
- Ayushma: https://github.com/coronasafe/ayushma/pull/493
- Care Backend: https://github.com/coronasafe/care/pull/2020
- Care Frontend: https://github.com/coronasafe/care_fe/pull/7481

## Architectural Design
image

## Changes Introduced
- **AI Form Filling**: Implemented the `AIFormFill` model and associated `AIFormFillSerializer` to manage voice-driven form submissions. This includes handling of audio files, transcription, AI-based response generation, and status tracking.
- **File Upload Adjustments**: Enhanced the `FileUpload` model to support a new file type category for AI-related data, ensuring proper handling and permissions for audio file uploads.
- **Endpoints and Routing**: Added new viewsets and routes (`AIFormFillViewSet`) to facilitate the creation, retrieval, update, and listing of AI form fills, secured with appropriate permissions to ensure data protection and privacy.
- **Background Processing**: Integrated Celery tasks (`process_ai_form_fill`) for asynchronous processing of voice to text transcription and AI-driven form data generation, ensuring efficient handling of potentially long-running operations.
- **Migration and Model Adjustments**: Included migrations to update database schemas, adding support for the new functionalities while ensuring backward compatibility and data integrity.
- **Configuration and Settings**: Updated configurations to enable toggling the Care Scribe feature and specify external service credentials, allowing for flexible deployment and integration with external AI services.

## Impact
The introduction of Care Scribe marks a pivotal step towards enhancing the digital infrastructure of healthcare systems with voice-enabled technologies. This feature is expected to:
- Reduce manual data entry errors.
- Increase the efficiency of form-filling processes.
- Provide an inclusive user experience for individuals with disabilities.


rithviknishad (approved):

care/facility/tasks/ai_scribe.py
```diff
@@ -0,0 +1,138 @@
...
+            # Update status to GENERATING_TRANSCRIPT
+            logger.info(f"Generating transcript for AI form fill {form.external_id}")
+            form.status = AIFormFill.Status.GENERATING_TRANSCRIPT
+            form.save()
+
+            if not form.transcript:
+                # Use Ayushma to generate transcript from the audio file
+                transcript = ""
+                for audio_file_url in audio_file_urls:
+                    audio_file_data = requests.request("GET", audio_file_url).content
```
sainak, line 65: we can use the FileUpload().file_contents() instead of making a get request

care/facility/tasks/ai_scribe.py
```diff
@@ -0,0 +1,138 @@
...
+                    "task": "ai_form_fill",
+                    "messages": [
+                        {
+                            "role": "system",
+                            "content": prompt_1,
+                        },
+                        {
+                            "role": "system",
+                            "content": prompt_2.replace(
+                                "{form_schema}", json.dumps(form.form_data, indent=2)
```
sainak, line 104: string formatting would be more appropriate for this case but, we are just adding something at the end we can just append the string right?

Ashesh3: Hey, I've tried to stick to the standard practice of including context into the prompt itself, whereas if there's a system prompt in a separate variable, far apart from the actual use, then we try to include any additional context inside the prompt itself, so if someone is glancing over this "prompt template" they know that it requires supply of an additional variable without digging through all the code. We use the same practice in ayushma where even though references are always added to the end of the prompt, we still require the template variable "{references}" in the prompt itself and even langchain enforces such practice. For the 3rd and 4th prompt, i have not used template variable since the entire prompt and appending variable happens at the same place, unlike prompt 2 where the prompt is away at the top, separate from it's actual use. Let me know if this makes sense or if you still feel string formatting would be better!

care/facility/tasks/ai_scribe.py
sainak: can be fixed in the next iteration

sainak (approved):

sainak:
closing this as we decided to implement scribe as a plugin 
https://github.com/coronasafe/care_scribe/pull/4

Issue #1336, opened 2023-06-02
Fix N+1 queries in `/api/v1/resource/`


sainak:
Sentry Issue: [CARE-M5](https://coronasafe-network.sentry.io/issues/4134976066/?referrer=github_integration)
|  |  |
| ------------- | --------------- |
| **Transaction Name** | /api/v1/resource/ |
| **Parent Span** | django.view - resource-request-list |
| **Repeating Spans (0)** |db - SELECT COUNT(*) AS __count FROM facility_patientregistration WHERE (facility_patientregistration.deleted = %s AND facility_patientregistration.facility_id = %s AND facility_patientregistration.is_active = %s) |
||db - SELECT COUNT(*) AS __count FROM facility_bed WHERE (facility_bed.deleted = %s AND facility_bed.facility_id = %s)|

dhruv-goyal-10:
@sainak @rithviknishad @vigneshhari Can you please assign this issue to me?

adityasuthar:
Hi @sainak can I work on this issue

adityasuthar:
> Parent Span | django.view - resource-request-list

I could not find this Django view in the current version of `develop` branch. It has been some time since this issue was raised and I think the view class name changed in the meantime. Can you help me finding the view and the query which we need to optimise @sainak @rithviknishad

adityasuthar:
I found the view `ResourceRequestViewSet`, working on it.

adityasuthar:
Hi @sainak @rithviknishad I have raised a pull request at https://github.com/coronasafe/care/pull/2035

PR #2035, opened 2024-03-29
fix N+1 queries in /api/v1/resource/

adityasuthar:
## Proposed Changes

Fix the N+1 queries in the resources/ API 

### Associated Issue

#1336 

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete

### DIFF
--- a/care/facility/api/serializers/facility.py
+++ b/care/facility/api/serializers/facility.py
@@ -49,10 +49,7 @@ class FacilityBasicInfoSerializer(serializers.ModelSerializer):
     read_cover_image_url = serializers.CharField(read_only=True)
     features = serializers.MultipleChoiceField(choices=FEATURE_CHOICES)
     patient_count = serializers.SerializerMethodField()
-    bed_count = serializers.SerializerMethodField()
 
-    def get_bed_count(self, facility):
-        return Bed.objects.filter(facility=facility).count()
 
     def get_patient_count(self, facility):
         return PatientRegistration.objects.filter(
@@ -81,7 +78,6 @@ class FacilityBasicInfoSerializer(serializers.ModelSerializer):
             "read_cover_image_url",
             "features",
             "patient_count",
-            "bed_count",
         )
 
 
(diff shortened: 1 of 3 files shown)


vigneshhari:
Can you add proof as well, the queries executed before vs after ?

care/facility/models/facility.py
sainak, line 1: these will still cause n+1 db queries
use queryset.annotate instead

adityasuthar:
Hi, @vigneshhari @sainak I tried to read the DRF docs for serializers and ORM, tried various solutions and finally was able to do it for `bed_count`, will need your review if this is what we wanted

but I got stuck again when doing the same for `patient_count` because the query started to give incorrect results when I also annotated the `patientregistration` table.

ORM started to join the bed table with `patientregistration` and fetch extra results
maybe we can achieve this by using raw SQL query 

Here is the query I tried to replicate the same process for patient_count also


```
    queryset = (
        ResourceRequest.objects.all()
        .select_related(
            "origin_facility",
            "origin_facility__ward",
            "origin_facility__local_body",
            "origin_facility__district",
            "origin_facility__state",
            "approving_facility",
            "approving_facility__ward",
            "approving_facility__local_body",
            "approving_facility__district",
            "approving_facility__state",
            "assigned_facility",
            "assigned_facility__ward",
            "assigned_facility__local_body",
            "assigned_facility__district",
            "assigned_facility__state",
            "assigned_to",
            "created_by",
            "last_edited_by",
        )
        .annotate(
            origin_facility_bed_count=Count("origin_facility__bed"),
            approving_facility_bed_count=Count("approving_facility__bed"),
            assigned_facility_bed_count=Count("assigned_facility__bed"),
        )
        .annotate(
            origin_facility_patient_count=Count("origin_facility__patientregistration"),
            approving_facility_patient_count=Count(
                "approving_facility__patientregistration"
            ),
            assigned_facility_patient_count=Count(
                "assigned_facility__patientregistration"
            ),
        )
    )
```

vigneshhari:
The implementation for counts is to be switched to redis, closing this MR as a part of that effort.

Issue #1697, opened 2023-11-06
Add created_by and updated_by fields to base model


sainak:
Add created_by and updated_by fields to base model to track user that made changes to the objects

AnkurPrabhu:
Hey can i work on this one ?

sainak:
sure

AnkurPrabhu:
@sainak just wanted to confirm this will have 
1) a migration file for the addition of two fields in class BaseModel 
2) all the changes where any object of a class extended by BaseModel is either created or modified right ?

sainak:
> @sainak just wanted to confirm this will have
> 
>     1. a migration file for the addition of two fields in class BaseModel
> 
>     2. all the changes where any object of a class extended by BaseModel is either created or modified right ?

Yes, and also override perform create/update function of viewsets

AnkurPrabhu:
@sainak will the migrations files be a separate pr ?

AnkurPrabhu:
@sainak can you pls take a look at the initial pr just wanted to confirm if im in the right direction or not.
made the migration field and updated the created_by and updated_by fields for some viewsets, once you approve the approach will do it for all the other viewsets

AnkurPrabhu:
@sainak  i have updated the pr with all the necessary changes pls take a look

Harsh-D-2004:
is this still open??

PR #2036, opened 2024-03-30
Add Plugin Support

### LINKED ISSUE #2031: Setup a backend for plugins
add a basic plugin system like [rails-engines](https://guides.rubyonrails.org/engines.html) to create a pluggable architecture for apps like hcx, abdm and scribe

vigneshhari:
## Proposed Changes

- Brief of changes made.

### Associated Issue

- closes #2031

ref: https://github.com/coronasafe/docs/pull/18

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete


care/facility/migrations/0424_alter_fileupload_file_category_and_more.py
```diff
@@ -0,0 +1,43 @@
+# Generated by Django 4.2.10 on 2024-04-02 11:49
+
+from django.db import migrations, models
```
vigneshhari: Why is this migration created even thou we had no model change?

care/facility/models/file_upload.py
```diff
@@ -1,79 +1,53 @@
...
-    )
+    file_type = models.IntegerField(default=0)
     file_category = models.CharField(
-        choices=FileCategoryChoices,
-        default=FileCategory.UNSPECIFIED.value,
+        choices=FileCategory.choices,
+        default=FileCategory.UNSPECIFIED,
         max_length=100,
     )
+    created_date = models.DateTimeField(
```
vigneshhari, line 34: inherit from facility base model.

care/facility/migrations/0424_alter_fileupload_file_category_and_more.py
sainak: the choice enums were converted to the django choice types. that's why the new migration, we are not changing the types

care/facility/models/file_upload.py
sainak: Inheriting the basefileupload from base model might add the new user dependent fields https://github.com/coronasafe/care/issues/1697
that's why I avoided inheriting from it

care/facility/migrations/0424_alter_fileupload_file_category_and_more.py
vigneshhari: The case was changed, that caused the migration, can you check if we are okay with the new case, there are no reverse lookups.

care/facility/models/file_upload.py
vigneshhari: Okays

plug_config.py
```diff
@@ -0,0 +1,21 @@
+import environ
+
+from plugs.manager import PlugManager
+from plugs.plug import Plug
+
+env = environ.Env()
+
+scribe_plug = Plug(
+    name="care_scribe",
```
vigneshhari: Lets remove this and get this MR Merged.

plug_config.py
```diff
@@ -0,0 +1,21 @@
...
+    package_name="git+https://github.com/coronasafe/care_scribe.git",
+    version="@scribe",
+    configs={
+        "TRANSCRIBE_SERVICE_PROVIDER_API_KEY": env(
```
vigneshhari: If its available from the env, we can fetch it from the env  in the plugin, no need to redefine, just make sure we document this somewhere

PR #2055, opened 2024-04-04
Occupation dropdown backend

hrit2773:
## Proposed Changes

- Changed the enum class


rithviknishad (changes requested):
Missing migrations. Run `python manage.py makemigrations`

Makefile
```diff
@@ -1,7 +1,7 @@
 .PHONY: build, re-build, up, down, list, logs, test, makemigrations
 
 
-DOCKER_VERSION := $(shell docker --version 2>/dev/null)
+DOCKER_VERSION := docker --version
```
rithviknishad: Why was this changed?

rithviknishad (changes requested):

hrit2773:
@rithviknishad I changed this in my windows machine just to run the command and pushed it by mistake I changed it back and i guess its fine now added migration also

rithviknishad (approved):

care/facility/migrations/0426_alter_patientmetainfo_occupation.py
```diff
@@ -0,0 +1,54 @@
...
+                    (2, "BUSINESSMAN"),
+                    (3, "HEALTH_CARE_WORKER"),
+                    (4, "HEALTH_CARE_LAB_WORKER"),
+                    (5, "ANIMAL_HANDLER"),
+                    (6, "HEALTHCARE_PRACTITIONER"),
+                    (7, "PARADEMICS"),
+                    (8, "Business_related"),
+                    (9, "ENGINEER"),
+                    (10, "TEACHER"),
+                    (11, "Other_Professions"),
```
rithviknishad: Do we need both "Other_Pressions" and "OTHERS"

rithviknishad (changes requested):

care/facility/migrations/0426_alter_patientmetainfo_occupation.py
hrit2773: @rithviknishad maybe we can consider other skilled professions that was mentioned in the dropdown suggstions list so i added

hrit2773: @rithviknishad in the frontend Other professional occupations were mentioned and professional occupation is a separate type and the corresponding value is OTHER_PROFESSIONS not to be confused with OTHERS and its not the same. This is needed.

rithviknishad: Rename it in backend too. It could lead to confusions.

Also change to use uppercase for all

hrit2773:
@rithviknishad I capitalized the names and also capitalized in the frontend

sainak:
@hrit2773 delete the migrations and rerun makemigrations

hrit2773:
@sainak @rithviknishad plz review this pr

care/facility/models/patient.py
```diff
@@ -570,7 +570,33 @@ class OccupationEnum(enum.Enum):
...
+        CONSTRUCTION_EXTRACTION = 23
+        AGRI_NATURAL = 24
+        PRODUCTION_OCCUPATION = 25
+        PILOT_FLIGHT = 26
+        VEHICLE_DRIVER = 27
+        MILITARY = 28
+        HOMEMAKER = 29
+        OTHERS = 30
+        UNKNOWN = 31
+        NOT_APPLICABLE = 32
```
rithviknishad: `OTHERS` was previously `6` and is now `30`. And `6` is now Healthcare Practitioner. This would mean, all previously entered patients with occupation as "Other" before this PR would become a Healthcare Practitioner once this is merged.

rithviknishad (changes requested):

hrit2773:
@rithviknishad yes you are right OTHERS would be misread as HEALTHCARE PRACTIOTIONER so changed it. And i made changes in the frontend also

rithviknishad (approved):

nihal467:
@hrit2773 there is migration issue, fix it

hrit2773:
@nihal467 but it's working fine when I'm running it the occupation is being added and displayed properly. Should I re-run the migrations?

hrit2773:
@nihal467 can you check it now maybe there were some conflicting migrations before

sainak (approved):

vigneshhari (approved):

PR #2095, opened 2024-04-16
Encounter Symptoms

hrit2773:


hrit2773:
@rithviknishad I have added a JSON field symptoms_with_dates where each object stores date and list of symptoms. And also overridden the to_representation() method of patient consultation serializers to sort the symptoms timeline based on date(viewing in chronological order as mentioned in the issue). Can you plz confirm these changes.

care/facility/models/patient_consultation.py
```diff
@@ -26,9 +27,24 @@
...
+    symptom = models.CharField(choices=NEW_SYMPTOM_CHOICES, blank=False)
+    other_symptom = models.CharField(default="", blank=True)
+    onset_date = models.DateField(null=True, blank=True)
+    cure_date = models.DateField(null=True, blank=True)
+    consultation = models.ForeignKey(
+        "PatientConsultation",
+        null=True,
+        blank=True,
+        on_delete=models.CASCADE,
+        related_name="consultation_symptoms",
```
rithviknishad:
```suggestion
        related_name="symptoms",
```

care/facility/api/serializers/patient_consultation.py
```diff
@@ -57,6 +61,18 @@
 MIN_ENCOUNTER_DATE = make_aware(settings.MIN_ENCOUNTER_DATE)
 
 
+class ConsultationSymptomsSerializers(serializers.ModelSerializer):
+    class Meta:
+        model = ConsultationSymptoms
+        fields = ("symptom", "other_symptom", "onset_date", "cure_date", "consultation")
```
rithviknishad: see how other serializers are done. Let's serialize all fields, including the id (external_id), created_by, modified_by, modified and created date, except `deleted`.

care/facility/api/serializers/patient_consultation.py
```diff
@@ -151,7 +167,10 @@ class PatientConsultationSerializer(serializers.ModelSerializer):
         help_text="Bulk create diagnoses for the consultation upon creation",
     )
     diagnoses = ConsultationDiagnosisSerializer(many=True, read_only=True)
-
+    create_symptoms_timeline = ConsultationCreateSymptomsSerializers(
+        many=True, write_only=True, required=False
+    )
+    symptoms_timeline = ConsultationSymptomsSerializers(many=True, read_only=True)
```
rithviknishad:
```suggestion
    symptoms = ConsultationSymptomsSerializers(many=True, read_only=True)
```

care/facility/api/serializers/patient_consultation.py
```diff
@@ -174,6 +193,7 @@ def get_discharge_prn_prescription(self, consultation):
     class Meta:
         model = PatientConsultation
         read_only_fields = TIMESTAMP_FIELDS + (
+            "symptoms_with_dates",
```
rithviknishad:
```suggestion
```

rithviknishad (changes requested):

hrit2773:
@rithviknishad ok thats fine but the issue is django.db.utils.ProgrammingError: column "symptom" of relation "facility_consultationsymptoms" does not exist
LINE 1: ..._id", "created_date", "modified_date", "deleted", "symptom",... bulk create is actually unable to identify symptom I guess

care/facility/api/serializers/patient_consultation.py
hrit2773: @rithviknishad fine but first the symptoms should get created then we can think of it. I have made migrations and migrated as well but still maybe while passing the attributes in the ConsultationSymptoms() object in the bulk create its unable to identify symptom field from the model

care/facility/models/patient_consultation.py
```diff
@@ -26,9 +27,24 @@
     reverse_choices,
 )
 from care.users.models import User
+from care.utils.models.base import BaseModel
 from care.utils.models.validators import JSONFieldSchemaValidator
 
 
+class ConsultationSymptoms(BaseModel, ConsultationRelatedPermissionMixin):
```
sainak:
```suggestion
class ConsultationSymptom(BaseModel, ConsultationRelatedPermissionMixin):
```

Model name should be singular

care/facility/models/patient_base.py
```diff
@@ -56,6 +56,8 @@ def reverse_choices(choices):
     (34, "NEW LOSS OF SMELL"),
 ]
 
+NEW_SYMPTOM_CHOICES = [(tup[1], tup[1].lower()) for tup in SYMPTOM_CHOICES]
```
sainak: Create a new integer choice enum `Symptom` and use that instead of this
https://docs.djangoproject.com/en/5.0/ref/models/fields/#enumeration-types

rithviknishad:
@sainak For the migration of existing data to the new table, as discussed the following is to be done:

- [x] **All symptoms collected on the consultation table should be present in the new table.** The onset date for each new record would be the same as the `symptoms_onset_date` present in the consultation record.
- [x] All symptoms collected from **ALL of the related daily rounds** also needs to be added to the new table **IF that symptom was NOT already recorded in the related consultation table**. The onset date for such records would be the `created_date` of the daily round.

care/facility/api/serializers/consultation_symptom.py
```diff
@@ -0,0 +1,125 @@
...
+        if (
+            ConsultationSymptom.objects.filter(
+                consultation=consultation,
+                symptom=validated_data.get("symptom"),
+                other_symptom=validated_data.get("other_symptom") or "",
+                cure_date__isnull=True,
+            )
+            .exclude(
+                clinical_impression_status=ClinicalImpressionStatus.ENTERED_IN_ERROR
+            )
```
rithviknishad: We could simply filter out by IN_PROGRESS status right?

```suggestion
            ConsultationSymptom.objects.filter(
                consultation=consultation,
                symptom=validated_data.get("symptom"),
                other_symptom=validated_data.get("other_symptom") or "",
                clinical_impression_status=ClinicalImpressionStatus.IN_PROGRESS,
            )
```

rithviknishad (changes requested):

care/facility/api/serializers/patient_consultation.py
```diff
@@ -547,14 +552,20 @@ def validate_create_symptoms(self, value):
...
-                )
+
+            if "cure_date" in obj:
+                if obj["cure_date"] < obj["onset_date"]:
+                    raise ValidationError(
+                        {"cure_date": "Cure date should be after onset date"}
+                    )
+                obj["clinical_impression_status"] = ClinicalImpressionStatus.COMPLETED
+            else:
+                obj["clinical_impression_status"] = ClinicalImpressionStatus.IN_PROGRESS
```
rithviknishad, line 568: "cure_date" may or may not be sent from client as `null` in the obj.
This would cause incorrect results right?

image

rithviknishad (changes requested):

rithviknishad (changes requested):
`created_by` and `updated_by` fields are not set from the viewset

care/facility/utils/reports/discharge_summary.py
```diff
@@ -97,6 +98,7 @@ def get_discharge_summary_data(consultation: PatientConsultation):
     )
     hcx = Policy.objects.filter(patient=consultation.patient)
     daily_rounds = DailyRound.objects.filter(consultation=consultation)
+    symptoms = ConsultationSymptom.objects.filter(consultation=consultation)
```
rithviknishad: let's exclude symptoms enterred in error from discharge summary?

rithviknishad (changes requested):

sainak:
Closing in favor of #2186

PR #2096, opened 2024-04-16
upgrade user skills

rash-27:
## Proposed Changes

- Updated the skills of users list

### Associated Issue

- fixes issue https://github.com/coronasafe/care_fe/issues/7616
## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete


sainak (changes requested):
@rash-27 you also need to create a custom migration that will add the newly added skills to db if not present, add a condition to skip if the skill table is empty (this would be the case with new instance, so users are expected to load the skill using the load skill command)

rash-27:
Ok will add it

rash-27:
@sainak is this the required migration , or am I missing something?

sainak (approved):

rithviknishad (approved):

vigneshhari (approved):

PR #2102, opened 2024-04-19
Increase area of specialisation in doctors

rash-27:
## Proposed Changes

- Updated list of specialisation in doctors  

### Associated Issue

- https://github.com/coronasafe/care_fe/issues/7612

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete


rithviknishad (changes requested):
Since this is now same as the records in Skills table, we should use foreign key relation to the skill table to track the count.

gigincg (approved):

vigneshhari (approved):

vigneshhari:
> Since this is now same as the records in Skills table, we should use foreign key relation to the skill table to track the count.

Move this as a seperate issue, let this go out as is.

PR #2114, opened 2024-04-24
Prevent destination facility from marking shift as complete

Ashesh3:
Related issue: https://github.com/coronasafe/care_fe/issues/7684
Frontend change: https://github.com/coronasafe/care_fe/pull/7697

Adds a permission check to disallow destination facility staff from marking the shifting request as complete, while still allowing state and district admin to do the same. Also auto discharge is removed for completed stage, and the UI handles the discharge + shifting change together.


care/facility/api/serializers/shifting.py
```diff
@@ -324,7 +324,18 @@ def update(self, instance, validated_data):
             "status" in validated_data
             and validated_data["status"] == REVERSE_SHIFTING_STATUS_CHOICES["COMPLETED"]
         ):
-            discharge_patient(instance.patient)
+            # Staff from the destination must not be able to mark the shift as complete.
+            destination_facility = (
+                validated_data["assigned_facility"] or instance.assigned_facility
+            )
+            if destination_facility and user in destination_facility.users.all():
```
sainak:
```suggestion
            if destination_facility and destination_facility.users.contains(user):
```

care/facility/api/serializers/shifting.py
```diff
@@ -324,7 +324,18 @@ def update(self, instance, validated_data):
             "status" in validated_data
             and validated_data["status"] == REVERSE_SHIFTING_STATUS_CHOICES["COMPLETED"]
         ):
-            discharge_patient(instance.patient)
+            # Staff from the destination must not be able to mark the shift as complete.
+            destination_facility = (
+                validated_data["assigned_facility"] or instance.assigned_facility
```
sainak:
```suggestion
                validated_data.get("assigned_facility") or instance.assigned_facility
```

sainak (approved):

care/facility/api/serializers/shifting.py
```diff
@@ -324,7 +324,18 @@ def update(self, instance, validated_data):
             "status" in validated_data
             and validated_data["status"] == REVERSE_SHIFTING_STATUS_CHOICES["COMPLETED"]
         ):
-            discharge_patient(instance.patient)
+            # Staff from the destination must not be able to mark the shift as complete.
+            destination_facility = (
+                validated_data.get("assigned_facility") or instance.assigned_facility
+            )
+            if destination_facility and destination_facility.users.contains(user):
```
vigneshhari: if a user is part of the destination and receiving facility this might become a problem.

vigneshhari (changes requested):

sainak (approved):

rithviknishad (approved):

nihal467:
the related frontend PR is having bug, don't merge it

nihal467:
LGTM @vigneshhari you can review it, the PR is working

care/facility/api/serializers/shifting.py
```diff
@@ -323,8 +323,15 @@ def update(self, instance, validated_data):
         if (
             "status" in validated_data
             and validated_data["status"] == REVERSE_SHIFTING_STATUS_CHOICES["COMPLETED"]
+            and not has_facility_permission(user, instance.origin_facility)
         ):
-            discharge_patient(instance.patient)
```
vigneshhari, line 327: Why are we deleting this statement?

vigneshhari (changes requested):

care/facility/api/serializers/shifting.py
khavinshankar: Previously, We automatically discharged the patient when the shifting reason was completed but now once the shifting is done (this API), we show the discharge form in the frontend so that user can enter details like discharge reason, etc before discharging a patient.

vigneshhari (approved):

PR #2116, opened 2024-04-24
Disable the Bed List of Already occupied Beds (back-end change)

### LINKED ISSUE #7569: Disable the Bed List of Already occupied Beds
Request to Disable the dropdown for already occupied Bed Number in assets-Configurations of HL7 vital Monitors.
Since it creates confusions by showing all the beds on Selecting the beds for the particular HL7 Vital Monitors instead to show only available-Non occupied Beds.
Screenshot 2024-04-08 170257

AnkurPrabhu:
## Proposed Changes

- Brief of changes made.
     - added a new filter not_occupied_by_hl7_monitor when true filter bed already occupied by hl7 monitors
### Associated Issue
 Fixes [link](https://github.com/coronasafe/care_fe/issues/7569)

### Architecture changes

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [x] Linting Complete


rithviknishad (changes requested):
LGTM, but there is no need to make the filter specific to hl7 monitor right? We can make a filter `not_occupied_by_asset_type` which accepts the asset type to exclude.

This way it'd work for excluding ONVIF cameras or ventilators too if needed in the future.

AnkurPrabhu:
> LGTM, but there is no need to make the filter specific to hl7 monitor right? We can make a filter `not_occupied_by_asset_type` which accepts the asset type to exclude.
> 
> This way it'd work for excluding ONVIF cameras or ventilators too if needed in the future.

yes makes sense and is also scalable

care/facility/api/viewsets/bed.py
```diff
@@ -38,6 +39,21 @@ class BedFilter(filters.FilterSet):
...
+    def filter_bed_is_not_occupied_by_asset_type(self, queryset, name, value):
+        if value == AssetClasses.HL7MONITOR.name:
+            return queryset.exclude(
+                id__in=Subquery(
+                    AssetBed.objects.filter(
+                        bed__id=OuterRef("id"),
+                        asset__asset_class=AssetClasses.HL7MONITOR.name,
+                    ).values("bed__id")
+                )
+            )
```
rithviknishad: This is still specific to HL7MONITOR and won't work for other asset types right?
Instead we could simply do something like this right?:

```suggestion
        if value:
            return queryset.exclude(
                id__in=Subquery(
                    AssetBed.objects.filter(
                        bed__id=OuterRef("id"),
                        asset__asset_class=value,
                    ).values("bed__id")
                )
            )
```

rithviknishad (changes requested):

care/facility/api/viewsets/bed.py
AnkurPrabhu: oh yess my bad

rithviknishad (approved):

care/facility/api/viewsets/bed.py
```diff
@@ -38,6 +38,21 @@ class BedFilter(filters.FilterSet):
...
+    def filter_bed_is_not_occupied_by_asset_type(self, queryset, name, value):
+        if value:
+            return queryset.exclude(
+                id__in=Subquery(
+                    AssetBed.objects.filter(
+                        bed__id=OuterRef("id"),
+                        asset__asset_class=value,
+                    ).values("bed__id")
+                )
+            )
```
sainak:
```suggestion
            return queryset.filter(
                ~Exists(
                    AssetBed.objects.filter(
                        bed__id=OuterRef("id"),
                        asset__asset_class=value,
                    )
                )
            )
```

for this usecase Exists results in more optimized query

sainak (changes requested):
add tests

rithviknishad (commented):
LGTM, tests needed

AnkurPrabhu:
@rithviknishad i have added a test

care/facility/tests/test_bed_api.py
```diff
@@ -36,3 +38,28 @@ def test_list_beds(self):
...
+        # 4 beds 1 linked with HL7MONITOR and 3 created in setup
+
+        response = self.client.get("/api/v1/bed/")
+
+        # Assert list returns 4 beds
+        self.assertEqual(response.json()["count"], 4)
+
+        response_with_not_occupied_bed = self.client.get(
+            "/api/v1/bed/?not_occupied_by_asset_type=HL7MONITOR"
+        )
```
rithviknishad:
```suggestion
        response_with_not_occupied_bed = self.client.get(
            "/api/v1/bed/",
            { "not_occupied_by_asset_type": "HL7MONITOR" },
        )
```

rithviknishad (approved):

sainak (approved):

nihal467:
LGTM @rithviknishad @sainak can you fix the lint issue

PR #2134, opened 2024-05-07
Skip creating event group if all fields are null

### LINKED ISSUE #2001: Events are being created for empty groups
event entries are being created for groups with no values

https://care.coronasafe.in/facility/0c95c7f0-e1d2-4aff-83fa-933cef60d3a8/patient/7e8fc20f-02da-47f7-8c33-f2d8c7e52dd8/consultation/edfe0bd6-0f50-4a4c-8cd0-2d4ad0e9cea1

sainak:
## Proposed Changes

- check and skip if all fields of the group are null in an event group

### Associated Issue

- fixes #2001

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete

### DIFF
--- a/care/facility/events/handler.py
+++ b/care/facility/events/handler.py
@@ -58,6 +58,15 @@ def create_consultation_event_entry(
     ).values_list("id", "fields")
     for group_id, group_fields in groups:
         if set(group_fields) & fields_to_store:
+            value = {}
+            for field in group_fields:
+                try:
+                    value[field] = data[field]
+                except KeyError:
+                    value[field] = getattr(object_instance, field, None)
+            # if all values in the group are Falsy, skip creating the event for this group
+            if all(not v for v in value.values()):
+                continue
             PatientConsultationEvent.objects.select_for_update().filter(
                 consultation_id=consultation_id,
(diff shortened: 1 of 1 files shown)


care/facility/events/handler.py
```diff
@@ -58,6 +58,15 @@ def create_consultation_event_entry(
...
     for group_id, group_fields in groups:
         if set(group_fields) & fields_to_store:
+            value = {}
+            for field in group_fields:
+                try:
+                    value[field] = data[field]
+                except KeyError:
+                    value[field] = getattr(object_instance, field, None)
+            # if all values in the group are Falsy, skip creating the event for this group
+            if all(not v for v in value.values()):
```
rithviknishad, line 68: Should we consider making the check against None instead?
What if any of the vital is 0 (falsy)?

rithviknishad (approved):

care/facility/events/handler.py
sainak: yeah makes sense, will update it

vigneshhari (approved):

PR #2160, opened 2024-05-13
Reply feature for Doctor notes 

UdaySagar-Git:
### Associated Issue

Frontend PR: https://github.com/coronasafe/care_fe/pull/7805

### Architecture changes

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete


care/facility/api/serializers/patient.py
```diff
@@ -539,6 +546,11 @@ def create(self, validated_data):
             # If the user is not a doctor then the user type is the same as the user type
             validated_data["user_type"] = user_type
 
+        if validated_data.get("reply_to"):
+            validated_data["reply_to"] = PatientNotes.objects.get(
+                external_id=validated_data["reply_to"]
+            )
+
```
shivankacker: Add conditions in `get` to check if the `reply_to` belongs to the same thread and consultation

shivankacker (changes requested):

care/facility/api/serializers/patient.py
UdaySagar-Git: added

shivankacker (approved):

nihal467:
@UdaySagar-Git fix the lint and test in the PR

UdaySagar-Git:
@nihal467 fixed !

care/facility/api/serializers/patient.py
```diff
@@ -518,6 +518,13 @@ class PatientNotesSerializer(serializers.ModelSerializer):
     thread = serializers.ChoiceField(
         choices=PatientNoteThreadChoices, required=False, allow_null=False
     )
+    reply_to = serializers.UUIDField(required=False, allow_null=True)
+
+    def to_representation(self, instance):
+        ret = super().to_representation(instance)
+        if ret.get("reply_to"):
+            ret["reply_to"] = instance.reply_to.external_id
+        return ret
```
sainak: use externalidrelatedfield

sainak (changes requested):

sainak:
@UdaySagar-Git how will you ensure that the notes that you fetch that are replies include the parent in response?, also what if the notes have replies, but it were not fetched in the current response

cc @rithviknishad

UdaySagar-Git:
@sainak , thanks for pointing that out. It won't fetch the parent if there are more notes. Instead of sending just the ID, sending the complete parent object along with the note would solve this. Should I consider that and make changes in frontend according to it?

rithviknishad:
@UdaySagar-Git we could add a `ReplyToPatientNoteSerializer` containing a lesser number of fields, just the required ones.

care/facility/api/serializers/patient.py
```diff
@@ -518,6 +532,17 @@ class PatientNotesSerializer(serializers.ModelSerializer):
...
+        queryset=PatientNotes.objects.all(),
+        required=False,
+        allow_null=True,
+    )
+
+    def to_representation(self, instance):
+        ret = super().to_representation(instance)
+        if instance.reply_to:
+            ret["reply_to"] = ReplyToPatientNoteSerializer(instance.reply_to).data
+        return ret
```
rithviknishad: We could simply do this instead right?


```suggestion
    reply_to_object = ReplyToPatientNoteSerializer(source="reply_to", read_only=True)
```

rithviknishad (changes requested):
Other than this, LGTM. 

Tests would be appreciated 😄

rithviknishad (approved):
Tests needed

UdaySagar-Git:
@rithviknishad forgot to commit 😅 , 
just added

rithviknishad (approved):

care/facility/api/serializers/patient.py
```diff
@@ -503,6 +503,20 @@ class Meta:
         exclude = ("patient_note",)
 
 
+class ReplyToPatientNoteSerializer(serializers.ModelSerializer):
+    created_by_object = UserBaseMinimumSerializer(source="created_by", read_only=True)
```
sainak, line 505:
```suggestion
    id = serializers.CharField(source="external_id", read_only=True)
    created_by_object = UserBaseMinimumSerializer(source="created_by", read_only=True)
```

the `id` in field would return the db `id`, we should avoid exposing it

sainak (changes requested):
add `select_related("reply_to")` in the notes viewset

nihal467:
LGTM

care/facility/api/viewsets/patient.py
```diff
@@ -910,7 +910,7 @@ class PatientNotesViewSet(
 ):
     queryset = (
         PatientNotes.objects.all()
-        .select_related("facility", "patient", "created_by")
+        .select_related("facility", "patient", "created_by", "reply_to")
```
```suggestion
        .select_related("facility", "patient", "created_by", "reply_to", "reply_to__created_by")
```

sainak (approved):

care/facility/api/serializers/patient.py
```diff
@@ -539,6 +561,17 @@ def create(self, validated_data):
             # If the user is not a doctor then the user type is the same as the user type
             validated_data["user_type"] = user_type
 
+        if validated_data.get("reply_to"):
+            reply_to_note = validated_data["reply_to"]
+            if reply_to_note.thread != validated_data["thread"]:
+                raise serializers.ValidationError(
+                    "Reply to note should be in the same thread"
+                )
+            if reply_to_note.consultation != validated_data.get("consultation"):
```
vigneshhari, line 571: consultation is not a required field, need better validations

nihal467:
@UdaySagar-Git can you make the changes requested

care/facility/api/serializers/patient.py
UdaySagar-Git: Could you please suggest what additional validations can be implemented?

nihal467:
@UdaySagar-Git what is the update on the PR

UdaySagar-Git:
@nihal467 this pr is ready for a final review

UdaySagar-Git:
@vigneshhari can you please review this , added a extra validation check for consultation
since the doctor notes will be specfic to consultation and is required

rithviknishad (approved):

UdaySagar-Git:
@nihal467 can you try re-running the tests now ?

khavinshankar (approved):

vigneshhari:
@sainak re-review required

sainak (approved):

bodhish:
@sainak why are we not merging this?

sainak:
@bodhish this needs a final review from vignesh

PR #2204, opened 2024-05-28
Adds functionality for `last_consultation_admitted_bed_type_list` in discharge patient filters

shivankacker:
## Proposed Changes

- Overwrites `last_consultation_admitted_bed_type_list` filter in Discharge Patient Viewset

### Associated Issue

- https://github.com/coronasafe/care_fe/issues/7869

## Merge Checklist

- [x] Tests added/fixed
- [x] Update docs in `/docs`
- [x] Linting Complete


nihal467:
QA test failed, @skks1212 make the necessary fixes

nihal467:
QA test failed, @skks1212 make the necessary fixes

nihal467:
LGTM

sainak (approved):

care/facility/api/viewsets/patient.py
```diff
@@ -609,9 +647,19 @@ class FacilityDischargedPatientViewSet(GenericViewSet, mixins.ListModelMixin):
         rest_framework_filters.OrderingFilter,
         PatientCustomOrderingFilter,
     )
-    filterset_class = PatientFilterSet
+    filterset_class = DischargePatientFilterSet
     queryset = (
-        PatientRegistration.objects.select_related(
+        PatientRegistration.objects.annotate(
```
vigneshhari, line 696: This might get expensive, if its really important can we just store it like last consultation?

shivankacker: sure

shivankacker: Although, if I make a field called `last_discharge_consultation` it would not work out... lets say a patient was discharged from two different facilities, then they would require different consultations in the discharge view set.

sainak: >lets say a patient was discharged from two different facilities, then they would require different consultations in the discharge view set.

in that case, we should also filter the consultation subquery based on users accessible facility

care/facility/api/viewsets/patient.py
```diff
@@ -599,6 +602,41 @@ def transfer(self, request, *args, **kwargs):
         return Response(data=response_serializer.data, status=status.HTTP_200_OK)
 
 
+class DischargePatientFilterSet(PatientFilterSet):
+
+    last_consultation_field = "last_discharge_consultation"
+
+    def filter_by_bed_type(self, queryset, name, value):
```
sainak, line 643: do we really need this filter?

shivankacker: Issue directs so

care/facility/api/viewsets/patient.py
shivankacker: Utilizing `PatientDRYFilter` for this, that works? I think it should cover.

care/facility/tests/test_patient_api.py
```diff
@@ -510,7 +510,7 @@ def get_base_url(self) -> str:
         )
 
     def test_filter_by_admitted_to_bed(self):
-        self.client.force_authenticate(user=self.user)
+        self.client.force_authenticate(user=self.super_user)
```
sainak: why is this failing for facility user? are permission not correct?

shivankacker: @sainak perms are correct, the tests were failing because the old user did not have access to the facility (it was created by super_user)

sainak: in that case you can set the users home facility while createing it `home_facility=cls.facility`

nihal467:
@skks1212 can you modify the test as requested before proceeding to further review

shivankacker:
@sainak made requested changes, please review

sainak (approved):

care/facility/api/viewsets/patient.py
```diff
@@ -599,19 +602,65 @@ def transfer(self, request, *args, **kwargs):
...
+        )
+        if "None" in values:
+            filter_q |= ~Q(
+                last_discharge_consultation__id__in=[
+                    x[1] for x in last_consultation_bed
+                ]
+            )
+            values.remove("None")
+
+        if values:
```
vigneshhari: values will always be there if value is evaluated as true.

care/facility/api/viewsets/patient.py
```diff
@@ -599,19 +602,65 @@ def transfer(self, request, *args, **kwargs):
...
+        last_consultation_bed = (
+            ConsultationBed.objects.filter(end_date__isnull=False)
+            .order_by("consultation__patient_id", "-end_date")
+            .distinct("consultation__patient_id")
+            .values_list("id", "consultation_id")
+        )
+        if "None" in values:
+            filter_q |= ~Q(
+                last_discharge_consultation__id__in=[
+                    x[1] for x in last_consultation_bed
```
vigneshhari: Subquery please. No point in the data making a roundtrip.

care/facility/api/viewsets/patient.py
```diff
@@ -599,19 +602,65 @@ def transfer(self, request, *args, **kwargs):
...
         filters.DjangoFilterBackend,
         rest_framework_filters.OrderingFilter,
         PatientCustomOrderingFilter,
     )
-    filterset_class = PatientFilterSet
+    filterset_class = DischargePatientFilterSet
     queryset = (
-        PatientRegistration.objects.select_related(
+        PatientRegistration.objects.annotate(
+            last_discharge_consultation__id=Subquery(
```
vigneshhari, line 697: You dont need to explicitly specify it as a subquery, it should automatically understand that its a subquery.
ids = Something.objects.values_list('id', flat=True)
SomethingElse.objects.filter(something_id__in=ids)

Will only have one query.

vigneshhari (changes requested):

care/facility/api/viewsets/patient.py
shivankacker: But will it not be required to use OuterRef?

shivankacker:
Lint failing but I have no idea why.
`pre-commit run --all-files` throws : 
```
flake8...................................................................Failed
- hook id: flake8
- exit code: 1

care/utils/assetintegration/hl7monitor.py:19:36: E713 test for membership should be 'not in'
care/utils/assetintegration/onvif.py:26:36: E713 test for membership should be 'not in'
care/utils/assetintegration/ventilator.py:19:36: E713 test for membership should be 'not in'
```

These files are not related to the PR
cc. @sainak

care/facility/api/viewsets/patient.py
```diff
@@ -599,19 +602,68 @@ def transfer(self, request, *args, **kwargs):
...
+
+        if "None" in values:
+            filter_q |= ~Q(
+                last_discharge_consultation__id__in=Subquery(
+                    last_consultation_bed_ids.values_list("consultation_id", flat=True)
+                )
+            )
+            values.remove("None")
+
+        if type(values) == list and len(values) > 0:
```
```suggestion
        if isinstance(values, list) and len(values) > 0:
```

sainak (approved):

care/facility/api/viewsets/patient.py
```diff
@@ -633,19 +636,68 @@ def transfer(self, request, *args, **kwargs):
...
     serializer_class = PatientListSerializer
     filter_backends = (
+        PatientDRYFilter,
         filters.DjangoFilterBackend,
         rest_framework_filters.OrderingFilter,
         PatientCustomOrderingFilter,
     )
-    filterset_class = PatientFilterSet
+    filterset_class = DischargePatientFilterSet
     queryset = (
```
anroopak, line 695: @shivankacker This seems like a pretty complex query. Can you add a simple comment about what this filter does?

vigneshhari (approved):
Ignore this approval

care/facility/api/viewsets/patient.py
shivankacker: @anroopak I have added some documentation for the filter, please have a look

anroopak (approved):

vigneshhari (approved):

PR #2222, opened 2024-06-03
limit user details based on user type

khavinshankar:
## Proposed Changes

- Removed sensitive fields from user list response for non admin users 

### Associated Issue

- https://github.com/coronasafe/care_fe/issues/7374

### DIFF
--- a/care/users/api/serializers/user.py
+++ b/care/users/api/serializers/user.py
@@ -432,3 +432,26 @@ class UserListSerializer(serializers.ModelSerializer):
             "home_facility",
             "video_connect_link",
         )
+
+    def to_representation(self, instance):
+        representation = super().to_representation(instance)
+        user = self.context["request"].user
+
+        if user.user_type < User.TYPE_VALUE_MAP["Reserved"]:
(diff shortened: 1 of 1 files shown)


nihal467:
@khavinshankar 
image

as a nurse id , iam able to view all restricted details mention in the issue, can you fix it

nihal467:
LGTM

rithviknishad (approved):

shivankacker (approved):

care/users/api/serializers/user.py
```diff
@@ -432,3 +432,26 @@ class Meta:
...
+                "local_body_object",
+                "district_object",
+                "state_object",
+                "user_type",
+                "created_by",
+            )
+
+            for key in list(representation.keys()):
+                if key not in whitelisted_fields:
+                    representation.pop(key)
```
anroopak, line 455: I think it should be the other way around. Instead of pop, can we use different serializers based on roles? The default one should be having the minimal fields and higher roles can have more fields?

PR #2255, opened 2024-06-08
Return files with consent lists and consent filters

shivankacker:
## Proposed Changes

- Adds files directly to consent API (prevents need to call two endpoints from the client)
- Adds the "has_consents" field to patient serializer
- Adds consent filters

## Merge Checklist

- [x] Tests added/fixed
- [x] Update docs in `/docs`
- [x] Linting Complete


nihal467:
LGTM

sainak (changes requested):
the better way of doing is to have a bool variable on the consultation itself so that we don't have to do these operations for each patient

shivankacker:
@sainak do you mean creating a Boolean field on consultation model that is changed with consent operations?

sainak:
> @sainak do you mean creating a Boolean field on consultation model that is changed with consent operations?

yes

care/facility/api/viewsets/patient.py
```diff
@@ -279,6 +283,53 @@ def filter_by_diagnoses(self, queryset, name, value):
...
+
+        filter_q = Q()
+
+        consultation_consent_files = (
+            FileUpload.objects.filter(
+                file_type=FileUpload.FileType.CONSENT_RECORD,
+                is_archived=False,
+            )
+            .order_by("associating_id")
+            .distinct("associating_id")
```
sainak: no need for distinct

care/facility/api/viewsets/patient.py
```diff
@@ -279,6 +283,53 @@ def filter_by_diagnoses(self, queryset, name, value):
...
+            .values_list("associating_id", flat=True)
+        )
+
+        consultations = PatientConsent.objects.filter(
+            external_id__in=[x for x in list(consultation_consent_files)],
```
sainak: consultation_consent_files is already a list

care/facility/models/patient.py
```diff
@@ -432,6 +432,10 @@ class TestTypeEnum(enum.Enum):
         AbhaNumber, on_delete=models.SET_NULL, null=True, blank=True
     )
 
+    has_consents = models.BooleanField(
```
sainak: consents are consultation specific 
wouldn't it be better to have this on consultation

care/facility/api/viewsets/patient.py
```diff
@@ -279,6 +283,52 @@ def filter_by_diagnoses(self, queryset, name, value):
...
+            FileUpload.objects.filter(
+                file_type=FileUpload.FileType.CONSENT_RECORD,
+                is_archived=False,
+            )
+            .order_by("associating_id")
+            .values_list("associating_id", flat=True)
+        )
+
+        consultations = PatientConsent.objects.filter(
+            external_id__in=[x for x in consultation_consent_files],
```
```suggestion
            external_id__in=consultation_consent_files,
```
consultation_consent_files is already a list

care/facility/api/viewsets/patient.py
```diff
@@ -279,6 +283,52 @@ def filter_by_diagnoses(self, queryset, name, value):
...
+            archived=False,
+        )
+
+        if "None" in values:
+            filter_q |= ~Q(
+                last_consultation__id__in=list(
+                    consultations.values_list("consultation_id", flat=True)
+                )
```
```suggestion
                last_consultation__id__in=(
                    consultations.values_list("consultation_id", flat=True)
                )
```

care/facility/api/viewsets/patient.py
```diff
@@ -279,6 +283,52 @@ def filter_by_diagnoses(self, queryset, name, value):
...
+            )
+            values.remove("None")
+
+        if values:
+            filter_q |= Q(
+                last_consultation__id__in=list(
```
```suggestion
                last_consultation__id__in=(
```

care/facility/api/viewsets/patient.py
```diff
@@ -279,6 +283,52 @@ def filter_by_diagnoses(self, queryset, name, value):
             )
         return queryset.filter(filter_q)
 
+    last_consultation__consent_types = MultiSelectFilter(
+        method="filter_by_has_consents"
+    )
+
+    def filter_by_has_consents(self, queryset, name, value: str):
```
sainak, line 286: @vigneshhari thoughts on this filter?

care/facility/api/viewsets/patient.py
shivankacker: this is for typecasting, without it you get errors.

sainak: Can you try to fix the queryset instead, it would be better if we can execute it without converting it to python

shivankacker: I dont quite understand what I should do to fix it, but even if we keep it this way, two queries are going to be made any ways, so is there any performance implications for this?

sainak (approved):

care/facility/api/viewsets/patient.py
vigneshhari: This is way too expensive, it loads up all consent records all the time, its easier to set a flag for a patient/consultation and then use that.

vigneshhari (changes requested):

care/facility/api/viewsets/patient.py
shivankacker: Can we merge this as it is P1 and I create a follow up PR with the flag approach?

vigneshhari: atleast move it as a  subquery to optimise it a bit, the current version is extremely unoptimised.

vigneshhari (changes requested):

care/facility/api/viewsets/patient.py
shivankacker: I have gone ahead and replaced the `has_consents` boolean field to an array field, that way we don't have to recalculate on the filter. Need your review

vigneshhari:
@sainak need another review and we are good to go

rithviknishad (approved):

sainak (approved):

khavinshankar (approved):

vigneshhari (approved):

PR #2288, opened 2024-07-02
Add taken at field for patient consultation events

sainak:
This pull request adds a new field called "taken_at" to the PatientConsultationEvent model. The "taken_at" field represents the date and time when the event was taken. This field is used to track the time entered by users while creating records. If not specified we fall back to created date


rithviknishad (approved):

sainak:
closing in favor of #2324

PR #2299, opened 2024-07-09
Fix race condition in consultation creation

sainak:
This pull request fixes a race condition that allowed duplicate consultations to be created. The issue was caused by concurrent requests creating consultations for the same patient. The fix ensures that only one consultation is created for each patient, preventing duplicates.

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -62,6 +62,9 @@ DATABASES["default"]["ATOMIC_REQUESTS"] = True
 DATABASES["default"]["CONN_MAX_AGE"] = env.int("CONN_MAX_AGE", default=0)
 DEFAULT_AUTO_FIELD = "django.db.models.BigAutoField"
 
+# timeout for setnx lock
+LOCK_TIMEOUT = env.int("LOCK_TIMEOUT", default=32)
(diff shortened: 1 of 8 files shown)


rithviknishad (approved):

config/settings/base.py
```diff
@@ -69,13 +69,10 @@
 # https://docs.djangoproject.com/en/dev/ref/settings/#caches
 CACHES = {
     "default": {
-        "BACKEND": "django_redis.cache.RedisCache",
+        "BACKEND": "redis_lock.django_cache.RedisCache",
```
vigneshhari: Lets keep the backend as the older one.

vigneshhari (changes requested):

config/settings/base.py
sainak: we need the redis_lock class as it adds the lock method on django_redis cache backend, it makes handling the connection for redis_lock easier


https://python-redis-lock.readthedocs.io/en/latest/readme.html

vigneshhari (approved):

vigneshhari (approved):

PR #2387, opened 2024-08-22
Improved asset bed relations for camera preset

rithviknishad:
## Proposed Changes

- Brief of changes made.

### Associated Issue

- https://github.com/coronasafe/care_fe/issues/7511
- FE: https://github.com/ohcnetwork/care_fe/pull/8253

### Architecture changes

## Merge Checklist

- [x] Tests added/fixed
- [ ] Update docs in `/docs`
- [x] Linting Complete


gigincg:
@rithviknishad Is this good to be tested/reviewed? Can you update the issue?

care/facility/api/serializers/camera_preset.py
```diff
@@ -0,0 +1,52 @@
...
+        read_only_fields = (
+            "created_date",
+            "modified_date",
+            "is_migrated",
+            "created_by",
+            "updated_by",
+        )
+
+    def get_asset_bed_obj(self):
+        return self.instance.asset_bed if self.instance else self.context["asset_bed"]
```
sainak:
```suggestion
        return self.instance.asset_bed if self.instance else self.context.get("asset_bed", {})
```

care/facility/api/serializers/camera_preset.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def validate_name(self, value):
+        presets_of_related_bed = (
+            CameraPreset.objects.filter(
+                asset_bed__bed_id=self.get_asset_bed_obj().bed_id
+            )
+            .only("name")
+            .values_list("name")
+        )
+        if value in [x[0] for x in presets_of_related_bed]:
```
sainak:
```suggestion
        if CameraPreset.objects.filter(
                asset_bed__bed_id=self.get_asset_bed_obj().bed_id, name=value
            ).exists():
```

care/facility/api/viewsets/camera_preset.py
```diff
@@ -0,0 +1,48 @@
...
+    queryset = CameraPreset.objects.all().select_related(
+        "asset_bed", "created_by", "updated_by"
+    )
+    lookup_field = "external_id"
+    permission_classes = (IsAuthenticated,)
+
+    def get_asset_bed_obj(self):
+        return get_object_or_404(
+            AssetBed.objects.filter(external_id=self.kwargs["assetbed_external_id"])
+        )
```
sainak: we should add some authz checks

care/facility/api/viewsets/camera_preset.py
```diff
@@ -0,0 +1,48 @@
...
+        asset_bed = self.get_asset_bed_obj()
+        return super().get_queryset().filter(asset_bed=asset_bed)
+
+    def get_serializer_context(self):
+        context = super().get_serializer_context()
+        context["asset_bed"] = self.get_asset_bed_obj()
+        return context
+
+
+class CameraPresetViewSet(GenericViewSet, ListModelMixin):
```
sainak, line 39: what are the requirements for this view?

rithviknishad: To list all presets that are either associated to a bed or associated to a camera asset (not specific to a single asset-bed link). Required to avoid N+1 API queries from FE when listing all accessible presets in the consultation feed tab and hub camera monitoring page.

care/facility/api/serializers/camera_preset.py
rithviknishad: is returning an empty dict a good idea? how about returning `None`?

care/facility/api/serializers/camera_preset.py
rithviknishad: silly mistake 😅 my bad

care/facility/api/viewsets/camera_preset.py
rithviknishad: Should we make two API endpoints instead? `GET /api/v1/(bed|asset)/{external_id}/camera-presets/`

Would be easier to handle authz. too this way.

sainak: yeah that would be better, you can reuse the same viewset and filter based on kwargs

sainak (approved):

nihal467:
LGTM

care/facility/api/viewsets/camera_preset.py
```diff
@@ -0,0 +1,94 @@
...
+    permission_classes = (IsAuthenticated,)
+
+    def get_asset_bed_obj(self):
+        user = self.request.user
+        queryset = AssetBed.objects.filter(
+            external_id=self.kwargs["assetbed_external_id"]
+        )
+        if user.is_superuser:
+            pass
+        elif user.user_type >= User.TYPE_VALUE_MAP["StateLabAdmin"]:
```
vigneshhari: Can we move this logic to use an abstracted function? like what we do in facilities.

vigneshhari (changes requested):

care/facility/api/viewsets/camera_preset.py
rithviknishad: Did for `get_asset_bed_obj`, `get_asset_obj` and `get_bed_obj` and also updated assetbed/asset/bed viewsets to reuse the same

sainak (approved):

care/facility/migrations/0466_camera_presets.py
```diff
@@ -0,0 +1,171 @@
...
+                    if asset_bed.row_number != 1:
+                        assetbeds_to_delete.append(asset_bed.id)
+                else:
+                    assetbeds_to_delete.append(asset_bed.id)
+
+            CameraPreset.objects.bulk_create(presets_to_create)
+            AssetBed.objects.filter(id__in=assetbeds_to_delete).update(
+                deleted=True, meta={}
+            )
+        AssetBed.objects.filter(deleted=False, asset__asset_class="ONVIF").update(
```
vigneshhari: Migrations usually delete data later on, deleting data right away is usually not the best approach.

care/facility/migrations/0466_camera_presets.py
```diff
@@ -0,0 +1,171 @@
...
+                            ).order_by("id")[0],
+                            position={
+                                "x": position["x"],
+                                "y": position["y"],
+                                "zoom": position["zoom"],
+                            },
+                            is_migrated=True,
+                        )
+                    )
+                    if asset_bed.row_number != 1:
```
vigneshhari, line 67: Why is 1 so special ?

care/facility/migrations/0466_camera_presets.py
```diff
@@ -0,0 +1,171 @@
...
+                if position := asset_bed.meta.get("position"):
+                    presets_to_create.append(
+                        CameraPreset(
+                            name=name,
+                            asset_bed=AssetBed.objects.filter(
+                                asset=asset_bed.asset, bed=asset_bed.bed
+                            ).order_by("id")[0],
+                            position={
+                                "x": position["x"],
+                                "y": position["y"],
```
vigneshhari: Are we really sure this is the schema that was followed, i see no exception handling anywhere.

care/facility/migrations/0466_camera_presets.py
```diff
@@ -0,0 +1,171 @@
...
+    def delete_asset_beds_without_asset_class(apps, schema_editor):
+        AssetBed = apps.get_model("facility", "AssetBed")
+        AssetBed.objects.filter(asset__asset_class__isnull=True).delete()
+
+    def backfill_camera_presets(apps, schema_editor):
+        AssetBed = apps.get_model("facility", "AssetBed")
+        CameraPreset = apps.get_model("facility", "CameraPreset")
+
+        paginator = Paginator(
+            AssetBed.objects.annotate(
```
vigneshhari, line 30: Do we have that many assetbed realtions in prod right now ?

care/facility/migrations/0466_camera_presets.py
```diff
@@ -0,0 +1,171 @@
...
+                row_number=Window(
+                    expression=RowNumber(),
+                    partition_by=[F("asset"), F("bed")],
+                    order_by=F("id").asc(),
+                )
+            )
+            .filter(deleted=False, asset__asset_class="ONVIF")
+            .order_by("asset", "bed", "id"),
+            1000,
+        )
```
sainak, line 40: the paginator might cause issue with the grouping logic in the for loop

sainak: nvm row number is preserved

vigneshhari:
This is a very very specific implementation, we should try to avoid creating features like this as much as possible. Care needs more abstract thinking.

vigneshhari (changes requested):

vigneshhari:
Also, i don't see any tests for this feature.

care/facility/migrations/0466_camera_presets.py
rithviknishad: Will remove it from this PR and create separate PR to remove it.

care/facility/migrations/0466_camera_presets.py
rithviknishad: We need to preserve the first asset-bed link since the camera preset would be associated to the asset-bed link. So far each old preset had it's own asset-bed record. However, asset-bed records would be now unique together for asset and bed and the newly created preset record would be linked to this asset bed, hence we need to preserve the first asset-bed.

care/facility/migrations/0466_camera_presets.py
rithviknishad: Tried with develop db copy and worked. Will handle exception anyways. Should we skip creating the preset record for such cases where schema is not followed?

care/facility/migrations/0466_camera_presets.py
rithviknishad: The duplicate asset-bed records would have to be marked as deleted for the unique constraint to be applied, and for the front-end to know which asset-bed record to link to when the preset is being created. 

We can however skip deleting the meta data from these old records so that data is still preserved, just that duplicate asset-bed link records would no longer be accessible by the viewset.

care/facility/migrations/0466_camera_presets.py
rithviknishad: So that the first asset-bed record is preserved, and only the subsequent (duplicates) are marked for deletion.

vigneshhari (approved):

PR #2399, opened 2024-08-26
HCX Plug

khavinshankar:
Seperates HCX into a plug


care/facility/utils/reports/discharge_summary.py
```diff
@@ -181,7 +179,6 @@ def get_discharge_summary_data(consultation: PatientConsultation):
     return {
         "patient": consultation.patient,
         "samples": samples,
-        "hcx": hcx,
```
sainak, line 184: we should think of a way to allow apps to expose discharge summary template sections and data

cc: @vigneshhari

sainak (approved):

rithviknishad (approved):

care/facility/api/serializers/patient.py
```diff
@@ -94,37 +91,6 @@ class PatientListSerializer(serializers.ModelSerializer):
 
     assigned_to_object = UserBaseMinimumSerializer(source="assigned_to", read_only=True)
 
-    # HCX
-    has_eligible_policy = serializers.SerializerMethodField(
```
anroopak, line 98: FE APIs which use these fields, how are they handled?

anroopak (approved):
LGTM.

sainak (approved):

rithviknishad (approved):

PR #2429, opened 2024-09-06
Add feature flags

vigneshhari:
This implementation adds the concept of a feature flag and how it can be used in API's


care/facility/models/feature_flag.py
```diff
@@ -0,0 +1,34 @@
...
+from care.utils.models.base import BaseModel
+
+from django.db import models
+
+
+class FacilityFlag(BaseModel):
+    facility = models.ForeignKey(
+        "Facility", on_delete=models.CASCADE, null=False, blank=False
+    )
+    flag = models.CharField(max_length=1024)
```
anroopak: Consider using `models.TextChoices`
Ref: https://docs.djangoproject.com/en/5.1/ref/models/fields/#enumeration-types

vigneshhari: I considered it but the choice field is dyamic in nature, plugins can create more flags but they wont have migrations associated with it as that will create unwanted migration dependencies, I think desigining the validations ourselves is easier than going through that path.

vigneshhari: Adding more context : Plugins can be added or removed from care at any point, each plugin might have their own feature flags, we should have some mechanism to keep a registry for all these feature flags and maintain them.

shivankacker:
@vigneshhari this is under progress right? I would like to go ahead with the frontend implementation for this. I am imagining there will be an API exposed that returns whether the particular user is allowed to use a feature or not, which we can use in the frontend to enable/disable features.

vigneshhari:
@sainak is taking over this MR

Issue #2237, opened 2024-06-05
Add a global 20 seconds timeout to middleware requests


sainak:
https://github.com/ohcnetwork/care/blob/4f6296ac76ceedc0a8c07a83689a3c7fcfe9f2a0/care/utils/assetintegration/base.py#L35-L61

add a settings variable to set timeout for middleware requests
use 20 seconds as defaults

praffq:
@sainak I'll like to work on this issue

sainak:
completed in https://github.com/ohcnetwork/care/pull/2434

PR #2434, opened 2024-09-07
added timeout to middleware requests

praffq:
## Proposed Changes

- Added a timeout setting for middleware requests with default to 20 seconds

### Associated Issue

- #2237 

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete


config/settings/base.py
```diff
@@ -663,3 +663,6 @@
 TASK_SUMMARIZE_DISTRICT_PATIENT = env.bool(
     "TASK_SUMMARIZE_DISTRICT_PATIENT", default=True
 )
+
+# Timeout for middleware request (in seconds)
+MIDDLEWARE_API_TIMEOUT = 20
```
sainak:
```suggestion
MIDDLEWARE_REQUEST_TIMEOUT = env.int("MIDDLEWARE_REQUEST_TIMEOUT", 20)
```

care/utils/assetintegration/base.py
```diff
@@ -16,6 +16,7 @@ def __init__(self, meta):
         self.host = self.meta["local_ip_address"]
         self.middleware_hostname = self.meta["middleware_hostname"]
         self.insecure_connection = self.meta.get("insecure_connection", False)
+        self.timeout = getattr(settings, "MIDDLEWARE_API_TIMEOUT", 20)
```
sainak: this is not required 
you can reference it directly from settings.MIDDLEWARE_REQUEST_TIMEOUT

praffq:
updated

care/utils/assetintegration/base.py
```diff
@@ -16,6 +16,7 @@ def __init__(self, meta):
         self.host = self.meta["local_ip_address"]
         self.middleware_hostname = self.meta["middleware_hostname"]
         self.insecure_connection = self.meta.get("insecure_connection", False)
+        self.timeout = settings.MIDDLEWARE_REQUEST_TIMEOUT
```
sainak, line 19: there is no use in setting this instance var before using it, we could instead accept timeout as an optional kwarg timeout in the init method

update the children classes as well

sainak (approved):

vigneshhari (approved):

PR #2447, opened 2024-09-13
Do not allow future log update taken_at time

### LINKED ISSUE #2439: Log Update API accepts "taken at" of future time
**Describe the bug**
Even though restricted through frontend, the value still gets accepted

shivankacker:
## Proposed Changes

- Will throw an error if the `taken_at` time is in the future
- Fixes #2439

## Merge Checklist

- [x] Tests added/fixed
- [x] Update docs in `/docs`
- [x] Linting Complete

### DIFF
--- a/care/facility/api/serializers/daily_round.py
+++ b/care/facility/api/serializers/daily_round.py
@@ -1,6 +1,7 @@
 from datetime import timedelta
 
 from django.db import transaction
+from django.utils import timezone
 from django.utils.timezone import localtime, now
 from rest_framework import serializers
 from rest_framework.exceptions import ValidationError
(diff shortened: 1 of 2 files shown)


care/facility/api/serializers/daily_round.py
```diff
@@ -295,3 +296,8 @@ def validate(self, attrs):
                 validated["bed_id"] = bed_object.id
 
         return validated
+
+    def validate_taken_at(self, value):
+        if value > timezone.now():
```
sainak: if taken at field allows null
this will throw value error on comparison with none

sainak (approved):

anroopak (approved):

vigneshhari (approved):

PR #2448, opened 2024-09-13
Adds necessary fields for Community Nurse's log update

### LINKED ISSUE #2441: New round type "Community Nurse's Log" and additional fields
Backend issue for: https://github.com/ohcnetwork/care_fe/issues/8488

rithviknishad:
## Proposed Changes

- Added fields necessary for Community Nurse's log update
- Added relevant event types

### Associated Issue

- Issue: https://github.com/ohcnetwork/care_fe/issues/8488
- Resolves #2441
- Front-end changes: https://github.com/ohcnetwork/care_fe/pull/8522

## Merge Checklist

- [x] Tests added/fixed
- [x] Linting Complete


sainak (approved):

anroopak (approved):
LGTM.

care/facility/api/serializers/daily_round.py
```diff
@@ -41,6 +41,27 @@ class DailyRoundSerializer(serializers.ModelSerializer):
 
     rounds_type = ChoiceField(choices=DailyRound.RoundsTypeChoice, required=True)
 
+    # Community Nurse's Log
+
+    bowel_issue = ChoiceField(
+        choices=DailyRound.BowelDifficultyType.choices, required=False
+    )
+    bladder_drainage = ChoiceField(
```
anroopak, line 49: Why do we need these fields in the serializer? It is already part of the model right?

rithviknishad: Without the serializer having the choices explicitly specified, front-end will have to reverse the choice text to it's integer equivalent when making the POST.

However, with the serializer explicitly specified, both are possible and that way front-end need not have to be concerned about what the integer equivalents are for each of the choice. Hence makes it simpler in the front-end.

Here's an attempt with the serializer fields removed:

image


image

PR #2449, opened 2024-09-15
Major dependencies update

sainak:
## Proposed Changes

- update docker images to 3.12-slim-bookworm
- update django to 5.1.1 (ran django-upgrade to fix deprecated warnings)
- dropped djdt as it is not compatibe with py312
- update dependabot config to create grouped pull request for dependency updates

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete


vigneshhari (approved):

docker/prod.Dockerfile
```diff
@@ -38,44 +41,26 @@ RUN ARCH=$(dpkg --print-architecture) && \
...
-
-WORKDIR ${APP_HOME}
-
 RUN apt-get update && apt-get install --no-install-recommends -y \
   libpq-dev gettext wget curl gnupg \
   && apt-get purge -y --auto-remove -o APT::AutoRemove::RecommendsImportant=false \
   && rm -rf /var/lib/apt/lists/*
 
-# copy typst binary from builder stage
-COPY --from=builder --chmod=0755 /usr/local/bin/typst /usr/local/bin/typst
```
vigneshhari, line 72: Revert, we need typst

config/settings/local.py
```diff
@@ -13,22 +13,6 @@
 # https://github.com/jazzband/django-silk#requirements
 INSTALLED_APPS += ["silk"]  # noqa F405
 MIDDLEWARE += ["silk.middleware.SilkyMiddleware"]  # noqa F405
-# https://github.com/jazzband/django-silk#profiling
-SILKY_PYTHON_PROFILER = True
```
vigneshhari, line 17: We can keep the profiler.

config/authentication.py
```diff
@@ -209,7 +209,7 @@ def authenticate_header(self, request):
         return "Bearer"
 
     def authenticate(self, request):
-        jwt_token = request.META.get("HTTP_AUTHORIZATION")
+        jwt_token = self.get_header(request)
```
vigneshhari, line 212: are you sure we get the Auth header with this method? it will default to the header we use in settings.

vigneshhari (changes requested):

config/settings/local.py
sainak: https://github.com/jazzband/django-silk?tab=readme-ov-file#profiling python 312 borked nested profilers

config/authentication.py
sainak: yes 
image

vigneshhari (approved):

PR #2505, opened 2024-09-29
Added care livekit plugin

khavinshankar:
Enables doctor connect through livekit

https://github.com/ohcnetwork/care_fe/pull/8661
https://github.com/ohcnetwork/care_livekit


vigneshhari:
The requirements should not be added here, it should be maintained in the source repo in its own requirements. 
Is this really needed for all our deployments?

gigincg:
@vigneshhari This is not required to be enabled by default,

@vigneshhari We can keep the requirements for livekit in the plugin itself right?

vigneshhari:
@gigincg Yes plugin requirements should not be maintained in the core app. You can remove the plug config as well if its not required by default. We can close this MR because no changes are required for this.

PR #2515, opened 2024-10-01
New RBAC Implementation

vigneshhari:
This PR includes the new RBAC system.
This PR will not be a complete rewrite of the current system. This PR will only introduce the new architecture, In time we will move all of the logic from the current implementation to the new architecture.


care/security/models/role.py
```diff
@@ -0,0 +1,42 @@
...
+    context = models.CharField(
+        max_length=1024
+    )  # We can add choices here as well if needed
+    is_system = models.BooleanField(
+        default=False
+    )  # Denotes if role was created on the fly
+    temp_deleted = models.BooleanField(default=False)
+
+    class Meta:
+        unique_together = ("name", "context")
```
sainak: ignore deleted and prefer unique constraints

care/security/apps.py
```diff
@@ -0,0 +1,11 @@
+from django.apps import AppConfig
+from django.utils.translation import gettext_lazy as _
+
+
+class FacilityConfig(AppConfig):
```
sainak:
```suggestion
class SecurityConfig(AppConfig):
```

care/security/permissions/permissions.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+class PermissionHandler:
+    pass
+
+
+from care.security.permissions.facility import FacilityPermissions  # noqa: E402
+
+
+class PermissionController:
```
sainak: we move controlleres to .controllers or ..controllers.permission_controllers

care/security/models/role.py
vigneshhari: Roles will not be deletable unless all its usages are deleted, at that point we can do a force delete and not an archive

PR #2572, opened 2024-10-28
Flushed care_static_data prefixed redis keys

### LINKED ISSUE #2249: Flush Redis before loading static data
Remove the existing keys of care_static_data before proceeding with the loading of static data in redis

BlessedRice2004:
* Used delete_pattern from django-redis repo to delete keys

## Proposed Changes

- Deletes redis keys with ``care_static_data`` prefix

### Associated Issue

- Fixed #2249, Used Cache(from django) which has a delete_pattern() function to delete keys with ``prefix care_static_data`` within try/catch blocks to prevent uncaught exceptions.

## Merge Checklist

- [x] Linting Complete

### DIFF
--- a/care/facility/management/commands/load_redis_index.py
+++ b/care/facility/management/commands/load_redis_index.py
@@ -17,6 +17,17 @@ class Command(BaseCommand):
     help = "Loads static data to redis"
 
     def handle(self, *args, **options):
+        try:
+            deleted_count = cache.delete_pattern("care_static_data*", itersize=25_000)
+            self.stdout.write(
+                f"Deleted {deleted_count} keys with prefix 'care_static_data'"
+            )
+        except Exception as e:
+            self.stdout.write(
+                f"Failed to delete keys with prefix 'care_static_data': {e}"
+            )
+            return
+
         if cache.get("redis_index_loading"):
             self.stdout.write("Redis Index already loading, skipping")
             return
--- a/care/facility/tasks/redis_index.py
+++ b/care/facility/tasks/redis_index.py
@@ -46,7 +46,7 @@ def load_redis_index():
         except ModuleNotFoundError:
             logger.debug("Module %s not found", module_path)
         except Exception as e:
-            logger.exception("Error loading static data for %s: %s", plug.name, e)
+            logger.error("Error loading static data for %s: %s", plug.name, e)
 
     cache.delete("redis_index_loading")
     logger.info("Redis Index Loaded")
--- a/care/facility/tasks/redis_index.py
+++ b/care/facility/tasks/redis_index.py
@@ -15,6 +15,13 @@ logger: Logger = get_task_logger(__name__)
 
 @shared_task
 def load_redis_index():
+    try:
+        deleted_count = cache.delete_pattern("care_static_data*", itersize=25_000)
+        logger.info("Deleted %s keys with prefix 'care_static_data'", deleted_count)
+    except Exception as e:
+        logger.error("Failed to delete keys with prefix 'care_static_data': %s", e)
+        return
+
     if cache.get("redis_index_loading"):
         logger.info("Redis Index already loading, skipping")
         return
@@ -37,9 +44,9 @@ def load_redis_index():
             if load_static_data:
                 load_static_data()
         except ModuleNotFoundError:
-            logger.info("Module %s not found", module_path)
+            logger.debug("Module %s not found", module_path)
         except Exception as e:
-            logger.info("Error loading static data for %s: %s", plug.name, e)
+            logger.exception("Error loading static data for %s: %s", plug.name, e)
 
     cache.delete("redis_index_loading")
     logger.info("Redis Index Loaded")


care/facility/tasks/redis_index.py
```diff
@@ -15,6 +16,13 @@
 
 @shared_task
 def load_redis_index():
+    try:
+        deleted_count = BaseCommand.delete_pattern("care_static_data*")
```
vigneshhari: There might be an API for this, can you try to use that instead?

care/facility/tasks/redis_index.py
```diff
@@ -15,6 +16,13 @@
...
+        logger.info("Deleted %s keys with prefix 'care_static_data'", deleted_count)
+    except Exception as e:
+        logger.info("Failed to delete keys with prefix 'care_static_data': %s", e)
```
vigneshhari: Can we log this as an exception itself ? info logs might not be picked up by sentry

care/facility/management/commands/load_redis_index.py
```diff
@@ -17,6 +17,17 @@ class Command(BaseCommand):
     help = "Loads static data to redis"
 
     def handle(self, *args, **options):
+        try:
+            deleted_count = self.delete_pattern("care_static_data*")
```
sainak:
```suggestion
            deleted_count = cache.delete_pattern("care_static_data*")
```

vigneshhari (changes requested):

care/facility/tasks/redis_index.py
sainak:
```suggestion
        deleted_count = cache.delete_pattern("care_static_data*")
```

care/facility/tasks/redis_index.py
```diff
@@ -4,6 +4,7 @@
 from celery import shared_task
 from celery.utils.log import get_task_logger
 from django.core.cache import cache
+from django.core.management import BaseCommand
```
sainak:
```suggestion
```

sainak (changes requested):
reference delete_pattern from cache backend 

https://github.com/jazzband/django-redis/blob/master/README.rst#scan--delete-keys-in-bulk


use itersize=25_000, to speed up the operation

care/facility/management/commands/load_redis_index.py
BlessedRice2004: Thank you for the feedback. I've changed it to...
```deleted_count = cache.delete_pattern("care_static_data*", itersize=25_000)```

File: care/facility/tasks/redis_index.py (hunk 1 above)
Comment (BlessedRice2004) [reply care/facility/tasks/redis_index.py]: Again, thank you for the feedback. I've changed it to...
```deleted_count = cache.delete_pattern("care_static_data*", itersize=25_000)```

Comment (BlessedRice2004) [reply care/facility/tasks/redis_index.py]: Thanks for letting me know. Sainak suggested using cache. I've changed it to ``deleted_count = cache.delete_pattern("care_static_data*", itersize=25_000)``

File: care/facility/tasks/redis_index.py (hunk 4 above)
Comment (BlessedRice2004) [reply care/facility/tasks/redis_index.py]: Since we're using cache, I've followed your recommendations and deleted the import to use cache

File: care/facility/tasks/redis_index.py (hunk 2 above)
Comment (BlessedRice2004) [reply care/facility/tasks/redis_index.py]: Thanks for the tip. I've changed line 23 to say ``logger.error("Failed to delete keys with prefix 'care_static_data': %s", e)`` instead. ``logger.warning()`` and ``logger.exception`` were both options, but no one else uses ``logger.exception`` in the codebase, and ``logger.warning()`` comes up less than ``logger.error()``

On a side note, how come the logger commands on line 47 and 49 say ``logger.info()`` when they're treated as exceptions?
image

Comment (sainak) [reply care/facility/tasks/redis_index.py]: yeah, it would be good to update the log statement on line 49 to be logger.exception but, the one on line 47 is just for debugging so logger.debug would be used for that, you can include these changes in your pr

**Comment by BlessedRice2004** [comment]
All the requested changes should be in place now 👍

File: care/facility/tasks/redis_index.py [hunk 5]
```diff
@@ -37,9 +44,9 @@ def load_redis_index():
             if load_static_data:
                 load_static_data()
         except ModuleNotFoundError:
-            logger.info("Module %s not found", module_path)
+            logger.debug("Module %s not found", module_path)
         except Exception as e:
-            logger.info("Error loading static data for %s: %s", plug.name, e)
+            logger.exception("Error loading static data for %s: %s", plug.name, e)
```
sainak:
```suggestion
            logger.error("Error loading static data for %s: %s", plug.name, e)
```

I meant to say logger.error in the previous review

sainak (approved):

vigneshhari (approved):

vigneshhari:
If the data load fails, we will end up with no data in the system, this is a flaw. An ideal implementation would only delete stale items maintaining the existing data somehow.
We will ignore this flaw for now because we are redesigning the coding system entirely.

PR #2595, opened 2024-11-11
Modified User Route for User Management Redesign

Jacobjeevan:
## Proposed Changes

- Added get user route
	- For fetching details of other users (existing method has permissions enabled and is used widely, so didn't want to change that).
- Added additional permissions check for change password
	- Allowing district admins and above to change password
- Enable username search for FacilityUsers

### Associated Issue

- FE issue [8878](https://github.com/ohcnetwork/care_fe/issues/8878)

## Merge Checklist

- [x] Tests added/fixed
- [ ] Update docs in `/docs`
- [x] Linting Complete

## Summary by CodeRabbit

- **New Features**
	- Introduced a case-insensitive `username` filter for user searches.
	- Added `last_login` field to user data representations.
	- Implemented a soft delete mechanism for user records.
	- Enhanced user retrieval logic with improved error handling.

- **Bug Fixes**
	- Improved validation and permission checks in user management operations.

- **Tests**
	- Expanded test coverage for user management functionalities, including permission checks and search capabilities.


care/users/api/viewsets/change_password.py
sainak, line 1: It would be better to combine this with the users viewset

cc: @vigneshhari

care/users/api/serializers/user.py
```diff
@@ -335,6 +335,7 @@ class Meta:
             "pf_auth",
             "read_profile_picture_url",
             "user_flags",
+            "last_login",
```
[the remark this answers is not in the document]
Jacobjeevan: The field has to be in fields to be included in serialized output, no? Changing it to read_only_fields makes it inaccessible via the API. As for modification, since the field is managed by Django, if frontend were to pass data for last_login, Django should ignore it.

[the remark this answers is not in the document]
Jacobjeevan: It wasn't accessible, likely due to inheritance from SignUpSerializer. I can add last_login to both fields and read_only_fields instead.

bodhish:
@vigneshhari can you review this, we want the user page out!

care/users/api/viewsets/change_password.py
```diff
@@ -29,7 +29,26 @@ class ChangePasswordView(UpdateAPIView):
...
 
     def update(self, request, *args, **kwargs):
-        self.object = self.request.user
+        username = request.data.get("username")
+        if not username:
+            return Response(
+                {"message": ["Username is required"]},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+        self.object = User.objects.filter(username=username).first()
```
vigneshhari: use the get_object_or_404 util from django.

care/users/api/viewsets/change_password.py
```diff
@@ -48,3 +67,11 @@ def update(self, request, *args, **kwargs):
             return Response({"message": "Password updated successfully"})
 
         return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+    def has_permission(self, request, user):
+        authuser = request.user
+        return (
+            authuser == user
+            or authuser.is_superuser
+            or authuser.user_type >= User.TYPE_VALUE_MAP["DistrictAdmin"]
```
vigneshhari: This is not required, lets only allow it for superadmins. we are going to remove the dist and state admins soon. 
also a dist admin only has access to a user if he is of the same dist.

care/users/api/viewsets/users.py
```diff
@@ -155,6 +167,16 @@ def get_queryset(self):
 
     def get_object(self) -> User:
         try:
+            if self.request.method == "GET" and not self.kwargs.get("username"):
```
vigneshhari: I don't see any new features here, why is this rebuilt ?

vigneshhari (commented):
Chnages Required

care/users/api/viewsets/users.py
Jacobjeevan: > I don't see any new features here, why is this rebuilt ?

For making the user's details available to others (DRYPermissions had limited who had access to it); logged in user would be be able to view user details even if they don't have edit permissions.

care/facility/api/viewsets/facility_users.py
```diff
@@ -16,6 +16,7 @@ class UserFilter(filters.FilterSet):
         choices=[(key, key) for key in User.TYPE_VALUE_MAP],
         coerce=lambda role: User.TYPE_VALUE_MAP[role],
     )
+    username = filters.CharFilter(field_name="username", lookup_expr="icontains")
```
rithviknishad, line 19: Doesn't line 35 already allow this?

image

Jacobjeevan: @rithviknishad Nah, that didn't work for some reason 🤔

vigneshhari: search fields will look for a search key and return anything that contains the term in any of the fields, filter works on one field only.

care/users/api/viewsets/users.py
```diff
@@ -203,6 +225,17 @@ def destroy(self, request, *args, **kwargs):
         user.save(update_fields=["is_active"])
         return Response(status=status.HTTP_204_NO_CONTENT)
 
+    def has_permission(self, user):
+        requesting_user = self.request.user
+        return (
+            requesting_user == user
+            or requesting_user.is_superuser
+            or (
+                requesting_user.state == user.state
```
vigneshhari: Shouldn't we also check if the user type is a distadmin or a stateadmin ?

vigneshhari (changes requested):

care/facility/api/viewsets/facility_users.py
Jacobjeevan: > search fields will look for a search key and return anything that contains the term in any of the fields, filter works on one field only.

Ah alright, search_fields were not being applied to the query set; I added that to get_queryset and removed the username filter and that works (FE only passes in username anyways, so gets the same result).

care/users/api/viewsets/users.py
Jacobjeevan: > Shouldn't we also check if the user type is a distadmin or a stateadmin ?

Nah, in this case the view should be available to all users in the same state/district (i.e. nurse user can view the details of another nurse user).

vigneshhari: I might be talking out of context here, why is the user only allowed to see users within the same state/district but nothing else ? also why would you check district when you check state? if state is the same then dist check is irrelevant right ?

Jacobjeevan: > I might be talking out of context here, why is the user only allowed to see users within the same state/district but nothing else ? also why would you check district when you check state? if state is the same then dist check is irrelevant right ?

So if you were to login using dummynurse1 and compare the results of following:

UserViewSet (/users)

image

FacilityUsersViewSet (/facility/id/users)

image

Latter would include devdistrictadmin as well (linked to the Dummy Facility 40 in the dummy data).

I used the state/district check so that devdistrictadmin's details page would be viewable to dummynurse1.

We could instead filter for accessible facilities here, which UserViewSet uses, but I'd need to know the conditional check to display "more details" button on FE.

```
user.id in Subquery(FacilityUser.objects.filter(
                        facility_id__in=get_accessible_facilities(requesting_user)
                    ).values("user_id")
                )

```

Jacobjeevan: @vigneshhari Following up based on the conversation in the call, so state check should suffice for now, right 🤔

care/facility/api/viewsets/facility_users.py
```diff
@@ -35,12 +35,21 @@ class FacilityUserViewSet(GenericViewSet, mixins.ListModelMixin):
...
             facility = Facility.objects.get(
                 external_id=self.kwargs.get("facility_external_id"),
             )
             queryset = facility.users.filter(
                 deleted=False,
             ).order_by("-last_login")
+            if search_fields:
+                queryset = queryset.filter(
+                    **{key: value for key, value in search_fields.items() if value}
+                )
```
rithviknishad: `search_fields` is handled by `drf_filters.SearchFilter` filter back-end for us. Wasn't it working?

ref: https://www.django-rest-framework.org/api-guide/filtering/#searchfilter

cc: @sainak

Jacobjeevan: > `search_fields` is handled by `drf_filters.SearchFilter` filter back-end for us. Wasn't it working?
> 
> ref: https://www.django-rest-framework.org/api-guide/filtering/#searchfilter
> 
> cc: @sainak

@rithviknishad It will work if `search_text` is passed in the query parameter from FE. The only issue is search for facility users is specified as username only (similar to users page), but this viewset supports username/first and last names.

For context, LinkedFacilityUsers calls the same route and the AutoComplete input supports search by name/username there.

Anyways, the primary reason why I added this is to have similar behavior as users;  so we can just pass in `username` as query parameter from FE.

rithviknishad (approved):

vigneshhari (approved):

care/users/api/viewsets/users.py
```diff
@@ -155,6 +155,9 @@ def get_queryset(self):
 
     def get_object(self) -> User:
         try:
+            if self.action == "retrieve":
+                username = self.kwargs.get("username")
+                return get_object_or_404(User, username=username)
```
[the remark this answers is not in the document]
sainak: this suggestion is not valid as the get_object_or_404 does not support passing a custom message

sainak (approved):

PR #2620, opened 2024-11-27
Fixing Manually cascade delete M2M models when related model is deleted

### LINKED ISSUE #1330: Manually cascade delete M2M models when related model is deleted.
**Describe the bug**
Currently, most M2M models are not deleted when their related model is deleted. The `on_delete` doesn't take effect as we use soft delete for all the models that inherits from `BaseModel`.

**Expected behavior**
`ModelAB` should be updated to `deleted=True` when `ModelA` or `ModelB` is deleted.

### Many to Many models in CARE that are potentially having this issue.

- [ ] `UserDefaultAssetLocation`
- [ ] `FacilityDefaultAssetLocation`
- [x] `AssetBed` #1322
- [ ] `ConsultationBed`
- [ ] `FacilityUser`
- [ ] `InvestigationValue`
- [ ] `PatientSample`

praffq:
## Proposed Changes

- Manually cascade delete M2M models when related model is deleted

### Associated Issue

- fixes #1330 

## Merge Checklist

- [x] Tests added/fixed
- [x] Linting Complete

## Release Notes

- **New Features**
	- Enhanced deletion logic across multiple models to prevent deletion if referenced elsewhere, ensuring data integrity.
	- New validation checks for deleting `AssetLocation`, `Bed`, `Facility`, `PatientRegistration`, `PatientConsultation`, and investigation-related records.
	- Improved user deletion logic with transaction handling and validation checks.

- **Bug Fixes**
	- Improved error handling during deletion processes to raise appropriate validation errors.

- **Tests**
	- Added comprehensive unit tests to validate the new deletion behaviors and ensure referential integrity across related models.

### DIFF
--- a/care/facility/models/asset.py
+++ b/care/facility/models/asset.py
@@ -59,6 +59,9 @@ class AssetLocation(BaseModel, FacilityRelatedPermissionMixin):
         if UserDefaultAssetLocation.objects.filter(location=self).exists():
             error = f"Cannot delete AssetLocation {self} because they are referenced as `location` in UserDefaultAssetLocation records."
             raise ValidationError(error)
+        if FacilityDefaultAssetLocation.objects.filter(location=self).exists():
+            error = f"Cannot delete AssetLocation {self} because they are referenced as `location` in FacilityDefaultAssetLocation records."
+            raise ValidationError(error)
         return super().delete(*args)
 
 
--- a/care/facility/models/asset.py
+++ b/care/facility/models/asset.py
@@ -2,6 +2,7 @@ import enum
 import uuid
 
 from django.contrib.contenttypes.models import ContentType
+from django.core.exceptions import ValidationError
 from django.db import models
 from django.db.models import JSONField, Q
 
@@ -53,6 +54,12 @@ class AssetLocation(BaseModel, FacilityRelatedPermissionMixin):
     middleware_address = models.CharField(
         null=True, blank=True, default=None, max_length=200
     )
+
+    def delete(self, *args):
+        if UserDefaultAssetLocation.objects.filter(location=self).exists():
+            error = f"Cannot delete AssetLocation {self} because they are referenced as `location` in UserDefaultAssetLocation records."
+            raise ValidationError(error)
+        return super().delete(*args)
 
 
 class AssetType(enum.Enum):
--- a/care/facility/models/bed.py
+++ b/care/facility/models/bed.py
@@ -59,6 +59,10 @@ class Bed(BaseModel):
         return super().save(*args, **kwargs)
 
     def delete(self, *args, **kwargs) -> None:
+        if ConsultationBed.objects.filter(bed=self).exists():
+            error = f"Cannot delete Bed {self} because they are referenced as `bed` in ConsultationBed records."
+            raise ValidationError(error)
+
         AssetBed.objects.filter(bed=self).update(deleted=True)
         super().delete(*args, **kwargs)
 
--- a/care/facility/models/facility.py
+++ b/care/facility/models/facility.py
@@ -1,6 +1,7 @@
 from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.contrib.postgres.fields import ArrayField
+from django.core.exceptions import ValidationError
 from django.core.validators import MinValueValidator
 from django.db import models, transaction
 from django.db.models import IntegerChoices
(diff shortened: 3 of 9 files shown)


care/facility/models/bed.py
```diff
@@ -59,6 +59,10 @@ def save(self, *args, **kwargs) -> None:
         return super().save(*args, **kwargs)
 
     def delete(self, *args, **kwargs) -> None:
+        if ConsultationBed.objects.filter(bed=self).exists():
+            error = f"Cannot delete Bed {self} because they are referenced as `bed` in ConsultationBed records."
+            raise ValidationError(error)
+
```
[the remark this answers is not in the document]
praffq: I don't think we need to add atomic transaction here , since we are just adding some validation and performing a single updation.

care/facility/models/patient_consultation.py
```diff
@@ -264,6 +265,25 @@ def save(self, *args, **kwargs):
...
+
+        if PatientSample.objects.filter(consultation=self).exists():
+            error = f"Cannot delete PatientConsultation {self} because they are referenced as `consultation` in PatientSample records."
+            raise ValidationError(error)
+
+        if ConsultationBed.objects.filter(consultation=self).exists():
+            error = f"Cannot delete PatientConsultation {self} because they are referenced as `consultation` in ConsultationBed records."
+            raise ValidationError(error)
+
+        return super().delete(*args)
```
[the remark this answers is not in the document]
praffq: have to keep import inside the function , since they are causing circular imports.

Also since patient consultation is referenced as PROTECTED field , so we don't need to soft delete them but to raise a validation error instead

care/facility/models/facility.py
```diff
@@ -302,14 +303,26 @@ def save(self, *args, **kwargs) -> None:
 
     @transaction.atomic
     def delete(self, *args):
-        from care.facility.models.asset import Asset, AssetLocation
+        from care.facility.models.asset import (
+            Asset,
+            AssetLocation,
+            FacilityDefaultAssetLocation,
+        )
+        from care.facility.models.patient_sample import PatientSample
```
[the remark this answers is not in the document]
praffq: causing circular imports

care/facility/models/bed.py
vigneshhari, line 65: Shouldn't we actually delete this ? ie call the super delete method ?

care/facility/models/patient.py
```diff
@@ -474,6 +475,14 @@ def save(self, *args, **kwargs) -> None:
         self._alias_recovery_to_recovered()
         super().save(*args, **kwargs)
 
+    def delete(self, *args):
```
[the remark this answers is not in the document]
praffq: we are only expecting `self`, `*args` as parameter in the delete function of our basemodel , so have to follow the same here as well

care/facility/models/asset.py
```diff
@@ -54,6 +55,15 @@ class RoomType(enum.Enum):
         null=True, blank=True, default=None, max_length=200
     )
 
+    def delete(self, *args):
+        if UserDefaultAssetLocation.objects.filter(location=self).exists():
+            error = f"Cannot delete AssetLocation {self} because they are referenced as `location` in UserDefaultAssetLocation records."
```
vigneshhari, line 60: It might make more sense to create a static errors and make it translatable as well, also keep the errors very human readable like

care/facility/models/facility.py
```diff
@@ -302,14 +303,26 @@ def save(self, *args, **kwargs) -> None:
...
+        if FacilityDefaultAssetLocation.objects.filter(facility=self).exists():
+            error = f"Cannot delete Facility {self} because they are referenced as `facility` in FacilityDefaultAssetLocation records."
+            raise ValidationError(error)
         AssetLocation.objects.filter(facility_id=self.id).update(deleted=True)
         Asset.objects.filter(
             current_location_id__in=AssetLocation._base_manager.filter(  # noqa: SLF001
                 facility_id=self.id
             ).values_list("id", flat=True)
         ).update(deleted=True)
+        FacilityUser.objects.filter(facility=self).delete()
```
vigneshhari, line 322: Shifting might also be affected

care/users/models.py
```diff
@@ -405,7 +406,23 @@ def has_add_user_permission(request):
...
+            error = f"Cannot delete User {self.username} because they are referenced as `created_by` in FacilityUser records."
+            raise ValidationError(error)
+
+        if UserDefaultAssetLocation.objects.filter(user=self).exists():
+            error = f"Cannot delete User {self} because they are referenced as `user` in UserDefaultAssetLocation records."
+            raise ValidationError(error)
+
+        FacilityUser.objects.filter(user=self).delete()
+        PatientSample.objects.filter(created_by=self).update(created_by=None)
+        PatientSample.objects.filter(last_edited_by=self).update(last_edited_by=None)
```
vigneshhari, line 425: WE can keep these, the whole reason we dont delete users is so that we can show these when the audit comes up.

vigneshhari (changes requested):

care/facility/models/bed.py
vigneshhari: My bad, didnt see the old bits !

care/users/models.py
praffq:
```
PatientSample.objects.filter(created_by=self).update(created_by=None)
PatientSample.objects.filter(last_edited_by=self).update(last_edited_by=None)
```
You mean that we don't need to set user null here?
@vigneshhari

PR #2672, opened 2024-12-22
EMR improvements

sainak:
## Proposed Changes

- TBD

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete


care/facility/models/patient.py
```diff
@@ -51,6 +51,13 @@ class RationCardCategory(models.TextChoices):
     APL = "APL", _("APL")
 
 
+class MobilityStatus(models.IntegerChoices):
```
vigneshhari: Lets switch away from integer choices forever

care/facility/models/patient.py
```diff
@@ -149,6 +156,12 @@ class TestTypeEnum(enum.Enum):
         choices=RationCardCategory, null=True, max_length=8
     )
 
+    mobility_status = models.IntegerField(
+        choices=MobilityStatus,
+        default=MobilityStatus.INDEPENDENTLY_MOBILE,
```
vigneshhari: We will need an unknown option as well, we cannot just assume that everyone in care is under this option.

PR #2676, opened 2024-12-22
Allow to use maintenance while the migrations are in progress

sainak:
## Proposed Changes

- Allow to use maintenance while the migrations are in progress

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete

## Summary by CodeRabbit

- **New Features**
	- Introduced a more structured backend approach for managing maintenance mode.
	- Added new environment variables for rate limiting, SMS notifications, and cloud storage configurations.
	- Implemented a decorator to ensure maintenance mode is disabled for specific endpoints.
	- Enhanced script to handle maintenance mode during database migrations.

- **Bug Fixes**
	- Adjusted script behavior to ensure proper toggling of maintenance mode during execution.

### DIFF
--- a/scripts/celery_beat.sh
+++ b/scripts/celery_beat.sh
@@ -1,14 +1,22 @@
 #!/bin/bash
 printf "celery-beat" > /tmp/container-role
 
-set -euo pipefail
+set -eo pipefail
 
 ./wait_for_db.sh
 ./wait_for_redis.sh
 
+if [[ "${AUTO_MAINTENANCE_MODE}" == "true" ]]; then
+    python manage.py maintenance_mode on
+fi
+
 python manage.py migrate --noinput
 python manage.py compilemessages -v 0
 python manage.py load_redis_index
+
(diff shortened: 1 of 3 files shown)


scripts/celery_beat.sh
```diff
@@ -1,15 +1,23 @@
...
 
+if [[ "${AUTO_MAINTENANCE_MODE}" == "true" ]]; then
+    python manage.py maintenance_mode on
+fi
+
 python manage.py migrate --noinput
 python manage.py compilemessages -v 0
 python manage.py load_redis_index
 
+if [[ "${AUTO_MAINTENANCE_MODE}" == "true" ]]; then
```
vigneshhari, line 17: Do you mean false here?

sainak: No it should be true, so the maintenance mode gets disabled automatically when the migrations ran successfully

sainak: we can split the behavior to separate variables, to enable and disable

PR #2683, opened 2024-12-24
Added medication administration specs, model and viewset

khavinshankar:
- Added medication administration specs, model and viewset
- Follows specs in https://ohcnetwork.notion.site/Medicine-Administration-12ae6e667ece80f3a6fbe12af2ef729a, removed site and route from the root as they were also repeated in dosage


care/emr/api/viewsets/medication_administration.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def filter_occurrence_period_end(self, queryset, name, value):
+        """
+        Filters by occurrence_period.end using DateTimeFromToRangeFilter.
+        """
+
+        if value:
+            queryset = (
+                queryset.annotate(
+                    occurrence_end_datetime=Cast(
```
vigneshhari: Nope, If you need filters, they need to be moved out of the json, no point in creating complicated queries.

care/emr/api/viewsets/medication_administration.py
```diff
@@ -0,0 +1,86 @@
...
+        return queryset
+
+
+class MedicationAdministrationViewSet(EMRModelViewSet):
+    database_model = MedicationAdministration
+    pydantic_model = MedicationAdministrationSpec
+    pydantic_read_model = MedicationAdministrationReadSpec
+    filterset_class = MedicationAdministrationFilter
+    filter_backends = [filters.DjangoFilterBackend]
+    CREATE_QUESTIONNAIRE_RESPONSE = False
```
vigneshhari: Questionnaire response needs to be created, this needs to be added in updates.

care/emr/resources/medication/administration/spec.py
```diff
@@ -0,0 +1,199 @@
...
+    entered_in_error = "entered_in_error"
+    stopped = "stopped"
+    in_progress = "in_progress"
+    on_hold = "on_hold"
+    unknown = "unknown"
+    cancelled = "cancelled"
+
+
+# https://build.fhir.org/valueset-reason-medication-not-given-codes.html#4.4.1.194
+class MedicationAdministrationStatusReason(str, Enum):
```
vigneshhari: This is a valueset that needs to be fetched from snowstorm, this cannot be stored as a valueset unless there is business logic associated to it in code.

care/emr/resources/medication/administration/spec.py
```diff
@@ -0,0 +1,199 @@
...
+
+class MedicationAdministrationPerformerFunction(str, Enum):
+    performer = "performer"
+    verifier = "verifier"
+    witness = "witness"
+
+
+class MedicationAdministrationPerformer(BaseModel):
+    actor: UUID4 = Field(
+        ...,
```
vigneshhari: ... is not needed.

care/emr/resources/medication/administration/spec.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+class MedicationAdministrationPerformerFunction(str, Enum):
+    performer = "performer"
+    verifier = "verifier"
+    witness = "witness"
+
+
+class MedicationAdministrationPerformer(BaseModel):
+    actor: UUID4 = Field(
```
vigneshhari, line 47: I am highly skeptical of storing a UUID without any relation to what the UUID is and not validating it.

vigneshhari (changes requested):

care/emr/resources/medication/administration/spec.py
khavinshankar: It always maps to User, added validation but it is still stored as JSON

care/emr/resources/medication/administration/spec.py
khavinshankar: When there is a performer, actor is a required field

vigneshhari: The three dots are not needed

care/emr/resources/medication/administration/spec.py
```diff
@@ -0,0 +1,198 @@
...
+    def perform_extra_deserialization(self, is_update, obj):
+        if not is_update:
+            obj.encounter = PatientConsultation.objects.get(
+                external_id=self.encounter
+            )  # Needs more validation
+            obj.patient = obj.encounter.patient
+            obj.request = MedicationRequest.objects.get(external_id=self.request)
+
+
+class MedicationAdministrationReadSpec(BaseMedicationAdministrationSpec):
```
vigneshhari, line 225: List does not need created_by and updated_by, move those to retrieve spec

care/emr/resources/medication/request/spec.py
```diff
@@ -33,6 +33,10 @@ class MedicationRequestStatus(str, Enum):
     draft = "draft"
     unknown = "unknown"
 
+    @classmethod
```
vigneshhari: Remove this

care/emr/api/viewsets/medication_request.py
```diff
@@ -8,15 +13,26 @@
...
     MedicationRequestSpec,
+    MedicationRequestStatus,
 )
 from care.emr.resources.questionnaire.spec import SubjectType
 from care.security.authorization import AuthorizationController
 
 
 class MedicationRequestFilter(filters.FilterSet):
     encounter = filters.UUIDFilter(field_name="encounter__external_id")
+    status = filters.MultipleChoiceFilter(
```
vigneshhari: no need for choices, just do an iexact match.

care/emr/api/viewsets/medication_request.py
```diff
@@ -44,5 +60,26 @@ def get_queryset(self):
...
+        responses={200: MedicationRequestReadSpec},
+        tags=["medication_request"],
+    )
+    @action(detail=True, methods=["POST"])
+    def discontinue(self, request, *args, **kwargs):
+        data = MedicationRequestDiscontinueRequest(**request.data)
+        request: MedicationRequest = self.get_object()
+
+        request.status = MedicationRequestStatus.ended
+        request.status_changed = datetime.now(UTC)
```
vigneshhari: use timezone.now()
from django.utils import timezone

care/emr/api/viewsets/medication_request.py
```diff
@@ -44,5 +60,26 @@ def get_queryset(self):
...
+        request.status_reason = data.status_reason
+        request.save()
+
+        return Response(
+            self.get_read_pydantic_model()
+            .serialize(request)
+            .model_dump(exclude=["meta"]),
```
vigneshhari: use to_json()

care/emr/api/viewsets/medication_request.py
```diff
@@ -8,15 +13,26 @@
...
 
 
 class MedicationRequestFilter(filters.FilterSet):
     encounter = filters.UUIDFilter(field_name="encounter__external_id")
+    status = filters.MultipleChoiceFilter(
+        field_name="status", choices=MedicationRequestStatus.choices()
+    )
+    is_prn = filters.BooleanFilter(method="filter_as_needed_boolean")
+
+    def filter_as_needed_boolean(self, queryset, name, value):
```
vigneshhari: We will not query JSON, rebuild the model or remove the filter.

vigneshhari (changes requested):

care/emr/api/viewsets/medication_request.py
khavinshankar: Created a new field in the model to record as_needed_boolean, which was initially captured in dosage instructions, a json field

care/emr/resources/medication/administration/spec.py
khavinshankar: We are using created_by in the list spec in the frontend,

image


_Actually this is similar to other specs, most of the list specs have them._

vigneshhari: We will clean up all specs with created_by in them, the joins does not make sense. If updated by is not being used remove it.

vigneshhari (approved):

PR #2711, opened 2025-01-06
Scheduling and Appointments: Validations and other minor fixes

rithviknishad:
## Proposed Changes

- [x] Separate API for cancelling a booking and prevent cancelling via PUT/PATCH on token booking viewset.
- [x] Heatmap / get slots for day: exclude cancelled statuses (inherently achieved without code changes as cancel API subtracts tokens allocated)
- [x] Fixes create exception not working
- [x] Prevent schedule deletions if active booking are present in future
- [x] Prevent updations to availability records

## Summary by CodeRabbit

- **New Features**
	- Added a new method to cancel appointments via OTP.
	- Enhanced booking cancellation functionality with reason tracking.
	- Introduced new validation mechanisms for schedules and availabilities.
	- Added a new `AvailabilityViewSet` for managing availability-related data.

- **Bug Fixes**
	- Improved authorization and validation logic for scheduling-related operations.
	- Added checks to prevent deletion of schedules with future bookings.

- **Refactor**
	- Simplified user and resource retrieval in availability exceptions.
	- Updated routing for availability management.
	- Streamlined booking and schedule update processes.
	- Enhanced time handling in slot calculations.


care/emr/api/viewsets/scheduling/booking.py
```diff
@@ -89,6 +104,23 @@ def get_queryset(self):
             .order_by("-modified_date")
         )
 
+    @classmethod
```
vigneshhari, line 99: Why is it a classmethod?

care/emr/api/viewsets/scheduling/booking.py
```diff
@@ -40,13 +52,20 @@ def filter_by_user(self, queryset, name, value):
         return queryset.filter(token_slot__resource=resource)
 
 
+def lock_cancel_appointment(token_booking):
+    token_slot = token_booking.token_slot
+    with Lock(f"booking:slot:{token_slot.id}"), transaction.atomic():
```
vigneshhari: Locks are useless unless the booking logic also uses the same lock

vigneshhari: There is only one save operation, why the atomic request ?

care/emr/api/viewsets/scheduling/booking.py
```diff
@@ -40,13 +52,20 @@ def filter_by_user(self, queryset, name, value):
...
+        token_slot.allocated -= 1
+        token_slot.save()
+
```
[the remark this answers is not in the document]
rithviknishad: If we add this check, that means somewhere an extra reduction is happening accidentally right, and the root cause of it would never see the day of light right? 🤔

vigneshhari (changes requested):

vigneshhari:
Create the API from the OTP perspective as well. User should be able to cancel a `booked` appointment

care/emr/api/viewsets/scheduling/booking.py
rithviknishad: i just tried to stick with similar pattern as of create appointment. will revert 😅 

image

vigneshhari: These are classmethods because they are called from the OTP based viewsets as well.

vigneshhari:
Is this ready for review ?

vigneshhari (approved):

Issue #2496, opened 2024-09-24
Debug issue with github not using cache for building docker images


sainak:
The docker builds actions are not using cache for building the images, the actions tab shows the cache was accessed, but the builds are ignoring it.
Tasks:
- Investigate cache issues
- modify the workflows to resolve the issues
- test them out on a fork

BarryByte:
@rithviknishad Hey, I would like to work on this issue.

himanshu-sharmav:
@sainak Hi, I would like to work on this issue.

ghost:
Hey @himanshu-sharmav are you working on this issue ? If not can i try ?

himanshu-sharmav:
I'm working on this one but down with fever.

dnyanesh-04:
@sainak Can I work on this issue?

sainak:
fixed in https://github.com/ohcnetwork/care/pull/2796

PR #2746, opened 2025-01-17
Cached deployment of Docker using github actions

kshitijk4poor:
This PR implements Docker build caching in our GitHub Actions workflow, resulting in significant performance improvements for the CI/CD pipeline.

## Proposed Changes

- Removed the local filesystem caching in favor of registry-based caching
- Added registry-based cache configuration:
  - Cache is now pulled from both the buildcache and latest tags
  - Cache is pushed to a dedicated buildcache tag
- Added BuildKit configuration to ensure latest buildkit image is used

### Associated Issue

- #2496 

## Merge Checklist

- [x] Tests added/fixed
- [ ] Update docs in `/docs`
- [x] Linting Complete

## Performance Improvements
- Initial build time: 2m 31s
- Cached build time: 45s
- Total reduction: ~70% faster builds
> can be confirmed [here](https://github.com/kshitijk4poor/care/actions), `.` is the initial build and `test-1` is the cached build.

## Verification
Added Docker layer caching which can be verified by the `CACHED` flags in the build output. Build logs clearly show the cache being utilized across different stages (marked as `CACHED` in the build output), confirming proper cache implementation and hit rates.
Screenshot 2025-01-17 at 5 16 48 PM

## Summary by CodeRabbit

- **Chores**
	- Updated deployment workflow configuration to allow manual execution.
	- Improved Docker build caching strategy using GitHub Container Registry.
	- Optimized build process with a revised remote caching mechanism.


sainak:
Hi Kshitij, I went through your changes, I don't think this will make a difference because we are missing cache due to some different reason, and there is a limit on how much data we can push to the registry so using it as a cache will exhaust it quickly, instead of this can you try https://docs.docker.com/build/ci/github-actions/multi-platform/#distribute-build-across-multiple-runners and see if this improves the caching behavior

kshitijk4poor:
> Hi Kshitij, I went through your changes, I don't think this will make a difference because we are missing cache due to some different reason, and there is a limit on how much data we can push to the registry so using it as a cache will exhaust it quickly, instead of this can you try https://docs.docker.com/build/ci/github-actions/multi-platform/#distribute-build-across-multiple-runners and see if this improves the caching behavior

Hi Aakash, I'm not sure how distributing the build across multiple runners would help with our cache issue. While it would help reduce overall build time, it doesn't directly address the underlying caching. 

Given the registry storage limits you mentioned, I think we should focus on identifying the root cause of the cache misses first.

sainak:
fixed in https://github.com/ohcnetwork/care/pull/2796

PR #2830, opened 2025-02-12
MFA v1 - TOTP

yash-learner:
This pull request introduces multi-factor authentication (MFA) using TOTP (Time-based One-Time Password) and backup codes, along with comprehensive tests for the new functionality. The changes span across multiple files, including new API endpoints, models, and utility functions.

Key changes include:

### MFA Implementation:
* **New MFA Login Endpoint**:
  - [`care/emr/api/viewsets/mfa_login.py`](diffhunk://#diff-9513817db460431c751ede96fece27c32fe354a0d89a56023eb2326776eac15cR1-R93): Added a unified MFA login endpoint that supports TOTP and backup codes.
  - Added `LoginMethod`, `MFALoginRequest`, and `MFALoginResponse` classes to handle login methods and responses.

* **TOTP Setup and Verification**:
  - [`care/emr/api/viewsets/totp.py`](diffhunk://#diff-d861ac28a70f8171ca8ff3405081e5ecd2c1b9d6c8a8c82504aaf36448f30e0aR1-R210): Added endpoints for TOTP setup, verification, disabling, and regenerating backup codes.
  - Added `TOTPSetupResponse`, `TOTPVerifyRequest`, `TOTPVerifyResponse`, and `PasswordVerifyRequest` classes for handling TOTP-related requests and responses.

### User Model Update:
* **User Specification**:
  - [`care/emr/resources/user/spec.py`](diffhunk://#diff-0795810250c6519766d8cbce1a7601a6bb03be778c2bd58e8d986c4bd0abc9d5R92-R100): Updated the `UserSpec` class to include an `mfa_enabled` attribute, indicating whether MFA is enabled for the user.

### Email Notifications:
* **TOTP Email Notifications**:
  - [`care/emr/tasks/totp.py`](diffhunk://#diff-cbadfdef44769c8d190d4a9d87ad7a2239f9fa6e5fe7cb38e4cb5c4bc1de565fR1-R55): Added tasks to send email notifications when TOTP is enabled or disabled.

### Testing:
* **MFA Login Tests**:
  - [`care/emr/tests/test_mfa_login.py`](diffhunk://#diff-a1f3ebdad3bed9ca271b3e5991faa3948888fb4132bd5550f1204c9c8c04de7bR1-R159): Added tests for the MFA login endpoint, covering various scenarios such as valid/invalid TOTP codes and backup codes.
  
* **TOTP Tests**:
  - [`care/emr/tests/test_totp.py`](diffhunk://#diff-8cb43f100f8525200597a3d1567055cee6718c361a0201f63b2f221f17eab1e0R1-R179): Added tests for TOTP setup, verification, disabling, and regenerating backup codes.

### Dependency Update:
* **Pipfile**:
  - Added `pyotp` dependency to support TOTP functionality.
  
  ----

## Merge Checklist

- [x] Tests added/fixed
- [ ] Update docs in `/docs`
- [x] Linting Complete

## Summary by CodeRabbit

- **New Features**
	- Introduced multi-factor authentication enhancements that enable a unified login using TOTP and backup codes.
	- Enabled users to set up, verify, disable two-factor authentication, and regenerate backup codes.
	- Implemented temporary tokens for MFA-enabled users and added real-time email notifications to keep you informed about authentication changes.
	- Updated account profiles to display MFA status for enhanced visibility and security.
	- Added HTML templates for user notifications regarding TOTP status changes.

- **Bug Fixes**
	- Improved error handling for various MFA scenarios to enhance user experience.

- **Tests**
	- Added comprehensive test suites for MFA login and TOTP functionalities to ensure reliability.


care/emr/api/viewsets/totp.py
```diff
@@ -0,0 +1,334 @@
...
+
+class TOTPDisableRequest(BaseModel):
+    password: str
+
+
+class TOTPDisableResponse(BaseModel):
+    message: str
+
+
+class TOTPViewSet(EMRBaseViewSet):
```
vigneshhari, line 24: There is no EMR context here, you can use rest framework base classes here.

.env.example
```diff
@@ -18,3 +18,5 @@ BUCKET_ENDPOINT=http://localhost:4566
 BUCKET_EXTERNAL_ENDPOINT=http://localhost:4566
 FILE_UPLOAD_BUCKET=patient-bucket
 FACILITY_S3_BUCKET=facility-bucket
+
+TOTP_SECRET_ENCRYPTION_KEY=your-encryption-key-here
```
vigneshhari: We already have a secret key, lets reuse that.

care/emr/api/viewsets/totp.py
```diff
@@ -0,0 +1,334 @@
...
+            }
+            for code in backup_codes
+        ]
+        user.mfa_settings = mfa_settings
+        user.save(update_fields=["mfa_settings"])
+
+        return Response({"backup_codes": backup_codes})
+
+    @staticmethod
+    def _check_totp_enabled(mfa_settings: dict):
```
vigneshhari: This function should return either true or false and let the implementer write the exception.

care/emr/api/viewsets/totp.py
```diff
@@ -0,0 +1,334 @@
...
+        if self._check_totp_enabled(mfa_settings) is None:
+            return Response(
+                {
+                    "error": "Two-factor authentication is already enabled for your account"
+                },
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+
+        secret = random_base32()
+        encrypted_secret = encrypt_string(secret)
```
vigneshhari: I didn't really understand how encrypting this string has a security advantage, for someone to hack around this, they need access to the database and not the application, but if you have access to the database then we have already lost right? 
And the backup codes are not encrypted, that already undoes all this work.

care/emr/api/viewsets/totp.py
```diff
@@ -0,0 +1,334 @@
...
+            400: {"type": "object", "properties": {"error": {"type": "string"}}},
+        },
+    )
+    @action(detail=False, methods=["POST"])
+    def verify(self, request):
+        request_data = TOTPVerifyRequest(**request.data)
+        user = request.user
+
+        if not user.totp_secret:
+            return Response(
```
vigneshhari: raise a validation error

care/emr/api/viewsets/totp.py
```diff
@@ -0,0 +1,334 @@
...
+                    for code in backup_codes
+                ],
+            }
+            user.mfa_settings = mfa_settings
+            user.save(update_fields=["mfa_settings"])
+
+            send_totp_enabled_email.delay(user.email, user.username)
+
+            response_data = TOTPVerifyResponse(
+                message="Two-factor authentication has been enabled successfully. Please save your backup codes in a secure location.",
```
vigneshhari: I dont think the backend should send any message, the message should be assumed based on the response, let the FE maintain the message.

care/emr/api/viewsets/totp.py
```diff
@@ -0,0 +1,334 @@
...
+    @action(
+        detail=False,
+        methods=["POST"],
+        permission_classes=[],
+        authentication_classes=[],
+    )
+    def login(self, request):
+        request_data = TOTPLoginRequest(**request.data)
+
+        token = RefreshToken(request_data.temp_token)
```
vigneshhari: Need exception handling around this, if an invalid token is given it will raise an error

care/emr/api/viewsets/totp.py
```diff
@@ -0,0 +1,334 @@
...
+                {"error": "Invalid token type"}, status=status.HTTP_400_BAD_REQUEST
+            )
+
+        user = User.objects.get(external_id=token["user_id"])
+        totp = TOTP(decrypt_string(user.totp_secret))
+
+        if totp.verify(request_data.code):
+            refresh = RefreshToken.for_user(user)
+
+            try:
```
vigneshhari: No need to blacklist, you can let this token exist, this token can only be used with an MFA option.

care/emr/api/viewsets/totp.py
```diff
@@ -0,0 +1,334 @@
...
+        token = RefreshToken(request_data.temp_token)
+        if not token.get("temp_token"):
+            return Response(
+                {"error": "Invalid token type"}, status=status.HTTP_400_BAD_REQUEST
+            )
+
+        user = User.objects.get(external_id=token["user_id"])
+        totp = TOTP(decrypt_string(user.totp_secret))
+
+        if totp.verify(request_data.code):
```
vigneshhari: How is the backup code used here? the backup codes are maintained by us or the package?

care/emr/api/viewsets/totp.py
```diff
@@ -0,0 +1,334 @@
...
+            400: {"type": "object", "properties": {"error": {"type": "string"}}},
+        },
+    )
+    @action(
+        detail=False,
+        methods=["POST"],
+        permission_classes=[],
+        authentication_classes=[],
+    )
+    def backup_login(self, request):
```
vigneshhari: Why cant we just use the same endpoint and check the backup codes if the others dont match?

care/emr/api/viewsets/totp.py
```diff
@@ -0,0 +1,334 @@
...
+            400: {"type": "object", "properties": {"error": {"type": "string"}}},
+        },
+    )
+    @action(
+        detail=False,
+        methods=["POST"],
+        permission_classes=[],
+        authentication_classes=[],
+    )
+    def login(self, request):
```
vigneshhari: We need ratelimiting, the token should expire within 5 mins as well.

vigneshhari (changes requested):

care/emr/api/viewsets/totp.py
yash-learner: Should I also use `DRF` `serializers` instead of Pydantic models? Because this would make error handle straightforward with DRF?  Or continue with Pydantic?

PS: I used EMR base because it is easy to propagate the Pydantic errors without any extra code

vigneshhari: Leave it, use the EMR view then

care/emr/api/viewsets/totp.py
yash-learner: By us, `pyotp` does not provide out of the box

care/emr/api/viewsets/totp.py
yash-learner: We can, but `pyotp` does not provide out of the box backup_codes support, and it is implemented by us, so keeping both in one endpoint will make the method bulkier 😁 and having separate endpoint provides clear separation

but a lot of shared code between two `login` and `backup_login` will think whether I can reduce some duplicity of code and will think what will it look like both in single endpoint

Do you have reasons to make it in a single endpoint ?

care/emr/api/viewsets/totp.py
yash-learner: > And the backup codes are not encrypted, that already undoes all this work.

Backup codes are hashed and stored  https://github.com/ohcnetwork/care/blob/f46573c9c8174c1e740a42ca8bc87eca79a60755/care/emr/api/viewsets/totp.py#L128

So even some get's access to DB. they can't use it

yash-learner: > but if you have access to the database then we have already lost right?

Yes 

but 

If it is plain text they can generate the TOTP codes and if it is encrypted, one cannot generate, and also it will save during accidental exposure 

🤔

.env.example
yash-learner: I can't find one in this https://github.com/ohcnetwork/care/blob/develop/.env.example, by any chance is it `django_secret_key` ?

care/emr/api/viewsets/totp.py
yash-learner: > This function should return either true or false and let the implementer write the exception.

Yeah, I thought this was the expected way when writting this method but wrote the exception inside the`_check_totp_enabled` as the exception is same for all three methods which are calling this so trying to not repeat the code made me write this way and also this will make understanding the code little tough and also misleading with its method name.

Will fix it 👍

care/emr/api/viewsets/totp.py
yash-learner: Fixed https://github.com/ohcnetwork/care/pull/2830/commits/ce9f6f8fa8eb24e451ec75a0902c78bb0767d3cc#diff-f7dbc85a1029e8a51e60fb570bee2b5040eeed394d3b146fbb40f3c30d91518d

care/emr/api/viewsets/totp.py
yash-learner: https://github.com/ohcnetwork/care/pull/2830/commits/ce9f6f8fa8eb24e451ec75a0902c78bb0767d3cc#diff-d861ac28a70f8171ca8ff3405081e5ecd2c1b9d6c8a8c82504aaf36448f30e0a

care/emr/api/viewsets/totp.py
yash-learner: https://github.com/ohcnetwork/care/pull/2830/commits/c0275423bc16e7d6cec2ebdbb9532cd8a18dc719

care/emr/api/viewsets/totp.py
yash-learner: Gave some thought:

I decided to keep it separate as it is a different Login method from TOTP login and created a separate viewset for [Backup Login](https://github.com/yash-learner/care/blob/issues/2614/mfa/care/emr/api/viewsets/backup_login.pyhttps://github.com/yash-learner/care/blob/issues/2614/mfa/care/emr/api/viewsets/backup_login.py) and also in future we will be adding other methods as `login using OTP via email/sms` and `passkeys`, so I think clear separation of concerns would be good.


Created a [utils](https://github.com/yash-learner/care/blob/issues/2614/mfa/care/emr/utils/mfa.py) file to store all the common methods used between all MFA methods.

What do you think ?

yash-learner: https://github.com/ohcnetwork/care/pull/2830/commits/3eeb85a0236e4d411c74d114ecbf3e9ab6313131

care/emr/api/viewsets/totp.py
yash-learner: https://github.com/ohcnetwork/care/pull/2830/commits/3eeb85a0236e4d411c74d114ecbf3e9ab6313131#diff-45f5631f3e9a5fc9b0868a9e2147f8a23287d49e6076a1c390dfaeae6ad82a60

care/emr/api/viewsets/mfa_login.py
```diff
@@ -0,0 +1,93 @@
...
+from care.emr.utils.mfa import (
+    check_mfa_ip_rate_limit,
+    check_mfa_user_rate_limit,
+    create_auth_response,
+    validate_temp_token,
+)
+from care.users.models import User
+
+
+class LoginMethod(str, Enum):
```
praffq: Shift these to specs folder , following the overall format

care/emr/api/viewsets/totp.py
```diff
@@ -0,0 +1,217 @@
...
+from rest_framework.exceptions import ValidationError
+from rest_framework.permissions import IsAuthenticated
+from rest_framework.response import Response
+
+from care.emr.api.viewsets.base import EMRBaseViewSet
+from care.emr.tasks.totp import send_totp_disabled_email, send_totp_enabled_email
+from care.emr.utils.mfa import verify_password
+
+
+class TOTPSetupResponse(BaseModel):
```
praffq: Move them to specs folder , following the overall format

care/emr/api/viewsets/totp.py
```diff
@@ -0,0 +1,217 @@
...
+class PasswordVerifyRequest(BaseModel):
+    password: str
+
+
+class TOTPDisableResponse(BaseModel):
+    message: str
+
+
+class TOTPViewSet(EMRBaseViewSet):
+    permission_classes = [IsAuthenticated]
```
praffq: not required, `IsAuthenticated` is already set as default

praffq:
@yash-learner  Are we considering a case where all the backup code are used. So we will have to alert the user to generate new set of codes? :thinking:

care/emr/api/viewsets/mfa_login.py
```diff
@@ -0,0 +1,93 @@
...
+        totp = TOTP(user.totp_secret)
+        if totp.verify(code, valid_window=1):
+            return create_auth_response(user)
+        raise ValidationError("Invalid TOTP code")
+
+    @staticmethod
+    def _handle_backup_login(user, code):
+        mfa_settings = user.mfa_settings or {}
+        backup_codes = mfa_settings.get("totp", {}).get("backup_codes", [])
+
```
praffq, line 58: We can maybe have a check like this here?
```python
if all(code_entry["used"] for code_entry in backup_codes):
        raise ValidationError("All your backup codes have been used. Please generate a new set from your account settings.")

```

yash-learner:
> @yash-learner Are we considering a case where all the backup code are used. So we will have to alert the user to generate new set of codes? 🤔

Nice catch @DraKen0009 

But practically, I would think this would never or rarely happen. In years I never logged in using a backup code. 

I think instead frontend should redirect user to user settings page whenever user logs in using backup code and prompt them to re-configure TOTP device properly, and this would regenerate the backup codes.

If a user ever continuously logs in using backup code, it indicates something is wrong with their totp device.

and 

A user can re-generate backup codes anytime as long as they are logged in

@vigneshhari what do you think should we handle this case ?

care/emr/api/viewsets/mfa_login.py
yash-learner: Nope 

Why would we raise an error and disrupt the login flow  ?

---

If we plan to handle this case then I think we should send a `boolean` field in response something like  `all_backup_codes_used`: `true`  along with `access_token` and `refresh_token` and then frontend should redirect user to user settings with banner `You have used all your backup codes, please regenerate`

care/emr/api/viewsets/totp.py
yash-learner: Yeah that's a nice thing I did't move them to separate spec file as they are really small, possibility of reusing them are rare and we have cases in codebase where we defined in viewsets 🤔.

But I just realized that I forgot to add `Spec` at end, is this as well common practice ? or is it mandatory ? @vigneshhari

care/users/migrations/0022_user_mfa_settings_user_totp_secret.py
```diff
@@ -0,0 +1,23 @@
...
+    operations = [
+        migrations.AddField(
+            model_name='user',
+            name='mfa_settings',
+            field=models.JSONField(blank=True, default=dict),
+        ),
+        migrations.AddField(
+            model_name='user',
+            name='totp_secret',
+            field=models.TextField(blank=True, null=True),
```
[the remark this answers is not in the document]
yash-learner: @coderabbitai we decided not to.

There is no point in encrypting it

care/emr/api/viewsets/mfa_login.py
praffq: >Why would we raise an error and disrupt the login flow ?

This will only blocking the request if all backup codes are used and we can handle this particular error on frontend (Show them the error then ask them to go to recovery method, redirecting to the suitable page)

vigneshhari:
Yeah let them regenerate the codes if they want to.

care/emr/api/viewsets/totp.py
```diff
@@ -0,0 +1,208 @@
...
+        return Response(response_data.model_dump())
+
+    @staticmethod
+    def _generate_backup_codes(count: int = 10) -> list[str]:
+        """Generate 8-digit backup codes."""
+        codes = []
+        for _ in range(count):
+            code = "".join(choice(digits) for _ in range(8))
+            codes.append(code)
+        return codes
```
yash-learner, line 59: The possibility of two codes being the same is very low. If it ever happens, then let's add more entropy.

care/emr/tests/test_totp.py
```diff
@@ -0,0 +1,179 @@
...
+
+        totp = TOTP(secret_key)
+        code = totp.now()
+        verify_response = self.client.post(
+            self.totp_verify_url, {"code": code}, format="json"
+        )
+        self.assertEqual(verify_response.status_code, 200)
+
+        return secret_key, verify_response.data["backup_codes"]
+
```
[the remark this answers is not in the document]
yash-learner: @coderabbitai 

You got it wrong it is already in https://github.com/ohcnetwork/care/blob/50bd4c85a08caa3cd34b3ecfa7a366b455616649/care/emr/tests/test_mfa_login.py#L88

care/emr/api/viewsets/totp.py
yash-learner: https://github.com/ohcnetwork/care/pull/2830/commits/9e4a7007e71833a0f4c68af93f042a7e84eb735d

care/emr/api/viewsets/totp.py
vigneshhari: I would move all specs to a different folder and maintain them there.

vigneshhari (approved):

yash-learner:
@DraKen0009 Thanks for doing it 🙌https://github.com/ohcnetwork/care/pull/2830/commits/d2ef52434b68853af09c329fbd8fff38f22bbed3, I thought of updating it sometime after done with reviewing MFA-FE PR.

yash-learner:
@vigneshhari @DraKen0009 Thanks for code reviews

PR #2870, opened 2025-02-25
fix: update parent has_children status on FacilityLocation child deletion

### LINKED ISSUE #2840: Location shows it has children even after deletion of its children
**Describe the bug**
Location still shows that it has children even after deleting all its children.

**To Reproduce**
1. Go to any facility settings
2. Click on locations
3. Open any location that has children
4. Delete its children ( with API since currently no option to do it in FE )
5. Go back to the parent location.
6. See error that location still shows it has children

**Expected behavior**
Parent Location should not show it has children once children are deleted.

**Additional context**
the property of location which is has_children is never being unset after deleting location.
cc: @rithviknishad

areebahmeddd:
Signed-off-by: Areeb Ahmed [hi@areeb.cloud](mailto:hi@areeb.cloud)

## Proposed Changes  

- Added logic in `destroy` of `FacilityLocationViewSet` to update `has_children` on the parent and prevent deletion if children exist, raising a `ValidationError`.  
- Updated `test_deleting_child_location_updates_parent` to create two children, delete one (check `has_children=True`), then delete the second (check `has_children=False`).  

### Associated Issue  

- Fixes #2840  

## Summary by CodeRabbit

- **New Features**
	- Enhanced the deletion process for facility locations, ensuring accurate updates to the parent's status when child locations are removed.
- **Tests**
	- Added a test to verify that deleting child locations correctly updates the parent’s status and confirms successful deletion operations.


care/emr/models/location.py
```diff
@@ -34,6 +34,15 @@ class FacilityLocation(EMRBaseModel):
     )  # Populated from FacilityLocationEncounter
     cache_expiry_days = 15
 
+    def delete(self, *args, **kwargs):
```
praffq: move this to  the viewset.

care/emr/tests/test_location_api.py
```diff
@@ -88,6 +88,33 @@ def setUp(self):
             "location-list", kwargs={"facility_external_id": self.facility.external_id}
         )
 
+    def test_deleting_child_location_updates_parent(self):
```
praffq, line 102: create 2 child locations , delete 1st check for `has_children=True` and delete 2nd to check for `has_children=False`

praffq (changes requested):

care/emr/models/location.py
areebahmeddd: Yep

care/emr/tests/test_location_api.py
areebahmeddd: Alright, got it

care/emr/api/viewsets/location.py
```diff
@@ -68,11 +68,18 @@ def get_facility_obj(self):
             Facility, external_id=self.kwargs["facility_external_id"]
         )
 
-    def validate_destroy(self, instance):
-        # Validate that there is no children if exists
-        if FacilityLocation.objects.filter(parent=instance).exists():
-            raise ValidationError("Location has active children")
-        # TODO Add validation to check if patient association exists
+    def destroy(self, request, *args, **kwargs):
```
praffq: This is not the right way to do it based on our architecture. Go through `EMRDestroyMixin` .

keep `validate_destroy` as it is.

And instead of destroy update `perform_destroy`

perform the update inside a transaction along with super.perform_destroy()

areebahmeddd: Hmm, make sense!

care/emr/api/viewsets/location.py
```diff
@@ -74,6 +75,17 @@ def validate_destroy(self, instance):
             raise ValidationError("Location has active children")
         # TODO Add validation to check if patient association exists
 
+    def perform_destroy(self, instance):
+        self.validate_destroy(instance)
```
praffq: This will be called in `super.perform_destroy` , so no need to call it again here.

vigneshhari:
Conflicts need to be resolved

areebahmeddd:
Looks good? @vigneshhari

vigneshhari (approved):

PR #2918, opened 2025-03-16
more validation, constraints and adding sort index

praffq:
## Proposed Changes

- Adding Questionnaire Text Area size limit
- Cannot remove the last user from the root org in an facility
- Max depth for organisation, facility organisation and location 10 (customizable)
- Max location and orrganisation in a facility is limited to 1000  (customizable)
- sort index support for location

## Summary by CodeRabbit

- **New Features**
  - Enhanced validations now enforce limits on hierarchical depth, organization counts, and questionnaire text lengths.
  - Improved ordering capabilities for facility locations provide a more intuitive data presentation.
  - Introduced a new sorting index for facility locations.

- **Tests**
  - Expanded test coverage verifies that hierarchy, quantity restrictions, and text length limits are applied correctly.

- **Chores**
  - Updated configuration settings enable adjustable limits for organizational structures, facility locations, and questionnaire inputs.


care/emr/api/viewsets/facility_organization.py
```diff
@@ -58,6 +59,28 @@ def validate_data(self, instance, model_obj=None):
                 raise PermissionDenied(
                     "Cannot create organizations under root organization"
                 )
+            if model_obj is None:
```
vigneshhari: We have the level_cache right ? use it.

care/emr/resources/location/spec.py
```diff
@@ -68,13 +68,29 @@ class FacilityLocationBaseSpec(EMRResource):
...
 class FacilityLocationSpec(FacilityLocationBaseSpec):
     status: StatusChoices
     operational_status: FacilityLocationOperationalStatusChoices
     name: str
     description: str
     location_type: Coding | None = None
     form: FacilityLocationFormChoices
+    sort_index: int | None = None
+
+    @field_validator("sort_index")
```
vigneshhari: Use Pydantic built in validations for max and min

care/emr/resources/questionnaire/spec.py
```diff
@@ -180,6 +183,17 @@ def validate_group_does_not_repeat(self):
             raise ValueError(err)
         return self
 
+    # we have to create this classmethod to load the limit from setting dynamically,
+    # setting max_length in field was not working for tests
```
vigneshhari: Check and figure out why, writing them out individually might be very hard to do.

config/settings/base.py
```diff
@@ -731,3 +731,17 @@
 
 # Maximum file upload size in MB
 MAX_FILE_UPLOAD_SIZE = env.int("MAX_FILE_UPLOAD_SIZE", default=5)
+
+QUESTIONNAIRE_TEXT_AREA_LIMIT_SIZE = env.int(
```
vigneshhari: Can we create these settings separate, having mission critical stuff and this together does not make sense.

vigneshhari (changes requested):

care/emr/resources/questionnaire/spec.py
praffq: only solution is to not use override settings and use the value present in settings env directly :melting_face:

vigneshhari (approved):

PR #2959, opened 2025-03-29
Add validation for date in Schedule creation and Exception creation

NikhilA8606:
## Add validation for date in Schedule creation and Exception creation

- Brief of changes made.
- Added two more conditions to the validate_period function in ScheduleCreateSpec
- Changed the valid_from and valid_to data types from datetime to date.
- Created the same validate_period function in AvailabilityExceptionWriteSpec.
- Wrote tests for both specifications.

### Associated Issue

- This proposed solution will solve this issue : [care_fe/#11009](https://github.com/ohcnetwork/care_fe/issues/11009)

### Architecture changes

## Merge Checklist

- [x] Tests added/fixed
- [ ] Update docs in `/docs`
- [x] Linting Complete

## Summary by CodeRabbit

- **New Features**
  - Enhanced scheduling and exception validations to reject dates in the past and prevent incorrect date ranges.
  - Updated schedule creation to use date-only inputs for better consistency and clarity.

- **Tests**
  - Introduced comprehensive tests to ensure the new date validations work as expected, including checks for valid date ranges and constraints, while removing outdated test cases.


rithviknishad (changes requested):
image


shouldn't these be field_validators instead of model validators? 🤔

NikhilA8606:
> shouldn't these be field_validators instead of model validators? 🤔

Comparing valid_from and valid_to together requires both fields to be assigned first, so this is best suited for @model_validator , for single fields field_validators are best

rithviknishad (approved):

care/emr/resources/scheduling/availability_exception/spec.py
```diff
@@ -32,6 +33,20 @@ class AvailabilityExceptionWriteSpec(AvailabilityExceptionBaseSpec):
...
+    def validate_dates(cls, value):
+        now = timezone.now().date()
+        if value < now:
+            raise ValueError("Date cannot be before the current date")
+        return value
+
+    @model_validator(mode="after")
+    def validate_period(self):
+        if self.valid_from > self.valid_to:
+            raise ValidationError("Valid from cannot be greater than valid to")
```
vigneshhari: Raise a value error here instead.

care/emr/resources/scheduling/schedule/spec.py
```diff
@@ -123,10 +124,18 @@ class ScheduleCreateSpec(ScheduleBaseSpec):
     user: UUID4
     facility: UUID4
     name: str
-    valid_from: datetime.datetime
-    valid_to: datetime.datetime
+    valid_from: datetime.date
```
vigneshhari: Technically speaking we should support datetime here right ? like the availability might only start from Tue noon on the 10th ( They might have a slot on tuesday morning ) ? @rithviknishad

vigneshhari (changes requested):

care/emr/resources/scheduling/schedule/spec.py
rithviknishad: oh yes

care/emr/resources/scheduling/availability_exception/spec.py
```diff
@@ -32,6 +33,20 @@ class AvailabilityExceptionWriteSpec(AvailabilityExceptionBaseSpec):
     facility: UUID4 | None = None
     user: UUID4
 
+    @field_validator("valid_from", "valid_to")
+    @classmethod
+    def validate_dates(cls, value):
+        now = timezone.now().date()
```
vigneshhari: use `care_now` for getting current time.

care/emr/resources/scheduling/schedule/spec.py
```diff
@@ -127,10 +127,18 @@ class ScheduleCreateSpec(ScheduleBaseSpec):
     valid_to: datetime.datetime
     availabilities: list[AvailabilityForScheduleSpec]
 
+    @field_validator("valid_from", "valid_to")
+    @classmethod
+    def validate_dates(cls, value):
+        now = datetime.datetime.now(tz=UTC).replace(tzinfo=None)
```
vigneshhari: `care_now` everywhere

vigneshhari (changes requested):

care/emr/resources/scheduling/schedule/spec.py
NikhilA8606: okay

NikhilA8606: do i need to change the current time in the test cases? @vigneshhari

vigneshhari (approved):

PR #2966, opened 2025-04-02
Allow none for endtime in FacilityLocationEncouterUpdate Spec

Jacobjeevan:
## Proposed Changes

- There's already validation logic to check when end_datetime isn't present for other statuses like planned/completed etc. However for active, end time isn't required.

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [x] Linting Complete

## Summary by CodeRabbit

- **New Features**
  - Facility encounter updates now offer greater flexibility by allowing the end date to be omitted when not provided, simplifying record management for end-users.

### DIFF
--- a/care/emr/resources/location/spec.py
+++ b/care/emr/resources/location/spec.py
@@ -173,7 +173,7 @@ class FacilityLocationEncounterUpdateSpec(FacilityLocationEncounterBaseSpec):
     status: LocationEncounterAvailabilityStatusChoices
 
     start_datetime: datetime.datetime
-    end_datetime: datetime.datetime | None
+    end_datetime: datetime.datetime | None = None
 
 
 class FacilityLocationEncounterListSpec(FacilityLocationEncounterBaseSpec):


care/emr/resources/location/spec.py
```diff
@@ -173,7 +173,7 @@ class FacilityLocationEncounterUpdateSpec(FacilityLocationEncounterBaseSpec):
     status: LocationEncounterAvailabilityStatusChoices
 
     start_datetime: datetime.datetime
-    end_datetime: datetime.datetime | None
```
vigneshhari, line 176: Just to be clear None was already allowed, the FE had to send None in that case, now if the FE does not send anything, BE will assume its None.

vigneshhari: This case if the value is None, BE might not send anything back at all.

vigneshhari (changes requested):

care/emr/resources/location/spec.py
Jacobjeevan: Alright, hmm I was getting end_datetime missing message from BE. But no longer an issue since that whole UI was rewritten.

PR #2995, opened 2025-04-15
Fix role and role-permission deletion logic

sainak:
Update deletion logic to prevent the removal of non-system roles and their permissions.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved the accuracy of role and permission synchronization by ensuring that only system roles and their permissions are updated or removed during the sync process. Non-system roles and their permissions remain unaffected.

### DIFF
--- a/care/security/management/commands/sync_permissions_roles.py
+++ b/care/security/management/commands/sync_permissions_roles.py
@@ -33,7 +33,7 @@ class Command(BaseCommand):
                 permission_obj.save()
             PermissionModel.objects.filter(temp_deleted=True).delete()
             # Create, update roles and delete old roles
-            RoleModel.objects.filter(system=True).update(temp_deleted=True)
+            RoleModel.objects.filter(is_system=True).update(temp_deleted=True)
             for role in roles:
                 role_obj = RoleModel.objects.filter(name=role.name).first()
                 if not role_obj:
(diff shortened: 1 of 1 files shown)


care/security/management/commands/sync_permissions_roles.py
```diff
@@ -44,7 +44,9 @@ def handle(self, *args, **options):
                 role_obj.save()
             RoleModel.objects.filter(temp_deleted=True).delete()
             # Sync permissions to role
-            RolePermission.objects.all().update(temp_deleted=True)
```
vigneshhari, line 47: Permissions are all system by default. Permissions cannot be dynamically created.

vigneshhari (changes requested):

care/security/management/commands/sync_permissions_roles.py
sainak: L47 would be required in the case, where a system role had a permission earlier but it has been removed from the permission.roles list

sainak: RolePermission is the through table

vigneshhari: Ignore dumb me and my dumb questions 😓

vigneshhari (approved):

PR #3002, opened 2025-04-21
fix(ci): workflow perms

areebahmeddd:
**Signed-off-by**: Areeb Ahmed [areebshariff@acm.org](mailto:areebshariff@acm.org)

## Summary of Changes
Part of care_fe [#11977](https://github.com/ohcnetwork/care_fe/issues/11997)


.github/workflows/deploy.yml
```diff
@@ -37,6 +38,7 @@ jobs:
 
   build:
     needs: test
+    if: github.repository == 'ohcnetwork/care'
```
sainak:
```suggestion
```

.github/workflows/deploy.yml
```diff
@@ -121,6 +123,7 @@ jobs:
 
   merge-manifests:
     needs: build
+    if: github.repository == 'ohcnetwork/care'
```
sainak:
```suggestion
```

.github/workflows/deploy.yml
```diff
@@ -177,7 +180,7 @@ jobs:
 
   notify-release:
     needs: merge-manifests
-    if: startsWith(github.event.ref, 'refs/tags/v')
+    if: github.repository == 'ohcnetwork/care' && startsWith(github.event.ref, 'refs/tags/v')
```
sainak:
```suggestion
    if: startsWith(github.event.ref, 'refs/tags/v')
```

.github/workflows/deploy.yml
```diff
@@ -187,7 +190,7 @@ jobs:
 
 #   deploy-staging-egov:
 #     needs: merge-manifests
-#     if: github.ref == 'refs/heads/develop'
+#     if: github.repository == 'ohcnetwork/care' && github.ref == 'refs/heads/develop'
```
sainak:
```suggestion
#     if: github.ref == 'refs/heads/develop'
```

.github/workflows/deploy.yml
sainak, line 1: no need to add the repository check condition on all jobs as they are already dependent on the `test` job so if that doesn't run the others will get skipped as well

sainak (changes requested):

.github/workflows/deploy.yml
areebahmeddd: ah yes, should've seen the `needs:` field, updated the pr with your changes

sainak (approved):

areebahmeddd:
@vigneshhari Looks good to you?
thanks : )

PR #3022, opened 2025-05-06
Add a validation check during consent creation.

NikhilA8606:
## Proposed Changes

- Implemented a validation error for consent creation to ensure the start of the period is not earlier than the consent date.
- The proposed solution check if the start date is less than consent date or not.

## Merge Checklist

- [x] Tests added/fixed
- [ ] Update docs in `/docs`
- [x] Linting Complete

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved validation to ensure the period start date cannot be earlier than the consent date when creating a consent.

- **Tests**
  - Updated tests to verify that creating a consent with a period start date before the consent date correctly returns a validation error.
  - Added a test confirming successful consent creation when the consent date matches the period start and end.


care/emr/resources/consent/spec.py
```diff
@@ -78,6 +78,10 @@ def validate_period_and_date(self):
             raise ValidationError(
                 "Consent date cannot be greater than the end of the period"
             )
+        if self.period.start and self.period.start < self.date:
```
vigneshhari, line 77: What about end > start ?

vigneshhari (changes requested):

care/emr/resources/consent/spec.py
NikhilA8606: Yeah, i'll add that too asap

NikhilA8606: actually we only need to check 2 condition:
1. start <= end
2. consent <= start
 then automatically consent <= end

NikhilA8606:
validation check for from and end date is already done here, so i think we can neglect that validation. @vigneshhari 

image

care/emr/resources/consent/spec.py
```diff
@@ -74,9 +74,9 @@ class ConsentBaseSpec(EMRResource):
 class ConsentCreateSpec(ConsentBaseSpec):
     @model_validator(mode="after")
     def validate_period_and_date(self):
-        if self.period.end and self.period.end <= self.date:
+        if self.period.start and self.period.start < self.date:
             raise ValidationError(
-                "Consent date cannot be greater than the end of the period"
+                "Start of the period cannot be less than the Consent date"
```
rithviknishad:
```suggestion
                "Start of the period cannot be before than the Consent date"
```

rithviknishad (changes requested):
i would also consider adding a test for same day

care/emr/resources/consent/spec.py
NikhilA8606: okay

PR #3024, opened 2025-05-06
Add a validation check during consent creation

NikhilA8606:
Related FE : https://github.com/ohcnetwork/care_fe/pull/12248
## Proposed Changes

- Implemented a validation error for consent creation to ensure the start of the period is not earlier than the consent date
- The proposed solution check if the start date is less than consent date or not.

## Merge Checklist

- [x] Tests added/fixed
- [ ] Update docs in `/docs`
- [x] Linting Complete

### DIFF
--- a/care/emr/resources/consent/spec.py
+++ b/care/emr/resources/consent/spec.py
@@ -74,9 +74,9 @@ class ConsentBaseSpec(EMRResource):
 class ConsentCreateSpec(ConsentBaseSpec):
     @model_validator(mode="after")
     def validate_period_and_date(self):
-        if self.period.end and self.period.end <= self.date:
+        if self.period.start and self.period.start < self.date:
             raise ValidationError(
-                "Consent date cannot be greater than the end of the period"
+                "Start of the period cannot be before than the Consent date"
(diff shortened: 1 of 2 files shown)


rithviknishad:
@NikhilA8606 check coderabbits nitpick comment

care/emr/resources/consent/spec.py
```diff
@@ -74,9 +74,9 @@ class ConsentBaseSpec(EMRResource):
 class ConsentCreateSpec(ConsentBaseSpec):
     @model_validator(mode="after")
     def validate_period_and_date(self):
-        if self.period.end and self.period.end <= self.date:
```
vigneshhari, line 77: start needs to be lower than end date

vigneshhari (changes requested):

care/emr/resources/consent/spec.py
NikhilA8606: image
But the PeriodSpec is already check that validation , the test catches that condition

vigneshhari (approved):

PR #3032, opened 2025-05-11
Filter consents by attachment name

yash-learner:
## Proposed Changes

- Add a Consent filter which filters the queryset by attachment_name if present, else return the original queryset.

context:

https://github.com/ohcnetwork/care_fe/pull/12033#issuecomment-2831399188

### Associated Issue

### Architecture changes

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete


yash-learner:
Actually, there is no product use case


Via @vigneshhari There will be at-most one file per consent, I don't see anything more than that. There might not be more than 1-2 consent per encounter. 

I agree with that

PR #3082, opened 2025-06-20
adding display value to answer options

praffq:
## Proposed Changes

- adding display field to asnwer_options

### Associated Issue

### Architecture changes

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete

## Summary by CodeRabbit

- New Features
  - Answer options now support optional user-facing display labels.
  - Option values may be string, integer, or coded entries with stricter validation.

- Tests
  - Test questionnaires updated to include display labels for choice options.

- Chores
  - Questionnaire fixtures refreshed to add display labels across multiple choice questions.


care/emr/resources/questionnaire/spec.py
```diff
@@ -90,16 +91,21 @@ class EnableWhen(QuestionnaireBaseSpec):
...
     initial_selected: bool = Field(
         default=False,
         description="Whether option is initially selected",
     )
 
     @field_validator("value")
     @classmethod
-    def validate_value(cls, value: str, info):
-        if not value.strip():
+    def validate_value_field(cls, value: str | int | Coding, info):
```
vigneshhari, line 105: We need to validate that the value was actually part of the options in the questionnaire and populate the display values also in the database. 
String, int and code values are stored in separate vars in the result dict if i remember correctly, can you reconfirm?

vigneshhari (changes requested):

care/emr/resources/questionnaire/spec.py
```diff
@@ -90,16 +91,21 @@ class EnableWhen(QuestionnaireBaseSpec):
 
 
 class AnswerOption(QuestionnaireBaseSpec):
-    value: Any = Field(description="Value based on question type")
+    display: str | None = Field(
+        default=None, description="Display value for the option (optional)"
+    )
+    value: str | int | Coding = Field(description="Value for the option")
```
vigneshhari, line 97: Check if all value have the same type for a given question

PR #3103, opened 2025-07-01
Add case-insensitive name filter to ProductFilters

sainak:
Introduce a name filter to enable case-insensitive searching for product names in the ProductFilters.


vigneshhari:
Can FE search Product knowledge using its API and then filter the product knowledge? This seems wasteful

PR #3107, opened 2025-07-01
fix:added an exception throw in authorize destroy

nandkishorr:
## Proposed Changes

- Added a exception throw to return error message when user deletion

### Associated Issue

- Causing the test cases to fail while deleting the user other than by super user.

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved permission handling when deleting users by providing a clear error message if the user lacks authorization.


vigneshhari (approved):

vigneshhari:
Good catch!

PR #3116, opened 2025-07-04
Add location filter to InventoryItemFilters

sainak:
Introduce a location filter to the InventoryItemFilters, allowing for more precise querying of inventory items based on their location.


vigneshhari:
This will be added along with AuthZ as AuthZ is always associated with a location.

PR #3120, opened 2025-07-07
fix:added field validator for onset datatime in allergy intolerance

nandkishorr:
## Proposed Changes

- Added a field validator for onset date time

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved date validation for allergy, condition, and medication records: onset/authored dates are normalized for timezone and cannot be set in the future. Users will see clear errors when entering invalid dates.
* **Tests**
  * Updated test expectations to reflect a standardized error message: "Date cannot be in the future."


vigneshhari:
^ Resolve the comments

care/emr/resources/common/validators.py
```diff
@@ -0,0 +1,16 @@
+import datetime
+
+from django.utils.timezone import is_aware, make_aware
+
+from care.utils.time_util import care_now
+
+
+def validate_onset_datetime_field(
+    onset_datetime: datetime.datetime,
```
vigneshhari: When you abstract something, the whole point is to abstract its features, the validation is checking if the date is in the future, it does not need to be called onset_check

vigneshhari (changes requested):

care/emr/resources/common/validators.py
nandkishorr: yep got it :)

praffq:
tests failing , please check @nandkishorr

praffq (approved):

care/emr/resources/common/validators.py
```diff
@@ -0,0 +1,16 @@
+import datetime
+
+from django.utils.timezone import is_aware, make_aware
+
+from care.utils.time_util import care_now
+
+
+def validate_datetime(
+    datetime: datetime.datetime,
+) -> datetime.datetime:
```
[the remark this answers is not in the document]
praffq: @nandkishorr  just fix this , and it should be fine

sainak (approved):

vigneshhari:
This is not just validation, we are also adding timezone data to data without timezone, was that intended?

sainak:
> This is not just validation, we are also adding timezone data to data without timezone, was that intended?

Its the same behavior as earlier, just refactored to be used in multiple places

We can reject dates without tz but that would make the frontend unhappy

PR #3123, opened 2025-07-08
Set accession_identifier default and add filter for specimens

sainak:
Assign `external_id` to `accession_identifier` when not provided and introduce a filter for `accession_identifier` in the specimen view set.

### DIFF
--- a/care/emr/api/viewsets/service_request.py
+++ b/care/emr/api/viewsets/service_request.py
@@ -96,7 +96,7 @@ class ServiceRequestViewSet(
 
     def convert_external_id_to_internal_id(self, instance):
         ids = []
-        for location in instance._locations:
+        for location in instance._locations:  # noqa: SLF001
             obj = (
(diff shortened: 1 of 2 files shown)


care/emr/api/viewsets/service_request.py
```diff
@@ -96,7 +96,7 @@ def get_facility_obj(self):
 
     def convert_external_id_to_internal_id(self, instance):
         ids = []
-        for location in instance._locations:
+        for location in instance._locations:  # noqa: SLF001
```
sainak, line 99: Someone had committed this without checks

vigneshhari (approved):

vigneshhari:
I hope we dont have to cast UUID to string here.

sainak:
> I hope we dont have to cast UUID to string here.

nope django takes care of it for us.

we have been doing this at a couple of other places as well

PR #3129, opened 2025-07-09
Add ordering support to MedicationDispense and ResourceRequest viewsets

sainak:
Implement ordering functionality for MedicationDispense and ResourceRequest viewsets by integrating the OrderingFilter and defining ordering fields.


yash-learner:
@sainak 

I do not see `created_date` being returned in the response for dispense 

I couldn't find anywhere in spec https://github.com/ohcnetwork/care/blob/vigneshhari/labs/care/emr/resources/medication/dispense/spec.py unlike medication request 

This field is required for doing grouping in FE 

Correct me, if I am wrong 

cc: @amjithtitus09

vigneshhari (approved):

PR #3135, opened 2025-07-10
Fix appointment tagging and add multi-user filtering

sainak:
Enhance the appointment management system by fixing the tagging functionality and implementing a filter that allows for multiple users to be specified when querying appointments.

## Summary by CodeRabbit

* **New Features**
  * Improved booking filtering to allow searching by multiple user external IDs at once.

* **Bug Fixes**
  * Enhanced accuracy of booking results when filtering by user.

* **Other Changes**
  * Facility information can now be accessed directly from a booking instance.

### DIFF
--- a/care/emr/api/viewsets/scheduling/booking.py
+++ b/care/emr/api/viewsets/scheduling/booking.py
@@ -1,6 +1,7 @@
 from typing import Literal
 
 from django.db import transaction
+from django.db.models.expressions import Subquery
 from django_filters import CharFilter, DateFromToRangeFilter, FilterSet, UUIDFilter
 from django_filters.rest_framework import DjangoFilterBackend
 from pydantic import UUID4, BaseModel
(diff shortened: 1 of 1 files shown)


care/emr/api/viewsets/scheduling/booking.py
```diff
@@ -49,20 +51,35 @@ class TokenBookingFilters(FilterSet):
...
+
+        token_slots = TokenSlot.objects.filter(
+            resource_id__in=Subquery(
+                SchedulableUserResource.objects.filter(
+                    user_id__in=Subquery(
+                        User.objects.filter(external_id__in=user_external_ids).values(
+                            "id"
+                        )
+                    ),
+                    facility_id__in=Subquery(
```
vigneshhari: This will always be a single object right ? can we fetch and filter instead of subquery, we will switch to a cached approach to get facility objects like user objects soon enough

vigneshhari (changes requested):

vigneshhari (approved):

PR #3137, opened 2025-07-10
fix: fixed the note thread filter

nandkishorr:
## Proposed Changes

- added filtering when the encounter is undefined

### Architecture changes

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete

## Summary by CodeRabbit

* **New Features**
  * Added a filter to allow users to find notes without an associated encounter.

### DIFF
--- a/care/emr/api/viewsets/notes.py
+++ b/care/emr/api/viewsets/notes.py
@@ -26,7 +26,12 @@ from care.security.authorization import AuthorizationController
 
 
 class NoteThreadFilters(filters.FilterSet):
-    encounter = filters.UUIDFilter(field_name="encounter__external_id")
+    encounter = filters.CharFilter(method="filter_encounter")
+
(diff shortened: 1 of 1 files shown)


care/emr/api/viewsets/notes.py
```diff
@@ -26,7 +26,12 @@
 
 
 class NoteThreadFilters(filters.FilterSet):
```
vigneshhari, line 29: Create a separate filter to filter for null conditions. Le'ts not over complicate.

vigneshhari (changes requested):

care/emr/api/viewsets/notes.py
nandkishorr: yep got it

vigneshhari (approved):

PR #3181, opened 2025-07-31
fix:updated the authorize_destroy

nandkishorr:
## Proposed Changes

- updated auth with current user can delete themselves.

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated user deletion permissions to allow users to delete their own accounts in addition to superusers.


vigneshhari (approved):
I knowingly removed this feature on account that users were deleting themselves, if this is a required feature we can merge this in, cc @bodhish

PR #3209, opened 2025-08-20
refact:fixed facility check

nandkishorr:
### Associated Issue

- Validation error due to improper uuid passing in authorize check.

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed authorization checks for patient identifier retrieval and updates so they reference the correct facility.
  * Reduced false "not found" errors and unnecessary access denials for users with valid permissions.
  * Improved reliability and consistency when accessing or updating patient identifiers across facilities.


care/emr/api/viewsets/patient_identifier.py
```diff
@@ -56,7 +56,9 @@ def authorize_create(self, instance):
 
     def authorize_update(self, request_obj, model_instance):
         if model_instance.facility:
-            facility = get_object_or_404(Facility, external_id=model_instance.facility)
+            facility = get_object_or_404(
```
vigneshhari: We already have the object, why are we fetching it again ? I think it was wrong from the get go.

care/emr/api/viewsets/patient_identifier.py
```diff
@@ -105,7 +107,9 @@ def validate_data(self, instance, model_obj=None):
 
     def authorize_retrieve(self, model_instance):
         if model_instance.facility:
-            facility = get_object_or_404(Facility, external_id=model_instance.facility)
+            facility = get_object_or_404(
```
vigneshhari: Same as above

vigneshhari (changes requested):

care/emr/api/viewsets/patient_identifier.py
sainak, line 1: @nandkishorr model_instance refers to the db model
checkout https://github.com/ohcnetwork/care/blob/674836b2668cbfa93b512106a4bdf64c8a799041/care/emr/api/viewsets/base.py

care/emr/api/viewsets/patient_identifier.py
```diff
@@ -56,7 +56,9 @@ def authorize_create(self, instance):
 
     def authorize_update(self, request_obj, model_instance):
         if model_instance.facility:
-            facility = get_object_or_404(Facility, external_id=model_instance.facility)
+            facility = get_object_or_404(
+                Facility, external_id=model_instance.facility.external_id
+            )
```
sainak:
```suggestion
```

care/emr/api/viewsets/patient_identifier.py
```diff
@@ -105,7 +107,9 @@ def validate_data(self, instance, model_obj=None):
 
     def authorize_retrieve(self, model_instance):
         if model_instance.facility:
-            facility = get_object_or_404(Facility, external_id=model_instance.facility)
+            facility = get_object_or_404(
+                Facility, external_id=model_instance.facility.external_id
+            )
```
sainak:
```suggestion
```

care/emr/api/viewsets/patient_identifier.py
sainak, line 1: we should handle the else clause for authorize_update to have 
```
        if not instance.facility and not self.request.user.is_superuser:
            raise PermissionDenied(
                "You are not authorized to create a patient identifier config"
            )
```
similar to authorize_create, 
we might not need it for authorize_retrieve if the other users are supposed to have read only access

nandkishorr: pydantic_update_model = BasePatientIdentifierSpec ,which excludes facility so it may cause error ?

care/emr/api/viewsets/patient_identifier.py
sainak: ignore this one

vigneshhari (approved):

PR #3210, opened 2025-08-20
remove unused code from specimen and minor bug in role update viewset

praffq:
## Proposed Changes

- for specimen we don't have List Apis so no use for filters and ordering 
- In roles create we were facing issue due to exclude default , so fixed that using `perform_extra_deserialization`
- Update ServiceRequestWriteSpec to use get_or_404 instead of objects.get() which can throw exceptions

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete

## Summary by CodeRabbit

* **Breaking Changes**
  * Specimen API: query-parameter filtering by accession identifier and ordering by created/modified date removed.
  * ServiceRequest: missing referenced healthcare service now returns a 404 error.

* **Behavior Changes**
  * Roles: creating a role without “is_archived” now defaults the role to archived; updates unchanged.


care/emr/resources/role/spec.py
```diff
@@ -22,7 +22,7 @@ class RoleBaseSpec(EMRResource):
     name: str | None = None
     description: str | None = None
     is_system: bool | None = False
-    is_archived: bool | None = False
+    is_archived: bool | None
```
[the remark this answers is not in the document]
yash-learner: @coderabbitai

but we are doing the following in `perform_extra_deserialization`

```
        if self.is_archived is None and not is_update:
            obj.is_archived = True

```

Shouldn't that handle ?

vigneshhari:
Let the default be None and let them use False and True. that should work

care/emr/api/viewsets/specimen.py
```diff
@@ -13,19 +11,12 @@
 from care.security.authorization.base import AuthorizationController
 
 
-class SpecimenFilters(filters.FilterSet):
```
sainak, line 16: we can go ahead if this is not required 

cc: @amjithtitus09

sainak (approved):

care/emr/api/viewsets/specimen.py
amjithtitus09: @sainak List API not required

care/emr/resources/role/spec.py
```diff
@@ -62,6 +62,9 @@ def perform_extra_deserialization(self, is_update, obj):
         else:
             obj.permissions = []
 
+        if self.is_archived is None and not is_update:
```
vigneshhari, line 65: What is the reasoning here, i did not understand.

praffq: on create we are defaulting the value to False if None is given

PR #3226, opened 2025-09-01
feat:added validation check for facility

nandkishorr:
## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized permission-denied messages for clarity and consistency.
  * Users without access now see “User cannot access patient” or “User cannot access facility.”
  * Improved handling of facility-scoped queries to reliably return a clear permission error when access is restricted.

### DIFF
--- a/care/emr/api/viewsets/encounter.py
+++ b/care/emr/api/viewsets/encounter.py
@@ -196,7 +196,7 @@ class EncounterViewSet(
                 "can_view_patient_obj", self.request.user, patient
             ):
                 return qs.filter(patient=patient)
-            raise PermissionDenied("User Cannot access patient")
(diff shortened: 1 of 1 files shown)


care/emr/api/viewsets/encounter.py
```diff
@@ -207,9 +207,12 @@ def get_queryset(self):
                 Facility, external_id=self.request.GET["facility"]
             )
 
-            return AuthorizationController.call(
```
vigneshhari, line 210: This was correct `get_filtered_encounters` correctly returns a queryset

vigneshhari (changes requested):

PR #3276, opened 2025-09-22
observation interpretation: age valuetype support

Jacobjeevan:
## Proposed Changes

- Observation Interpretation age value type support
- Changed condition evaluator so that all conditions have to be met for an interpretation (earlier, it was just any condition).

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete

## Summary by CodeRabbit

- New Features
  - More precise age calculation using birth and current/deceased dates.
  - Support for age evaluation in years, months, or days.
  - Ability to validate age-based rules for equality or within specified ranges.

- Bug Fixes
  - Corrected condition evaluation logic to return false when any rule fails, ensuring consistent outcomes.

- Chores
  - Observation definition ranges now require an explicit list instead of defaulting to an empty list (affects validation).

### DIFF
--- a/care/emr/resources/observation_definition/spec.py
+++ b/care/emr/resources/observation_definition/spec.py
@@ -141,7 +141,7 @@ class ObservationDefinitionComponentSpec(BaseModel):
     code: ValueSetBoundCoding[CARE_OBSERVATION_VALUSET.slug]
     permitted_data_type: QuestionType
     permitted_unit: ValueSetBoundCoding[CARE_UCUM_UNITS.slug] | None = None
-    qualified_ranges: list[QualifiedRangeSpec] = []
+    qualified_ranges: list[QualifiedRangeSpec]
 
     @field_validator("permitted_data_type")
     @classmethod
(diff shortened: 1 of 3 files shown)


care/utils/evaluators/evaluation_metric/patient_age.py
```diff
@@ -15,7 +20,29 @@ class PatientAgeMetric(EvaluationMetricBase):
     ]
 
     def get_value(self):
-        return self.context_object.age
+        start = self.context_object.date_of_birth or date(self.context_object.year_of_birth, 1, 1)
```
nandkishorr:
```
 start = self.context_object.date_of_birth or date(
            self.context_object.year_of_birth, 1, 1
        )
```

nandkishorr (changes requested):

nandkishorr (approved):

care/emr/resources/observation_definition/spec.py
```diff
@@ -165,7 +165,7 @@ class BaseObservationDefinitionSpec(EMRResource):
     method: ValueSetBoundCoding[CARE_OBSERVATION_COLLECTION_METHOD.slug] | None = None
     permitted_unit: ValueSetBoundCoding[CARE_UCUM_UNITS.slug] | None = None
     derived_from_uri: str | None = None
-    qualified_ranges: list[QualifiedRangeSpec] = []
+    qualified_ranges: list[QualifiedRangeSpec] | None = None
```
vigneshhari: Can you remove the default value, otherwise we wont be able to unset once set.

vigneshhari (changes requested):

vigneshhari:
This MR is also changing the way conditions are evaluated, now all condition must be met for an observation to be interpreted, earlier it was any. 
This is fine, just need to update the description to reflect this change.

care/emr/resources/observation_definition/spec.py
Jacobjeevan: Done

vigneshhari (approved):

PR #3313, opened 2025-10-23
Filters for delivery orders endpoint

nandkishorr:
## Proposed Changes

- Added filters to delivery orders endpoint in supply delivery viewset.

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete

## Summary by CodeRabbit

* **Bug Fixes**
  * Delivery orders list now consistently applies all filter parameters so results match chosen criteria.
  * Removed an unintended limit that could omit delivery orders, returning the full set of matching records.
  * Serialization of delivery items now uses the explicitly filtered set, improving alignment between filters and returned data.


care/emr/api/viewsets/inventory/supply_delivery.py
```diff
@@ -203,11 +203,17 @@ def filter_location_queryset(
 
     @action(detail=False, methods=["GET"])
     def delivery_orders(self, request, *args, **kwargs):
+        from care.emr.api.viewsets.inventory.delivery_order import DeliveryOrderFilters
```
AdityaJ2305: Shouldn't this import placed at the top of this file ? 🤔

care/emr/api/viewsets/inventory/supply_delivery.py
```diff
@@ -203,11 +203,17 @@ def filter_location_queryset(
...
+
+        delivery_filter = DeliveryOrderFilters(
+            self.request.GET, queryset=DeliveryOrder.objects.all()
+        )
+        delivery_qs = delivery_filter.qs
         queryset = self.get_queryset()
         if "request_order" not in request.GET:
             raise ValidationError("request_order is required")
         orders = queryset.values("order_id").distinct()[:100]
```
praffq: Use
```py
orders_ids = queryset.values_list("order_id", flat=True).distinct()
```
  `.values()` returns list of dict - `[{'order_id': 1}, {'order_id': 2}]` .

care/emr/api/viewsets/inventory/supply_delivery.py
```diff
@@ -203,11 +203,17 @@ def filter_location_queryset(
 
     @action(detail=False, methods=["GET"])
     def delivery_orders(self, request, *args, **kwargs):
+        from care.emr.api.viewsets.inventory.delivery_order import DeliveryOrderFilters
+
+        delivery_filter = DeliveryOrderFilters(
+            self.request.GET, queryset=DeliveryOrder.objects.all()
```
praffq: `queryset=DeliveryOrder.objects.all()` - Get's all the orders, including one which user don't have permissions on. 

So better approach is to 
```py
  queryset = self.get_queryset() -> only gets authorized orders
  orders_ids = queryset.values_list("order_id", flat=True).distinct()
  delivery_filter = DeliveryOrderFilters(
      self.request.GET, queryset=DeliveryOrder.objects.filter(id__in=orders_ids) 
  )
```

care/emr/api/viewsets/inventory/supply_delivery.py
```diff
@@ -203,11 +203,17 @@ def filter_location_queryset(
...
+        )
+        delivery_qs = delivery_filter.qs
         queryset = self.get_queryset()
         if "request_order" not in request.GET:
```
praffq, line 222: All the validations check should be performed on the top of the function so we don't perform db queries for invalid data

praffq (changes requested):

care/emr/api/viewsets/inventory/supply_delivery.py
nandkishorr: No , It was intentional since it was a single use case and for lazy loading the module and also to prevent circular import issues in future

sainak: It is preferred to keep the imports at the top unless it is causing circular imports

care/emr/api/viewsets/inventory/supply_delivery.py
```diff
@@ -203,14 +203,18 @@ def filter_location_queryset(
...
+        orders_ids = queryset.values_list("order_id", flat=True).distinct()
+        delivery_filter = DeliveryOrderFilters(
+            self.request.GET, queryset=DeliveryOrder.objects.filter(id__in=orders_ids)
+        )
         response = [
             SupplyDeliveryOrderReadSpec.serialize(order).to_json()
-            for order in orders_qs
+            for order in delivery_filter.qs
         ]
         return Response({"results": response})
```
sainak, line 233: shouldnt we filter the delivery api by order_id? This is complicating the API

cc: @vigneshhari

care/emr/api/viewsets/inventory/supply_delivery.py
nandkishorr: I've reviewed the code, and it seems that keeping imports inside the endpoints are being used in the rest of the viewsets eg: patient.py. Given that it's a followed practice should I change it ?
@sainak

care/emr/api/viewsets/inventory/supply_delivery.py
```diff
@@ -203,14 +204,16 @@ def filter_location_queryset(
 
     @action(detail=False, methods=["GET"])
     def delivery_orders(self, request, *args, **kwargs):
-        queryset = self.get_queryset()
         if "request_order" not in request.GET:
             raise ValidationError("request_order is required")
-        orders = queryset.values("order_id").distinct()[:100]
-        orders_qs = DeliveryOrder.objects.filter(id__in=orders)
+        queryset = self.get_queryset()
+        orders_ids = queryset.values_list("order_id", flat=True).distinct()
```
vigneshhari: Why was :100 removed? that was a fail safe check

vigneshhari (changes requested):

PR #3321, opened 2025-10-27
adding configrable limit to patient name

### LINKED ISSUE #142: Add configurable limit to patient name specs
Context - https://github.com/ohcnetwork/care_fe/issues/14107

praffq:
## Proposed Changes

- Added configurable limit to Patient's name field in specs

Fixes : https://github.com/ohcnetwork/roadmap/issues/142


care/emr/resources/patient/spec.py
```diff
@@ -51,7 +52,7 @@ class PatientBaseSpec(EMRResource):
     __store_metadata__ = True
 
     id: UUID4 | None = None
-    name: str = Field(max_length=200)
+    name: str = Field(max_length=settings.PATIENT_NAME_MAX_LENGTH)
```
sainak: this can break retrieve/list if the data in the db is already out of the valid limit,
it would be better to remove the limit in the base spec and override this field in the create spec.

config/settings/base.py
sainak, line 1: move it to config/settings/config.py

care/emr/resources/patient/spec.py
praffq: got it

sainak (approved):

care/emr/resources/patient/spec.py
vigneshhari: Remove this to avoid confusion

vigneshhari (changes requested):

sainak (approved):

vigneshhari (approved):

PR #3326, opened 2025-10-28
Added make commands to switch to test db during playwright tests

shivankacker:


data/test_db.dump
sainak, line 1: this should not be pushed to the repo, instead handle the case where it doesn't exist: it should run the migrations and load fixtures and then dump it to test_db.dump.
add it to gitignore.

sainak (changes requested):

sainak (approved):

vigneshhari (approved):

PR #3343, opened 2025-11-04
fix:rename internal version field to _version

nandkishorr:
## Proposed Changes

- Renamed constructed["version"] to constructed["_version"] in EMRResource 
serialization to prevent overwriting the version field for models

- Updated the version field in questionnaire json

### Associated Issue

- https://github.com/ohcnetwork/care_fe/pull/14012

## Merge Checklist

- [x] Tests added/fixed
- [ ] Update docs in `/docs`
- [x] Linting Complete

### DIFF
--- a/care/emr/resources/questionnaire/spec.py
+++ b/care/emr/resources/questionnaire/spec.py
@@ -179,7 +179,7 @@ class Question(QuestionnaireBaseSpec):
 
 
 class QuestionnaireWriteSpec(QuestionnaireBaseSpec):
-    version: str = Field("1.0", frozen=True, description="Version of the questionnaire")
+    version: str = Field(frozen=True, description="Version of the questionnaire")
     slug: SlugType | None = None
     title: str
     description: str | None = None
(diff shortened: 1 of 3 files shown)


nandkishorr:
@abhimanyurajeesh

care/emr/resources/questionnaire/spec.py
```diff
@@ -255,6 +254,7 @@ def validate_tags(cls, tags):
 
     def perform_extra_deserialization(self, is_update, obj):
         obj._organizations = self.organizations  # noqa SLF001
+        obj.version = "1.0"  # Initial version
```
praffq, line 257: The version shouldn't be hardcoded to "1.0", instead keep this value as default. Users should be able to specify the version when creating a questionnaire. Just add the `version` field to `QuestionnaireSpec` and keep it excluded from `QuestionnaireUpdateSpec` to prevent updates.

praffq (changes requested):

care/emr/resources/questionnaire/spec.py
nandkishorr: I spoke with @vigneshhari, and  clarified that the version field is primarily used for backend purposes, specifically for tracking and versioning questionnaires. It's not intended for user access or input, and currently, pydantic model is ignoring any provided values and instead passing an empty string to the database when default="1.0" is passed.

nandkishorr: cc : @vigneshhari

PR #3346, opened 2025-11-05
tag metric: support for multi condition interpretation

Jacobjeevan:
## Proposed Changes

- For all matching conditions, return the first matching interpretation
   - With earlier logic, if we had a case where both qualified ranges/conditions were matching, but only one of the interpretations match, the result would depend on the order of those conditions.
- Adjust context object for tags for apply_charge_item_definition (patient and facility in the context).
- Reset interpretation and reference range if user changes value and interpretation doesn't apply.

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete

## Summary by CodeRabbit

* **Bug Fixes**
  * Observation interpretation now tolerates missing values and supplies sensible defaults for display and processing.
  * Data model and migration updated so interpretation can be null, avoiding validation errors for absent interpretation data.

* **New Features**
  * Facility context is now included in charge evaluations for more accurate results.
  * Interpretation evaluation supports multiple matching conditions and improved tag/context handling for metrics.


care/emr/utils/compute_observation_interpretation.py
```diff
@@ -18,6 +18,9 @@ def compute_observation_interpretation(model_instance, metrics_cache):
         if interpretation:
             model_instance.interpretation = interpretation
             model_instance.reference_range = ranges
+        else:
+            model_instance.interpretation = "Normal"
```
praffq: defaulting to `Normal` might be misleading.

care/emr/utils/compute_observation_interpretation.py
```diff
@@ -18,6 +18,9 @@ def compute_observation_interpretation(model_instance, metrics_cache):
         if interpretation:
```
praffq: Update this to specifically compare with `None`
```py
if interpretation is not None:
```

praffq (changes requested):

care/emr/utils/compute_observation_interpretation.py
```diff
@@ -43,6 +46,9 @@ def compute_observation_interpretation(model_instance, metrics_cache):
             if interpretation:
```
vigneshhari, line 46: use the same condition as above.

care/utils/evaluators/evaluation_metric/encounter_tag.py
```diff
@@ -7,7 +7,7 @@
 
 
 class EncounterTagsMetric(EvaluationMetricBase):
-    context = "encounter"
+    context = "tag"
```
vigneshhari, line 10: context should be encounter itself. The check is applied on an encounter not a tag.

care/utils/evaluators/evaluation_metric/encounter_tag.py
```diff
@@ -32,12 +32,17 @@ def clean_rule(self, rule):
         return tag_config
 
     def get_value(self):
-        encounter = self.context_object
-        facility_external_id = str(encounter.facility.external_id)
+        encounter = self.context_object.get("encounter")
+        patient = self.context_object.get("patient")
+        facility = self.context_object.get("facility")
+        if encounter:
+            patient = patient or encounter.patient
```
vigneshhari, line 39: if encounter its always encounter.patient.

care/utils/evaluators/evaluation_metric/encounter_tag.py
```diff
@@ -32,12 +32,17 @@ def clean_rule(self, rule):
...
+            facility = facility or encounter.facility
         if self._value_type == "encounter":
             return [*encounter.tags]
-        patient = encounter.patient
-        patient_facility_tags = patient.facility_tags.get(facility_external_id, [])
+        if not patient or not facility:
+            return []
+        patient_facility_tags = patient.facility_tags.get(facility.external_id, [])
```
vigneshhari, line 45: Lets cast the external id to string as well here.

care/utils/evaluators/interpretation_evaluator.py
```diff
@@ -107,31 +107,35 @@ def evaluate_conditions(self, conditions, context):
                 metric_evaluator_obj = self.metric_cache[metric]
             else:
                 metric_evaluator = EvaluatorMetricsRegistry.get_evaluator(metric)
-                metric_evaluator_obj = metric_evaluator(
-                    context.get(metric_evaluator.context)
-                )
+                if metric_evaluator.context == "tag":
```
vigneshhari, line 110: Evaluator core logic cannot be changed for each context, needs to be redesigned.

care/utils/evaluators/interpretation_evaluator.py
```diff
@@ -107,31 +107,35 @@ def evaluate_conditions(self, conditions, context):
...
 
-    def get_matching_condition(self, context: dict, value: Any):
+    def get_matching_conditions(self, context: dict, value: Any):
+        matching_conditions = []
         for rule in self.rules:
             conditions = rule.get("conditions", {})
             if self.evaluate_conditions(conditions, context):
-                return rule, value
-        return None, None
+                matching_conditions.append(rule)
```
vigneshhari, line 128: We will only support the first one that was matched, Otherwise we will run into conflicts. If we need multiple interpretations, there should be another interpreter class built just for those types and it should be an option in each observation to choose which class is providing the interpretation.
That said, i dont see a case where you would need multiple interpretations.

vigneshhari (changes requested):

PR #3430, opened 2025-12-23
Fix PlugConfig cache miss for empty list

vipulpandey21:
## **Proposed Changes**

* Fixed a race condition where the PlugConfig list cache was cleared before the database write, allowing concurrent requests to repopulate the cache with stale data.
* Updated the cache invalidation order so the database is always written before clearing the cache.
* Added missing docstrings to improve code documentation and satisfy quality checks.

---

## **Associated Issue**

* This issue was discovered during local backend validation while working on caching behavior. There was no existing GitHub issue at the time of discovery.

---

## **Architecture Changes**

* No architecture changes.

---

## **Technical Details**

**Previous behavior:**

* Cache was deleted before `serializer.save()`.
* A concurrent list request could read the old database state and re-cache stale values.

**Fix:**

```python
serializer.save()
cache.delete(self.cache_key)
```

This guarantees that the cache is invalidated only after the database is consistent.

---

## **Verification**

```bash
python manage.py test
curl http://127.0.0.1:8000/api/v1/plug_config/
```

Verified that the list endpoint populates the cache correctly and that the cache is invalidated after create, update, and delete operations.

---

## **Merge Checklist**

* [x] Fixed race condition in cache invalidation
* [x] Local tests passing
* [x] Docstrings added
* [x] No breaking API changes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved caching behavior for plug configuration queries so empty results are correctly recognized and unnecessary refreshes are reduced.

* **New Features**
  * Read-only access to plug configurations (listing and retrieval) is available without authentication; create/update/delete remain restricted to admin users.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/care/users/api/viewsets/plug_config.py
+++ b/care/users/api/viewsets/plug_config.py
@@ -19,7 +19,7 @@ class PlugConfigViewset(
     def list(self, request, *args, **kwargs):
         # Cache data and return
         response = cache.get(self.cache_key)
-        if not response:
+        if response is None: # cache miss; allow cached empty list
             serializer = self.get_serializer(self.get_queryset(), many=True)
             response = serializer.data
             cache.set(self.cache_key, response)
--- a/care/users/api/viewsets/plug_config.py
+++ b/care/users/api/viewsets/plug_config.py
@@ -1,3 +1,10 @@
+"""
+API views for managing PlugConfig entries.
+
+Provides cached list endpoint and ensures cache invalidation happens
+after database writes to avoid stale cache races.
+"""
+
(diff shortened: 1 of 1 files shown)


vigneshhari (changes requested):
Remove all the comments and keep just the changes requested.

vipulpandey21:
Hi @vigneshhari, I’ve removed all comments and kept only the cache invalidation fix as requested.
Please review when convenient. Thanks!

vigneshhari (approved):

PR #3440, opened 2026-01-02
added service account flag and token generation endpoints

### LINKED ISSUE #201: Add support for service accounts in Care for programmatic access
Add support for service accounts in Care for programmatic access

- Service accounts could use the underlying user implementation for access and identity.
- Add a flag to distinguish service accounts from normal users.
-  Service accounts must NOT require a password.
- Regular user accounts should NOT be able to generate tokens; only service accounts can.

praffq:
## Proposed Changes

- Added service account flag and token genration endpoints

### Associated Issue

- Fixes : https://github.com/ohcnetwork/roadmap/issues/201

### Architecture changes

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete

## Summary by CodeRabbit

* **New Features**
  * Administrators can create, list, and manage dedicated service accounts; generate and revoke their API tokens.
  * Service accounts are flagged and excluded from regular user listings by default; filters allow including them.

* **Security**
  * Token-based API authentication enabled for service accounts.
  * Role-based permission added to control who can create service accounts and manage tokens.

* **Chores**
  * Database schema updated to add service-account flag.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


care/emr/api/viewsets/user.py
```diff
@@ -192,3 +197,84 @@ def pnconfig(self, request, *args, **kwargs):
                 setattr(user, field, request.data[field])
         user.save()
         return Response({})
+
+    @action(detail=False, methods=["POST"])
+    def create_service_account(self, request, *args, **kwargs):
```
bodhish: Why recreate the user creation flow? can't we reuse the same endpoint?

care/emr/api/viewsets/user.py
```diff
@@ -192,3 +197,84 @@ def pnconfig(self, request, *args, **kwargs):
...
+        if not request.user.is_superuser:
```
bodhish: I don't think this is a super user level action;

bodhish (changes requested):

care/emr/api/viewsets/user.py
praffq: There are alot of validations and roles involved in the user create endpoint, which are not required for service account, therefore makes it difficult to reuse for creating service account. Creating a seprate endpoint keeps it simpler

care/emr/api/viewsets/user.py
```diff
@@ -192,3 +197,84 @@ def pnconfig(self, request, *args, **kwargs):
...
+    @action(detail=False, methods=["GET"])
+    def get_service_accounts(self, request, *args, **kwargs):
+        if not request.user.is_superuser:
+            raise PermissionDenied("Only superusers can list service accounts")
+        return Response(
+            User.objects.get_entire_queryset()
+            .filter(is_service_account=True, deleted=False)
+            .values("external_id", "username"),
+            status=200,
+        )
```
rithviknishad: shouldn't the other list APIs exclude service accounts in that case? 🤔

praffq: @rithviknishad I did update the User manager to exclude the service account but migrations started throwing errors so had to revert changes.
context - https://github.com/ohcnetwork/care/actions/runs/20658115097/job/59314852938

praffq: Added fix for this in latest commit

care/security/permissions/service_account.py
```diff
@@ -0,0 +1,45 @@
...
+            VOLUNTEER_ROLE,
+            PHARMACIST_ROLE,
+        ],
+    )
+
+    can_manage_service_account = Permission(
+        "Can manage service account in care",
+        "",
+        PermissionContext.FACILITY,
+        [ADMIN_ROLE, FACILITY_ADMIN_ROLE, ADMINISTRATOR],
```
bodhish: I don't think we need to assign it to anyone by default; let deployments choose to add if they want.

care/emr/api/viewsets/user.py
```diff
@@ -192,3 +197,101 @@ def pnconfig(self, request, *args, **kwargs):
...
+            {
+                "id": str(service_account.external_id),
+                "username": service_account.username,
+                "is_service_account": True,
+            },
+            status=201,
+        )
+
+    @action(detail=False, methods=["GET"])
+    def get_service_accounts(self, request, *args, **kwargs):
```
bodhish: Again; we are duplicate efforts here, i am not sure what's stopping us from giving a simple filter is users list 🤔

bodhish (changes requested):

care/emr/api/viewsets/user.py
```diff
@@ -192,3 +197,101 @@ def pnconfig(self, request, *args, **kwargs):
...
+            raise PermissionDenied(
+                "You do not have permission to list service accounts"
+            )
+
+        return Response(
+            User.objects.get_entire_queryset()
+            .filter(is_service_account=True, deleted=False)
+            .values("external_id", "username"),
+            status=200,
+        )
```
rithviknishad: this makes it harder to get service accounts scoped under a facility's dept.

we should treat service accounts just like user accounts right?

it might be simpler to add a filter in the filterset for is_service_account and maybe default it to false if none provided?

care/security/authorization/service_account.py
rithviknishad, line 1: this needs to be imported in `authorization/__init__.py` in order for it to be registered right?

Image

care/security/permissions/service_account.py
rithviknishad, line 1: this needs to be added to the `internal_permission_handlers` in `permissions/base.py` right?

rithviknishad (changes requested):

care/emr/api/viewsets/user.py
NikhilA8606: @praffq Service Account should have all validations of a User, only difference would be that it has the flag enabled.

NikhilA8606 (commented):
@praffq what is the status now?

nandkishorr (approved):

care/emr/api/viewsets/facility_organization.py
```diff
@@ -371,6 +376,11 @@ def get_queryset(self):
             raise PermissionDenied(
                 "User does not have the required permissions to list users"
             )
-        return FacilityOrganizationUser.objects.filter(
+        queryset = FacilityOrganizationUser.objects.filter(
             organization=organization
         ).select_related("organization", "user", "role")
+
+        if "is_service_account" not in self.request.query_params:
```
vigneshhari: Use default in the boolean filter

care/emr/api/viewsets/facility_organization.py
```diff
@@ -371,6 +376,11 @@ def get_queryset(self):
...
+            queryset = queryset.filter(user__is_service_account=False)
```
rithviknishad: @praffq apply this defaulting only if the action is list, retrieve will fail if no query param is passed

rithviknishad (changes requested):

rithviknishad (approved):

vigneshhari (approved):

PR #3485, opened 2026-01-22
add nomad support for care backend

winterx64:


scripts/nomad-up.sh
```diff
@@ -0,0 +1,34 @@
...
+fi
+
+echo "Deploying Postgres..."
+nomad job run nomad/postgres.nomad
+
+echo "Deploying Redis..."
+nomad job run nomad/redis.nomad
+
+echo "Deploying Care API..."
+nomad job run nomad/backend.nomad
```
areebahmeddd, line 32: here it's `backend.nomad` but your file is named `care-backend.nomad`
consider changing it to `backend.nomad`

winterx64: okay fixed it

PR #3524, opened 2026-02-10
Fix: Facility PATCH endpoint behaving like PUT

pratx08:
## Proposed Changes

- Fixed PATCH endpoint update flow in EMRUpdateMixin to correctly propagate partial update intent (partial=True) instead of defaulting to full update behavior. [BCP Guidelines](https://datatracker.ietf.org/doc/html/rfc5789).
- Added FacilityUpdateSpec with optional fields for update operations.
- Updated FacilityViewSet to use pydantic_update_model = FacilityUpdateSpec.
- Added regression test `care/emr/tests/test_facility_api.py` to ensure `PATCH /api/v1/facility/{id}/` works with partial payload.

#### Root Cause
- partial_update() was not passing partial semantics through shared update handling.
- Facility update path was using create and required field validation, causing PATCH to act like PUT.

#### Validation
- Manual API validation in Postman:

Before Fix
image

After Fix
image

### Associated Issue
- Reproduced bug: `PATCH /api/v1/facility/{id}/ `with payload` { "name": "..." }` returned 400 with missing required fields.

## Merge Checklist

- [x] Tests added/fixed
- [ ] Update docs in `/docs`
- [x] Linting Complete

## Release Notes

* **New Features**
  * Facility records can now be updated using partial updates (PATCH requests), allowing users to modify specific fields without affecting others.
  * Added validation to ensure facility names remain unique during updates.

* **Tests**
  * Added test coverage for facility partial update operations.

### DIFF
--- a/care/emr/api/viewsets/base.py
+++ b/care/emr/api/viewsets/base.py
@@ -210,11 +210,11 @@ class EMRUpdateMixin:
 
     def update(self, request, *args, **kwargs):
         instance = self.get_object()
-        return Response(self.handle_update(instance, request.data))
+        partial = kwargs.pop("partial", False)
+        return Response(self.handle_update(instance, request.data, partial=partial))
 
     def partial_update(self, request, *args, **kwargs):
-        kwargs["partial"] = True
-        return self.update(request, *args, **kwargs)
+        return self.update(request, *args, partial=True, **kwargs)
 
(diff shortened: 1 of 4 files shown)


vigneshhari:
The Patch implementation is being worked on, Closing this MR in support of that.

PR #3537, opened 2026-02-18
Add paient level list diagnostic reports

NikhilA8606:
## Proposed Changes

- Retrive diagnostic report list by using patient id for clinical history page
- Realted FE pr : https://github.com/ohcnetwork/care_fe/pull/15815

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete

## Summary by CodeRabbit

* **New Features**
  * Diagnostic reports can now be filtered and retrieved by patient when a patient parameter is provided.

* **Security**
  * Patient-level authorization added so only users with appropriate patient permissions (and superusers) can read diagnostic reports.


nandkishorr (approved):

care/emr/api/viewsets/diagnostic_report.py
```diff
@@ -129,6 +129,16 @@ def get_queryset(self):
...
+                Patient, external_id=self.request.GET.get("patient")
+            )
+            if AuthorizationController.call(
+                "can_read_patient_diagnostic_report",
+                self.request.user,
+                patient,
+            ):
+                return queryset.filter(patient=patient)
         elif self.request.GET.get("service_request"):
             service_request = get_object_or_404(
```
rithviknishad, line 143: since it's if, elif conditions on the query params presence, client would not be aware of the order of precedence of which query param would be used first and ignored.

for example if client supplies more than one of either patient/encounter/service_request, the logic would become unpredictable from the client's perspective right?

care/emr/api/viewsets/diagnostic_report.py
```diff
@@ -129,6 +129,16 @@ def get_queryset(self):
                 encounter,
             ):
                 return queryset.filter(encounter=encounter)
+        elif self.request.GET.get("patient"):
```
vigneshhari, line 132: In a patient scoped API why do you need the patient id in the get parameter as well ?

vigneshhari (changes requested):
:/

PR #3562, opened 2026-03-06
Fix: prevent 500 error when request body is not JSON object

RCcoders:
## Summary

Fixes a Broken Object Level Authorization (BOLA) vulnerability in
`PatientViewSet.update_identifier`.

Previously, a user could update a patient identifier using a
`PatientIdentifierConfig` belonging to another facility if they knew
the configuration UUID.

## Fix

Added facility-level validation before updating identifiers:

- Ensure the identifier configuration belongs to the correct facility context.
- Enforce authorization using
  `can_write_facility_patient_identifier_config`.

## Tests

Added a security test (`test_patient_security_api.py`) that verifies:

- Cross-facility identifier updates return **403 Forbidden**
- No identifier is created in the database

All existing tests pass locally.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced input validation across API endpoints to properly handle and reject malformed requests.
  * Added authorization checks for facility-specific configuration access and updates.
  * Improved robustness of facility data handling with safer fallback values.

* **Tests**
  * Added security validation tests for configuration access control scenarios.

### DIFF
--- a/care/emr/api/viewsets/facility.py
+++ b/care/emr/api/viewsets/facility.py
@@ -7,7 +7,7 @@ from drf_spectacular.utils import extend_schema
 from rest_framework import filters as drf_filters
 from rest_framework import serializers
 from rest_framework.decorators import action, parser_classes
-from rest_framework.exceptions import PermissionDenied
+from rest_framework.exceptions import PermissionDenied, ValidationError
 from rest_framework.parsers import MultiPartParser
 from rest_framework.response import Response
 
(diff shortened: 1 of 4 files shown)


RCcoders:
Thanks for the review! I’ve addressed the CodeRabbit comments in the latest commit.

Changes made:

- Replaced facility resolution in "update_identifier" to avoid using "get_serializer_list_context()" and removed the unintended permission coupling.
- Enforced facility context validation to prevent bypassing the cross-facility check.
- Updated facility type handling in "spec.py" to fail fast on invalid mappings and removed the ""Unknown"" fallback.
- Fixed the test to grant the correct facility-scoped permission and include the facility in the request payload so the BOLA check is properly exercised.

All tests pass locally after the changes. Please let me know if any further adjustments are needed.

vigneshhari:
Pydantic already performs the required validations

PR #3587, opened 2026-03-24
Fix component serialization during observation definition update

### LINKED ISSUE #16188: Observation definition component values do not clear on update
Clearing all the present components from Observation definition and updating does not remove it from Observation Definition

yash-learner:
## Proposed Changes

- Fixes https://github.com/ohcnetwork/care_fe/issues/16188
- Fixes the component not being cleared during update when an empty `component: []` is sent in PUT request 
- Add test case for the fix

### Associated Issue

### Architecture changes

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserve omitted vs explicitly empty observation components during create/update so components are not unintentionally removed or misrepresented.
* **Tests**
  * Added API tests verifying creating, updating (including setting components to an empty list), and retrieving observation definitions behave as expected.

### DIFF
--- a/care/emr/resources/observation_definition/spec.py
+++ b/care/emr/resources/observation_definition/spec.py
@@ -217,6 +217,9 @@ class ObservationDefinitionUpdateSpec(BaseObservationDefinitionSpec):
 
     def perform_extra_deserialization(self, is_update, obj):
         obj.slug = self.slug_value
+        obj.component = [
+            c.model_dump(mode="json", exclude_defaults=True) for c in self.component
+        ]
 
 
 class ObservationDefinitionReadSpec(BaseObservationDefinitionSpec):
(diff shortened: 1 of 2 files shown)


care/emr/resources/observation_definition/spec.py
```diff
@@ -217,6 +217,9 @@ class ObservationDefinitionUpdateSpec(BaseObservationDefinitionSpec):
 
     def perform_extra_deserialization(self, is_update, obj):
         obj.slug = self.slug_value
+        obj.component = [
```
praffq: We can replace it with this
```
if self.component == []:
    obj.component = []
```

praffq (approved):

PR #3673, opened 2026-06-03
[ENG-499] Make Authz Stricter

vigneshhari:
## Proposed Changes

Stricter AuthZ
### Associated Issue

ENG-499

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved permission validation for questionnaire operations to ensure appropriate access control based on encounter and patient context.
  * Added authorization checks for questionnaire template deletion operations.


vigneshhari:
@greptileai The patient object is rare, adding it as select related would do the join for a rare edge case, its better to be lazy here.

PR #3702, opened 2026-07-04
fix: attach_account_to_invoice assigns queryset instead of list to ArrayField

KeerthiKumarR:
## Proposed Changes

- Fixes a bug in `attach_account_to_invoice` where `invoice.charge_items`
  was assigned a lazy `ValuesListQuerySet` instead of a `list`, causing
  incorrect behaviour when saving to the `ArrayField(IntegerField)` backing
  this field.
- Root cause: `charge_items.values_list("id", flat=True)` returns a
  `ValuesListQuerySet`, not a list. Django's `ArrayField.get_db_prep_value`
  only handles `list` or `tuple` — anything else is passed through raw to
  psycopg, which either crashes or silently misbehaves, meaning charge
  items are not correctly attached to the invoice.
- The identical operation appears correctly in two other places in the same
  file (lines ~143 and ~248) using `list(charge_items.values_list(...))` —
  this is a copy-paste omission where the `list()` call was dropped.
- Fix: wrap the assignment in `list()`, consistent with the two correct
  call sites already in the same file.
- Added regression tests covering: successful attachment of all billable
  charge items, rejection on non-draft invoice, and permission enforcement.

### Associated Issue

No existing GitHub issue. Found via manual code audit — the missing
`list()` call was identified by comparing the three `values_list` call
sites in the same file, two of which correctly wrap with `list()`.

## Merge Checklist

- [x] Tests added/fixed
- [ ] Update docs in `/docs`
- [x] Linting Complete

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved invoice account attachment so charge items are handled reliably during invoice updates.
  * Prevented invalid account attachment when an invoice is not in draft status.
* **Tests**
  * Added API coverage for attaching an account to an invoice.
  * Verified successful charge item attachment, draft-only behavior, and permission checks.

### DIFF
--- a/care/emr/api/viewsets/invoice.py
+++ b/care/emr/api/viewsets/invoice.py
@@ -298,7 +298,7 @@ class InvoiceViewSet(
                     account=invoice.account,
                     status=ChargeItemStatusOptions.billable.value,
                 )
-                invoice.charge_items = charge_items.values_list("id", flat=True)
+                invoice.charge_items = list(charge_items.values_list("id", flat=True))
                 sync_invoice_items(invoice)
                 invoice.updated_by = self.request.user
                 invoice.save()
--- a/care/emr/tests/test_invoice_api.py
+++ b/care/emr/tests/test_invoice_api.py
@@ -1,0 +1,147 @@
+from decimal import Decimal
+
+from django.urls import reverse
+from model_bakery import baker
+from rest_framework import status
+
+from care.emr.models import Account
+from care.emr.resources.account.spec import (
+    AccountBillingStatusOptions,
+    AccountStatusOptions,
+)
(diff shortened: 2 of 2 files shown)


KeerthiKumarR:
@coderabbitai Confirmed! The Django URL reverse name is actually `invoice-attach-account-to-invoice` (hyphenated). Attempting to reverse `invoice-attach_account_to_invoice` (with underscores) results in a `NoReverseMatch` error. The hyphenated format is correct and verified by the passing tests.

KeerthiKumarR:
Hey @vigneshhari, could you please take a look at this when you get a chance? Reviewed by both CodeRabbit and Greptile and came back clean. Could you please review the changes i made. Thank you!

PR #3704, opened 2026-07-06
Enhance logging configuration to separate console and error outputs

jesbinjoseph:
## Proposed Changes

- Brief of changes made.

### Associated Issue

### Architecture changes

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved log handling so routine messages and errors are separated more cleanly.
  * Error-level output is routed to error streams, while lower-severity logs go to standard output for clearer console visibility.
  * Updated test, deployment, and base logging configurations to keep behavior consistent across environments.


vigneshhari:
Can you make this as a plug, that was its fully safe for existing deployments ?

eeshsaxena (commented):
Clean split, and it correctly sidesteps the trap that the `below_error` filter sets up. Things I checked:

- The filter callback references `logging.ERROR`, and `import logging` is present in both `base.py` (L5) and `deployment.py` (L3), so the `CallbackFilter` won't `NameError` when it runs per record.
- The level routing has no duplication or gaps: with `console` filtered to `< ERROR` on stdout and `console_error` at `ERROR` on stderr, an INFO record goes only to stdout and an ERROR record goes only to stderr.
- The subtle part: now that `console` drops `>= ERROR`, any `propagate: False` logger left on `["console"]` would have its errors silently eaten. The PR handles that by moving `django.db.backends` and `sentry_sdk` to `["console_error"]`, and since both are `level: ERROR` they only emit error records anyway, so nothing is lost. `root` carries both handlers, so everything else is covered.

Reads correct.

PR #3708, opened 2026-07-10
[ENG-401] fix: token queue should respect sub queues while calling serve next

nandkishorr:
## Proposed Changes

- Added proper token calling in the sub_queue
- Added proper validation and current token clearing on token update
- Testcases pending !

### Associated Issue

- ENG-401

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete

## Summary by CodeRabbit

* **New Features**
  * Added a new `set-next` endpoint for sub-queues to atomically pick the earliest available token, set it as the current token, and transition it into progress.

* **Bug Fixes**
  * Strengthened token update validation to block sub-queue changes when a current token exists and to require “set-next” when status is `IN_PROGRESS`.
  * Improved compatibility checks and current-token clearing when moving away from `IN_PROGRESS`.
  * Tightened next-token selection to only consider unassigned tokens and refined persistence to update only required fields.

* **Tests**
  * Updated token API tests to use `CREATED` payloads and expect `CREATED` responses.


eeshsaxena (commented):
Traced the refactor and the behavior lines up. The resource-mismatch check is the same rule as before, just flattened into one condition (raise when the token moves to a different sub_queue whose resource differs). The current-token clear now fires on two cases instead of one: the existing "sub_queue changed" case, plus the new "was IN_PROGRESS and is no longer" case, so completing or cancelling the current token releases the sub_queue's `current_token` rather than leaving it pointing at a finished token. That pairs well with the new `validate_data` guard that routes IN_PROGRESS assignment through the `set_next` endpoint.

One thing worth a sanity check: `obj.sub_queue != instance.sub_queue` and especially `instance.sub_queue.resource` assume `instance.sub_queue` is the resolved sub-queue object by `perform_update` time, whereas in `validate_data` it is compared against `model_obj.sub_queue.external_id` as a raw id. That looks like the intended resolved-vs-raw split between the two methods, just calling it out since they treat `instance.sub_queue` as different types.

PR #3712, opened 2026-07-15
fix: revisit discount day calculation uses abs(timedelta).days instead of abs(timedelta.days)

KeerthiKumarR:
## Proposed Changes

- Fixes a bug in `lock_create_appointment` where the revisit discount
  day-gap calculation gives the wrong answer when the patient's previous
  paid appointment was in the evening and the new slot is in the morning
  (i.e. time-of-day crosses midnight between the two datetimes).
- Root cause: the code computed
  `diff_days = (booking_start_time - new_booking_start_time).days`,
  where the subtraction always yields a negative `timedelta` (since
  `paid_on` is in the past relative to the new slot). Python's
  `timedelta.days` floors toward negative infinity on negative values —
  so a gap of 3 days 2 hours gives `diff.days = -4`, meaning
  `abs(diff.days) = 4` instead of the correct `3`. A patient who paid
  within the revisit window (e.g. 3 days ago at 11pm, new slot at 1am)
  gets their discount **incorrectly denied** purely because of a
  time-of-day difference crossing midnight.
- Fix: changed to `abs(booking_start_time - new_booking_start_time).days`
  — take `abs()` on the full `timedelta` first, then extract `.days`.
  This gives the correct non-negative day count regardless of time-of-day.
  Also removed the now-redundant `abs()` in the downstream comparison
  since `diff_days` is now always non-negative.
- Added regression tests covering: discount correctly applied when gap is
  within window but crosses midnight, discount correctly denied when gap
  exceeds window, and no discount when no previous booking exists.

## Merge Checklist

- [x] Tests added/fixed
- [ ] Update docs in `/docs`
- [x] Linting Complete

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected revisit-discount eligibility for bookings made across different times of day by fixing the day-gap calculation.
  * Revisit discounts are applied only when the elapsed time falls within the configured revisit window; otherwise full pricing is used.
  * If there is no prior paid booking, full pricing is consistently applied.
* **Tests**
  * Added automated tests covering revisit-discount selection, including cross-midnight timing and boundary/out-of-window cases.

### DIFF
--- a/care/emr/api/viewsets/scheduling/availability.py
+++ b/care/emr/api/viewsets/scheduling/availability.py
@@ -168,13 +168,13 @@ def lock_create_appointment(token_slot, patient, created_by, note):
                 diff_days = None
             else:
                 new_booking_start_time = token_slot.start_datetime
-                diff_days = (booking_start_time - new_booking_start_time).days
+                diff_days = abs(booking_start_time - new_booking_start_time).days
         else:
             diff_days = None
         if (
             schedule.revisit_allowed_days
             and diff_days is not None
-            and abs(diff_days) <= schedule.revisit_allowed_days
+            and diff_days <= schedule.revisit_allowed_days
         ):
             charge_item_definition = schedule.revisit_charge_item_definition
         else:
--- a/care/emr/tests/test_booking_api.py
+++ b/care/emr/tests/test_booking_api.py
@@ -7,12 +7,20 @@ from django.utils import timezone
 from care.emr.models import (
     Availability,
     AvailabilityException,
+    ChargeItemDefinition,
     SchedulableResource,
     Schedule,
     TokenBooking,
     TokenSlot,
 )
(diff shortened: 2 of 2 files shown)


KeerthiKumarR:
Hey @vigneshhari, could you take a look at this when you get a chance!

vigneshhari:
This is correct but changing this will impact application logic of existing users, Folks who set this intentionally will have to change it to get the intended behavior.

KeerthiKumarR:
> This is correct but changing this will impact application logic of existing users, Folks who set this intentionally will have to change it to get the intended behavior.

@vigneshhari Okay, Will it be good to document the current behavior explicitly in the code so it says it's intentional?

vigneshhari:
We are building our internal documentation for care now, I will make sure that this is added into it.

PR #3719, opened 2026-07-29
[ENG-737] Questionnaire Redesign

vigneshhari:
## Proposed Changes

- Did stuff

### Associated Issue

ENG-737

## Summary by CodeRabbit

* **New Features**
  * Added support for facility and resource-specific questionnaires, including submissions for locations, devices, and facilities.
  * Added resource questionnaire response and observation APIs with filtering, analysis, and access controls.
  * Added questionnaire revision tracking and cleaned response data for easier reporting.
  * Expanded value sets with facility scoping, inheritance, organization management, and user preferences.
* **Bug Fixes**
  * Improved response handling so questionnaire revisions and related observations remain synchronized.
  * Strengthened authorization for questionnaire and value-set access across facilities and organizations.


eeshsaxena:
One edge on the parent chain: `create_composition` walks `self.parent` recursively with no visited set or depth cap, and `perform_extra_deserialization` doesn't stop a parent from pointing back at a descendant. A cyclic link (A→B→A) would then hit RecursionError on every search/lookup. A `_seen` set, or an ancestor check when assigning the parent, would guard it.

PR #3721, opened 2026-07-31
Improve phone number validator type validation

4arjun:
## Proposed Changes

- Added explicit validation for unsupported `PhoneNumberValidator` types.
- Changed unsupported validator type handling from raw `KeyError`/`TypeError` failures to clear `ValueError` messages.
- Require `types` to be a non-empty collection so one-shot iterators/generators are rejected for safer Django validator deconstruction.
- Preserved existing behavior for supported phone number types.
- Added tests for invalid `types`, unsupported types, non-string values, unhashable values, and iterator input.

### Associated Issue

- No linked issue. This is a small validation hardening change for shared phone number validator behavior.

## Merge Checklist

- [x] Tests added/fixed
- [ ] Update docs in `/docs`
- [x] Linting Complete
- [x] Verified no migration changes are generated

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved phone-number validation configuration handling.
  * Empty, unsupported, unhashable, or non-string phone-number types now produce clear validation errors.
  * Phone-number type configurations must now use a non-string collection, with supported mobile and landline types validated correctly.

* **Tests**
  * Added coverage for invalid, empty, unsupported, unhashable, and non-string configuration inputs.
  * Added validation coverage for supported mobile and landline types.

### DIFF
--- a/care/utils/models/validators.py
+++ b/care/utils/models/validators.py
@@ -1,5 +1,5 @@
 import re
-from collections.abc import Iterable
+from collections.abc import Collection, Iterable
 from fractions import Fraction
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


eeshsaxena (commented):
Nice tightening. Two real improvements here:

- Materializing `types = tuple(types)` before the emptiness check avoids the old `len(types)` blowing up on a generator (generators have no `__len__`), and requiring `Collection` rather than `Iterable` makes the empty check well-defined.
- The unsupported-type loop is safe against unhashable inputs: `not isinstance(type_, str) or type_ not in self.regex_map` short-circuits on the isinstance check, so a `[]` never reaches `[] in self.regex_map`, which would otherwise raise `TypeError: unhashable type`. The `[]` and `1` test cases exercise that path.

Reads correct.

PR #3743, opened 2026-08-18
[ENG-925] removed encounter_class_history field

praffq:
## Proposed Changes

- remove `encounter_class_history` field from Encounter
- added validation to not allow encounter class to be changed once created. 

## Merge Checklist

- [ ] Tests added/fixed
- [ ] Update docs in `/docs`
- [ ] Linting Complete

## Summary by CodeRabbit

* **Changes**
  * Simplified encounter data by removing encounter-class history from records and API responses.
  * Encounter class is now provided when creating encounters and included in encounter listings.
  * Existing encounters are migrated automatically to the updated data structure.


nandkishorr (approved):

care/emr/api/viewsets/encounter.py
```diff
@@ -148,6 +148,9 @@ def validate_data(self, instance, model_obj=None):
             if not Facility.objects.filter(external_id=instance.facility).exists():
                 raise ValidationError("Facility does not exist")
 
+        elif model_obj.encounter_class != instance.encounter_class:
```
vigneshhari: We can just not accept it right?

vigneshhari (changes requested):

PR #3761, opened 2026-09-19
fix(emr): register report models (Template, ReportUpload) at app load

amjithtitus09:
## Problem

At app load, `care/emr/models/__init__.py` imported every model subpackage **except** `report`. As a result `emr.Template` and `emr.ReportUpload` were never registered eagerly — they only got registered lazily when something imported them (e.g. the report viewsets via the URLconf).

This is fine on its own, but it breaks as soon as a plug imports `ReportUpload` in its `AppConfig.ready()` **without** also importing `Template`. `ReportUpload.template` is a `ForeignKey("emr.Template", ...)` declared as a lazy string, so importing `ReportUpload` does not pull in `Template`. Django then runs model system checks against a state where `ReportUpload` is registered but `Template` is not:

```
SystemCheckError: System check identified some issues:

ERRORS:
emr.ReportUpload.template: (fields.E300) Field defines a relation with model 'emr.Template',
which is either not installed, or is abstract.
```

This is exactly what happens in the `care_create` reference stack once the `care-abdm` plug added a `post_save` receiver on `ReportUpload` in its `ready()` path — the failure surfaces on the first management command at container start (e.g. `migrate` / `collectstatic`).

It is also intermittent with a plain `manage.py check`: whether the URLconf (which imports both report models) is imported before the model checks run depends on Django's check-registry ordering (a `set`). `manage.py check --tag models` fails deterministically.

## Fix

Import the `report` subpackage models like every other subpackage, at the end of `care/emr/models/__init__.py` (after the base models they depend on are defined):

```python
from .report.template import *        # noqa F403
from .report.report_upload import *   # noqa F403
```

This guarantees both `Template` and `ReportUpload` are registered at app load, independent of import order or which plug is installed.

## Notes

- `report` was the only model subpackage not imported here (`scheduling` already is).
- Appended at the end because `report/template.py` and `report/report_upload.py` do `from care.emr.models import SlugBaseModel / EMRBaseModel`, so the base models must be defined first.

## Summary by CodeRabbit

* **New Features**
  * Added access to report templates and report uploads through the EMR models package.


vigneshhari:
Need jira ticket

